PR #4, opened 2018-10-03
server:latest -> server:CURRENT_VERSION everywhere

dadlerj:
Is there a reason why we use `:latest` in all of the cloud host deployment docs? I think it's a bit irresponsible to recommend using our insiders/RC for "production" deployments (i.e., when users deploy to cloud hosts).

Also, it's just generally inconsistent with other parts of the docs. 

cc @KattMingMing


dadlerj:
Also, non-official releases break update checks (so, if I'm not mistaken, they'll never get a notification to upgrade):

image

dadlerj:
Hah — looks like this issue has already been solved! https://github.com/sourcegraph/website/blob/master/src/components/ServerVersionNumber.tsx

I'll update this PR

I found this component, by the way, from the server release process doc @slimsag

dadlerj:
Thanks!

attfarhan:
@nicksnyder you can go to Netlify to see the build log. The logs aren’t public yet. You will need to merge or rebase off of master to get this to pass, as I added to build script after this PR was opened.

PR #16, opened 2018-10-09
Add gitlab docs

ijsnow:


attfarhan:
@ijsnow separately, the images on https://about.sourcegraph.com/docs/features/browser-extension/configuration/ are outdated. We should update those.

PR #19, opened 2018-10-10
Update style guide for OSS

francisschmaltz:
- Move style guide to root directory
- Revert references for title case back to sentence case 😢 
- Update naming convention for OSS and paid tiers
- Add callout that OSS is **not** the same as Core


STYLEGUIDE.md
```diff
@@ -32,8 +31,10 @@ Assume the reader is a busy non-native English speaker.
 ### Referring to the Product and Features
 
 - Sourcegraph: main product, prefer using this name unless you need to be more precise
-  - Sourcegraph Server: the form of our product that ships as the `sourcegraph/server` Docker image and runs on a single node
-  - Sourcegraph Data Center: the form of our product that runs on Kubernetes
+  - Sourcegraph OSS: the open source code for Sourcegraph. This is **not** the same as Sourcegraph Core.
+  - Sourcegraph Core: previously was called Sourcegraph Server, this build is the free tier of Sourcegraph built separately from Sourcegraph OSS.
```
dadlerj: > previously was called Sourcegraph Server

This isn't _exactly_ right, since what was previously called "Server" really now maps to both "Core" and "Enterprise Starter" in the new model

STYLEGUIDE.md
```diff
@@ -32,8 +31,10 @@ Assume the reader is a busy non-native English speaker.
...
+  - Sourcegraph Enterprise Starter: the tier of Sourcegraph that includes some features required for dev-ops/admins to deploy in their corporate environment.
+  - Sourcegraph Enterprise: the tier of Sourcegraph that includes all enterprise features for dev-ops/admins to deploy Sourcegraph at a large scale.
```
dadlerj: "Enterprise includes the cluster deployment option formerly called Data Center"

STYLEGUIDE.md
```diff
@@ -43,7 +44,7 @@ Assume the reader is a busy non-native English speaker.
       - Sourcegraph['s] Firefox add-on
       - Sourcegraph['s] Safari extension
 
-When specifically distinguishing between Server and Data Center, it might help to say "the [single-node] Server deployment option" and "the Data Center [cluster] deployment option".
+When specifically distinguishing between Core and OSS it is important to note that Core is not built from the OSS code base and OSS does not include tracking data, ability to upgrade, or Sourcegraph extensions. These are included in the Core version of Sourcegraph, that is built from the same code as Sourcegraph Enterprise and has the ability to upgrade later.
```
dadlerj: > does not include tracking data
We don't collect any telemetry from "dev"/oss, but we also haven't ripped this code out of the common codebase. I'd remove this for now.

STYLEGUIDE.md
francisschmaltz: Should I remove this?

dadlerj: Yeah, I'd just remove that, because it's not correct. If you prefer, you could replace it with something else that explains how the old "Server" is now both "Core" and "Enterprise Starter" but it's your call on whether that's necessary at all.

STYLEGUIDE.md
```diff
@@ -12,8 +12,7 @@ Our copy should be:
 ## General
 
 - Seek to make content available, and coherent, to all peoples.
-- Use Title Case for headers, buttons, features, and short lists.
-- Use sentence case for non-header text. When sentence case is used, full punctuation should be used.
+- Use sentence case for all text. When sentence case is used, full punctuation should be used.
```
sqs: Full punctuation (periods, etc.) should not be used in many cases: buttons, labels, nav links, headlines, etc. I think you can just omit the "When sentence case is used, full punctuation should be used." sentence and it will be clear.

STYLEGUIDE.md
```diff
@@ -32,8 +31,10 @@ Assume the reader is a busy non-native English speaker.
 ### Referring to the Product and Features
 
 - Sourcegraph: main product, prefer using this name unless you need to be more precise
-  - Sourcegraph Server: the form of our product that ships as the `sourcegraph/server` Docker image and runs on a single node
-  - Sourcegraph Data Center: the form of our product that runs on Kubernetes
+  - Sourcegraph OSS: the open source code for Sourcegraph. This is **not** the same as Sourcegraph Core.
```
sqs: open-source (with hyphen when used as an adjective)

Also I would not include Sourcegraph OSS in this list, as it's not really a product. You could mention it in a bullet point underneath and say something like "When referring to the build result of the open-source repository, use the name Sourcegraph OSS."

STYLEGUIDE.md
sqs: BTW, the model for us to follow is GitHub. They have *very* consistent style in all their pages and materials, and they use sentence case. If in doubt about a rule, look at what GitHub does.

sqs:
@francisschmaltz 	Ping on this, I'd like to have this merged so I can link to it from the new docs

francisschmaltz:
@sqs Updated from these notes

STYLEGUIDE.md
```diff
@@ -32,8 +31,9 @@ Assume the reader is a busy non-native English speaker.
 ### Referring to the Product and Features
 
 - Sourcegraph: main product, prefer using this name unless you need to be more precise
-  - Sourcegraph Server: the form of our product that ships as the `sourcegraph/server` Docker image and runs on a single node
-  - Sourcegraph Data Center: the form of our product that runs on Kubernetes
+  - Sourcegraph Core: This build is the free tier of Sourcegraph built separately from Sourcegraph OSS.
+  - Sourcegraph Enterprise Starter: the tier of Sourcegraph that includes some features required for dev-ops/admins to deploy in their corporate environment.
+  - Sourcegraph Enterprise: the tier of Sourcegraph that includes all enterprise features for dev-ops/admins to deploy Sourcegraph at a large scale. Enterprise includes the cluster deployment option formerly called Data Center.
```
sqs: Remove "for dev-ops/admins" because

1. it is not clear who these personas are, specifically (ie they aren't defined anywhere)
1. I would say "dev tools team members" is more accurate than "dev-ops", but that's a mouthful
1. saying who is not necessary here

STYLEGUIDE.md
```diff
@@ -43,7 +43,9 @@ Assume the reader is a busy non-native English speaker.
       - Sourcegraph['s] Firefox add-on
       - Sourcegraph['s] Safari extension
 
-When specifically distinguishing between Server and Data Center, it might help to say "the [single-node] Server deployment option" and "the Data Center [cluster] deployment option".
+When referring to the build result of the open-source repository, use the name Sourcegraph OSS.
+
+When specifically distinguishing between Core and Sourcegraph OSS it is important to note that Core is not built from the open-source code base and Sourcegraph OSS does not include the ability to upgrade, or Sourcegraph extensions. These are included in the Core version of Sourcegraph, that is built from the same code as Sourcegraph Enterprise and has the ability to upgrade later.
```
sqs: Technically Sourcegraph OSS supports Sourcegraph extensions, just not accessing the extension registry on Sourcegraph.com.

francisschmaltz: How's this:

sqs: Why that instead of just "...the ability to upgrade or access the extension registry on Sourcegraph.com"?

francisschmaltz: I want to call out Sourcegraph extensions by name

sqs: My version is more accurate. Referring to "public Sourcegraph extensions" implies that they can use private Sourcegraph extensions (which they can't, at least not in the same way that Sourcegraph Enterprise can). It also makes the mechanism unclear and sound less legitimate, as though we're disabling functionality instead of just limiting access to an online service.

PR #28, opened 2018-10-16
docs: update Sourcegraph extension docs

### LINKED ISSUE #27: Remove references to 2.11 in docs
Now that 2.12 is out, this page should be updated. Ideally, we should never hard-code version numbers into docs if we can avoid it https://about.sourcegraph.com/docs/extensions

chrismwendt:
Fixes #27

### DIFF
--- a/docs/server/tutorials/extensions.md
+++ b/docs/server/tutorials/extensions.md
@@ -4,63 +4,43 @@ title: Sourcegraph extensions - setup and usage instructions (alpha)
 permalink: docs/extensions
 ---
 
-Sourcegraph is getting an extension API that makes easy to add new features and information to Sourcegraph, GitHub, and other code hosts and review tools that our browser extensions integrate with. The [Sourcegraph extension API](https://github.com/sourcegraph/sourcegraph-extension-api) allows extensions to provide hovers, definitions, references, line decorations, buttons, menu items, and other similar contributions. For more information, see the open-source [sourcegraph-extension-api repository](https://github.com/sourcegraph/sourcegraph-extension-api).
+Sourcegraph is getting an extension API that makes it easy to add new features and information to Sourcegraph, GitHub, and other code hosts and review tools that our browser extensions integrate with. The [Sourcegraph extension API](https://github.com/sourcegraph/sourcegraph-extension-api) allows extensions to provide hovers, definitions, references, line decorations, buttons, menu items, and other similar contributions. For more information, see the open-source [sourcegraph-extension-api repository](https://github.com/sourcegraph/sourcegraph-extension-api).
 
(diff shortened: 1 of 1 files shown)


docs/server/tutorials/extensions.md
```diff
@@ -4,63 +4,43 @@ title: Sourcegraph extensions - setup and usage instructions (alpha)
 permalink: docs/extensions
 ---
 
-Sourcegraph is getting an extension API that makes easy to add new features and information to Sourcegraph, GitHub, and other code hosts and review tools that our browser extensions integrate with. The [Sourcegraph extension API](https://github.com/sourcegraph/sourcegraph-extension-api) allows extensions to provide hovers, definitions, references, line decorations, buttons, menu items, and other similar contributions. For more information, see the open-source [sourcegraph-extension-api repository](https://github.com/sourcegraph/sourcegraph-extension-api).
+Sourcegraph is getting an extension API that makes it easy to add new features and information to Sourcegraph, GitHub, and other code hosts and review tools that our browser extensions integrate with. The [Sourcegraph extension API](https://github.com/sourcegraph/sourcegraph-extension-api) allows extensions to provide hovers, definitions, references, line decorations, buttons, menu items, and other similar contributions. For more information, see the open-source [sourcegraph-extension-api repository](https://github.com/sourcegraph/sourcegraph-extension-api).
```
dadlerj: Maybe change "is getting" to "has"? Not sure if you intend to present this as a future thing (since we say it's in alpha below) or current. Either is fine.

docs/server/tutorials/extensions.md
```diff
@@ -4,63 +4,43 @@ title: Sourcegraph extensions - setup and usage instructions (alpha)
...
-   ```
-
-   There are no known issues that occur when this site configuration feature flag is enabled. It enables extensions-related API endpoints for users, but users won't see any change in the behavior of Sourcegraph until they individually opt in (in the next step).
-
-1. Then, each user who wants to use Sourcegraph extensions must run the following JavaScript code in their browser's devtools JavaScript console (when their browser is open to page on the Sourcegraph instance):
-
-   ```
-   localStorage.platform=true; location.reload()
-   ```
+Sourcegraph extensions are available in alpha on Sourcegraph.com, Sourcegraph Enterprise and [Sourcegraph for Chrome](https://chrome.google.com/webstore/detail/sourcegraph/dgjhfomjieaadpoljlnidmbgkdffpack) or [Sourcegraph for Firefox](https://addons.mozilla.org/en-US/firefox/addon/sourcegraph/).
```
dadlerj: Also, see @sqs's comment at https://github.com/sourcegraph/about/pull/19#discussion_r225365698

docs/server/tutorials/extensions.md
chrismwendt: :+1: to "has", since they're here already

PR #39, opened 2018-10-19
Primary workflow user experience improvements

sqs:
WIP doc describing this month's work


sqs:
@vanesa 	Please check this over to see if there are any other changes before it can go live, then assign me.

PR #44, opened 2018-10-24
link to friendlier new-subscription page instead of "Contact us"

sqs:
After https://github.com/sourcegraph/enterprise/pull/13708 is merged, the new-subscription page on Sourcegraph.com will be friendlier: you can price out a subscription even if you're not signed in. Previously it would dump you on a sign-in page, so we didn't feel comfortable linking directly to it from this pricing page.

screenshot from 2018-10-24 01-54-57


sqs:
@KattMingMing 	Either before or after this is merged in a separate PR, you should add the trial link for Enterprise.

dadlerj:
Updated some analytics issues. Several of these have been messed up for a while.

PR #45, opened 2018-10-25
feat: add trial form page

KattMingMing:
Updates the contact us form to a free trial form.

screenshot 2018-10-25 12 34 27

screenshot 2018-10-25 12 34 33

### DIFF
--- a/website/src/pages/contact/sales.tsx
+++ b/website/src/pages/contact/sales.tsx
@@ -28,12 +28,8 @@ export default class Trial extends React.Component<any, any> {
                     <div className="jumbotron text-right dark-7">
                         <div className="row">
                             <div className="sales py-0 d-flex flex-grow flex-column flex-lg-row align-items-center justify-content-around">
-                                <img className="logo" src="/sourcegraph/sourcegraph-mark.png" alt="logo" />
                                 <div>
-                                    <h1 className="subtitle">Try Sourcegraph Enterprise free for 30 days</h1>
-                                    <p className="text-center">
-                                        Sourcegraph is a fast, solid, full-featured code navigation engine.
-                                    </p>
+                                    <h1 className="text-center">Try Sourcegraph Enterprise free for 30 days</h1>
                                 </div>
                             </div>
                         </div>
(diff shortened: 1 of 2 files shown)


website/src/pages/contact/sales.tsx
```diff
@@ -24,57 +24,56 @@ export default class Trial extends React.Component<any, any> {
...
-                            <div id="hubspotTrialForm" className="d-flex justify-center" />
+                <div>
+                    <div className="jumbotron text-right dark-7">
+                        <div className="row">
+                            <div className="sales py-0 d-flex flex-grow flex-column flex-lg-row align-items-center justify-content-around">
+                                <img className="logo" src="/sourcegraph/sourcegraph-mark.png" alt="logo" />
+                                <div>
+                                    <h1 className="subtitle">Try Sourcegraph Enterprise free for 30 days</h1>
+                                    <p className="text-center">
+                                        Sourcegraph is a fast, solid, full-featured code navigation engine.
```
dadlerj: This subheader seems a bit strange here. Maybe instead make this a link back to the pricing page? Or just remove it, and add a link to pricing to the footer of the section on the left side

website/src/pages/contact/sales.tsx
```diff
@@ -24,57 +24,56 @@ export default class Trial extends React.Component<any, any> {
...
                             </div>
                         </div>
-                    </div>
-                </section>
+                        <div className="form-area">
+                            <div>
+                                <h1>Register for your free trial</h1>
+                            </div>
+                            <div className="form">
+                                <div id="hubspotTrialForm" className="d-flex justify-center" />
```
dadlerj: I can't comment on the form itself, but the mixing of verbs over here, with "Register" in the header, and "Get Started" in the button, seems a bit confusing. What do you think?

website/src/pages/contact/sales.tsx
dadlerj: (and the link to the pricing page can be something like "learn more about products and packaging", not just "pricing", to illustrate that it explains what "Enterprise" actually includes)

KattMingMing: I'll remove it. There's several links to the pricing page here so I don't think it's necessary.

website/src/pages/contact/sales.tsx
KattMingMing: I didn't find it confusing (I've seen other forms use the same verbs) - You put in your information which is registering and then you click get started. If you have any suggestions that you think sound better happy to that instead.

Also, I removed submit because a user has next steps immediately whereas before they only had a confirmation they would get a response.

PR #46, opened 2018-10-25
Simpler browser extension options menu

sqs:


sqs:
Separated this from https://github.com/sourcegraph/about/pull/38

PR #48, opened 2018-10-26
feat: browser extension testing project rough draft

ijsnow:


projects/browser-extension-testing.md
```diff
@@ -0,0 +1,25 @@
...
+- Keep calls to browser extension APIs out of business logic so we can unit
+  test that logic
+- Rearchitect logic integrating with individual code hosts so that we can unit
+  test the integration of code hosts and code intelligence
+
+1. Add more e2e tests
+
+- Run each test on all supported code hosts - code hosts won't be considered
+  fully supported until we're able to run tests on it
+- Test with different feature flags enabled or disabled
```
sqs: We should remove some features and feature flags while we're at it:

- :heavy_check_mark:  inline symbol search has proven stable and not to interfere, so remove feature flag and enable by default /cc @ggilmore (I also would like to extend and promote this feature more, and make it so extensions can contribute to it!) <-- if this delays your work on this project, then you can punt
- :x:  remove GitHub file tree navigation. Some people use this, but it is not the killer feature of our extension. Octotree does it better and does it for private code more easily than us. We also won't be providing this for other code hosts, and I want to minimize the number of features we have that are only for a single code host (it confuses users and makes them feel 2nd-class).
- :x: remove the feature "Open a new window with Sourcegraph search results" because users didn't like it
- keep "Render mermaid.js" diagrams" as a feature flag as is

projects/browser-extension-testing.md
```diff
@@ -0,0 +1,25 @@
+# Browser extension testing and reliability
```
sqs: You intend to work on this immediately after shipping https://github.com/sourcegraph/browser-extensions/pull/271, right? (Do you have an ETA on merging that PR?) That sounds right to me. I already moved the rest of "Better config and depoyment for code host UI integrations" to November 2018 on https://docs.sourcegraph.com/dev/roadmap#november-2018 and split out the options page blog post to https://github.com/sourcegraph/about/pull/46.

ijsnow: Correct. I just have to pull in access token changes that I made the other day into the new options menu.

PR #50, opened 2018-10-30
Redirect root directory to sourcegraph.com/start

francisschmaltz:
Create a redirect from about.sourcegraph.com to sourcegraph.com/start

### DIFF
--- a/website/src/components/Footer.tsx
+++ b/website/src/components/Footer.tsx
@@ -96,22 +96,22 @@ export default class Footer extends React.Component<any, any> {
                                 <input type="checkbox" />
                                 <ul>
                                     <li>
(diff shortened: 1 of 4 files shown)


website/src/components/Footer.tsx
```diff
@@ -96,22 +96,34 @@ export default class Footer extends React.Component<any, any> {
                                 <input type="checkbox" />
                                 <ul>
                                     <li>
-                                        <Link to="/#search-datacenter" onClick={this.codeSearchClicked}>
+                                        <Link
+                                            to="https://sourcegraph.com/start#search-datacenter"
```
attfarhan: do these hashes actually take you anywhere?

francisschmaltz: The will. The will open the modal – future enhance meant

PR #54, opened 2018-11-07
Update pricing page with $10 starting price callout

francisschmaltz:
This adds a callout to the pricing cards. Specifically, it adds a callout that Enterprise Starter is a $10 license for the first 10 users.

screen shot 2018-11-07 at 10 32 26 am


francisschmaltz:
screen shot 2018-11-07 at 12 00 11 pm
@dadlerj updated

PR #62, opened 2018-11-26
docs: add lists section to styleguide

ryan-blunden:


STYLEGUIDE.md
```diff
@@ -92,6 +92,22 @@ Never use any of the following as link text:
 - instructions
 - these instructions
 
+### Lists
+
+#### Introductory sentences
+
+Lists should begin with an introductory sentence. The sentence should end with a colon unless other content, such as an image or admonition is between the introductory sentence and the list.
```
sqs: Comma after admonition

STYLEGUIDE.md
```diff
@@ -92,6 +92,22 @@ Never use any of the following as link text:
...
+
+#### Punctuation and capitalization
+
+Each list item should begin with a capital letter.
```
sqs: What’s the goal of this? Since this is getting into general English style, not Sourcegraph-specific, you should just link to some existing style guide instead of creating our own English style.

ryan-blunden: It's useful for editors in particular. When performing edits on a list, should they be adding periods at the end of sentences if they don't exist? How would they know?

When we can link to an English style guide, then we should remove it but until then, I think this adds value.

sqs: This doesn’t account for the more common (in my experience) way of writing prose lists where the

- sentence continues, and
- no artificial capitalization is introduced. 

Like that. So I think the rule you listed won’t account for all the cases we need and will be weird in many cases. Therefore we should refer to an existing style guide for these things that has fully accounted for all of the cases.

ryan-blunden: Ok, I'll close this.

PR #63, opened 2018-11-28
Update pricing page with Enterprise Starter volume discounts

dadlerj:
Relies on https://github.com/sourcegraph/sourcegraph/pull/1159

@francisschmaltz I ended up removing the subheaders from each tier — I felt like they were too easy to ignore, and the $10 to start for Starter wasn't getting the visibility it deserved. 

Instead, I added it and the new volume discounts bullet to the "features" of the Starter tier.

image


francisschmaltz:
Can we change 10-20% to "Volume" ?

dadlerj:
Yeah I had that at first, but preferred being specific. I'm happy to use volume instead with https://github.com/sourcegraph/sourcegraph/pull/1159/files

PR #66, opened 2018-12-05
Update pricing page

francisschmaltz:
@dadlerj Updated pricing page. Thoughts?

screen shot 2018-12-07 at 11 18 08 am
screen shot 2018-12-07 at 11 18 35 am
screen shot 2018-12-07 at 11 18 52 am

### DIFF
--- a/website/src/css/pages/_pricing.scss
+++ b/website/src/css/pages/_pricing.scss
@@ -99,32 +99,18 @@
     padding: 0 0;
     .pricing-section {
         padding-top: 15vh;
+		.row {
+			justify-content: space-evenly;
+		}
     }
 
     .pricing-card-col {
+
         &:nth-child(2) {
-            transform: translateX(-100%);
-            animation: slide1 650ms ease-in-out 750ms forwards;
-            @keyframes slide1 {
-                from {
-                    transform: matrix3d(0.87, 0, 0.7, 0, 0, 1, 0, 0, -0.5, 0, 0.87, 0, -100%, 0, 0, 1);
-                }
-                to {
-                    transform: matrix3d(1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1);
-                }
-            }
+
         }
         &:nth-child(3) {
-            transform: translateX(-35%);
-            animation: slide2 650ms ease-in-out 750ms forwards;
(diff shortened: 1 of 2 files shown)


francisschmaltz:
> Could you also add "Limited starter package available" to the bottom of the Enterprise column, with a link to the contact form page if you click it?

We already have a free trial button that takes the user to the contact page. We could add copy on there if you wanted.

> Looks like we could make these wider and potentially get "registry" onto the first line :)

I wanted to but it looks odd at a wider frame.

dadlerj:
> We already have a free trial button that takes the user to the contact page. We could add copy on there if you wanted.

Is your reticence to add it because that would cause the number of lines on the two cards to look imbalanced? I'm debating whether to remove one of the other lines if so (or if there's some other way to get "registry" onto a single line :) )

sqs:
- "HA cluster support" - Customers don't care about HA nearly as much as they do about high scalability. I know this is on there now, but we should change it. It should be "High-scale cluster support (Kubernetes)"
- "SSO Support" - "S" in "Support" should be lowercase
- The "for 200 users" is small. I missed it initially even when I was looking for it.

francisschmaltz:
> Is your reticence to add it because that would cause the number of lines on the two cards to look imbalanced?

The lines can always be balanced with more added. I thought you meant adding a link elsewhere, which I thought was redundant. If it's just a block of text then I'm okay with it – it just wouldn't be a link.

> The "for 200 users" is small. I missed it initially even when I was looking for it.

On the site it's still visible enough for someone looking for information for the first time.

dadlerj:
>The lines can always be balanced with more added. I thought you meant adding a link elsewhere, which I thought was redundant. If it's just a block of text then I'm okay with it – it just wouldn't be a link.

Ah, nope! I just wanted to add an extra line to the features list with that text. LMK if that sounds doable.

francisschmaltz:
@dadlerj are you okay having the limited start pack in the sosumi copy below? 
screen shot 2018-12-07 at 11 21 37 am

francisschmaltz:
Mobile screenshots. Changed button to block buttons on mobile.
screen shot 2018-12-07 at 11 53 34 am
screen shot 2018-12-07 at 11 53 40 am
screen shot 2018-12-07 at 11 53 51 am

PR #81, opened 2019-02-11
Add Tomás and Loïc to team page

tsenart:
@nicksnyder: Shall we add this step to our on-boarding process?


dadlerj:
Heh, it is on the onboarding list, but it must have gotten skipped over when you were in town for the holiday event. I want to remove that section regardless, see: https://github.com/sourcegraph/about/pull/82

PR #84, opened 2019-02-20
fix: remove 3.0 from upgrade language

dadlerj:
The current language here (just "or 3.0") makes it sound like all 3.0 patch releases need the docs. That's not true — it's ONLY `3.0.0` (and none of the -beta, etc. ones are included).

There are relatively few `3.0.0` instances left, so for clarity, I think we would be smart to just remove that part entirely.


dadlerj:
Eh, I chickened out on removing this entirely. Will merge this version.

PR #93, opened 2019-03-15
Update core tier user limit

KattMingMing:


dadlerj:
Please also update the changelog and http://docs.sourcegraph.com/adopt/trial

And of course, our emails (e.g. HubSpot, outbound), etc

PR #101, opened 2019-04-04
use yarn instead of npm

sqs:


sqs:
merged as part of #111

PR #103, opened 2019-04-04
clean up unused, unlinted

sqs:
See commit messages.


sqs:
merged as part of #111

PR #113, opened 2019-05-02
Remove extra pair of parenthesies from footer

minikomi:
Removes an extra pair of parenthesies `)}` I noticed in the footer just under the social buttons:

Screen Shot 2019-05-02 at 19 23 17

Looks like they snuck in with https://github.com/sourcegraph/about/pull/112/commits/20317c5aa42b23b694456e2970129d6d16200708 .


sqs:
Thanks!

Issue #125, opened 2019-05-29
Black bar on page RHS when window width less than 820px


ryan-blunden:
A black bar appears on the right of the website if the window width is narrowed past some point. It basically looks like this on my phone (which is where I noticed it), the same issue happens on desktop when making the window smaller.

sqs:
@ryan-blunden Is this problem still occurring?

PR #126, opened 2019-05-30
Temporary fix for blogpost text overflow into black margin

rvantonder:
Temporary but immediate fix for #125:

- Remove fixed widith so text stays in the white area.
- The long code line doesn't break lines, and causes a run over still. I'm manually breaking it up so that it stays within width. This will look reasonable on mobile devices:

Screen Shot 2019-05-30 at 7 23 44 PM

There is a minor ugliness due to the space in the long codeline in the larger-width desktop view:

Screen Shot 2019-05-30 at 7 23 54 PM


rvantonder:
There's still some overflow on mobile view due to the long `https://docs.aws.amazon...` link not breaking, but the the post is readable so I'm going ahead with the merge.

PR #130, opened 2019-06-04
Added Vimeo component and new code review video

ryan-blunden:


ryan-blunden:
> Nice! Is the id the only required parameter (is it even required, should it be)?

Yeah, it's the id of the video to embed.

PR #132, opened 2019-06-06
Added video for tracking package use and version management

ryan-blunden:
Video at https://vimeo.com/340774081

### DIFF
--- a/website/src/pages/product/code-search-navigation.tsx
+++ b/website/src/pages/product/code-search-navigation.tsx
@@ -357,6 +357,8 @@ export default ((props: any) => (
                     Use Sourcegraph to find answers in code, faster!
                 </h2>
                 <Vimeo id={340491490} />
+                <p>&nbsp;</p>
+                <Vimeo id={340774081} />
             </ContentSection>
             <ContentSection color="purple" className="py-6">
                 <h2 className="text-center display-4 pb-4">Demo: Sourcegraph code navigation</h2>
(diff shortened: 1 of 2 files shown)


website/src/pages/product/code-search-navigation.tsx
```diff
@@ -357,6 +357,8 @@ export default ((props: any) => (
                     Use Sourcegraph to find answers in code, faster!
                 </h2>
                 <Vimeo id={340491490} />
+                <p>&nbsp;</p>
```
christinaforney: Why is this needed? Would it be better to add a margin or padding to a css class on the component? This is adding a second video to the black section? Do we need to introduce/title the two videos separately? Or we allow the splash page of the video to be that title (thinking about SEO...)?

christinaforney:
Video looks good - I like the pauses you took to let the user read the message before performing the action.

website/src/pages/product/code-search-navigation.tsx
ryan-blunden: My latest commit fixes the hacky spacing code. Just in a rush.

This is not how I want to be long term, but for now, the videos are grouped under the **Use Sourcegraph to find answers in code, faster!** heading.

The video splash page is acting as a title for now. SEO is definitely a consideration but isn't a priority right now.,

christinaforney: Sounds good!

PR #137, opened 2019-06-21
Set a max width of 740px for blog post content column

ryan-blunden:
Based on the max width of medium.com's content column.

ScreenFlow

### DIFF
--- a/website/src/css/pages/_blog-post.scss
+++ b/website/src/css/pages/_blog-post.scss
@@ -8,6 +8,7 @@
     &__wrapper {
         width: 66.667%;
         margin: auto;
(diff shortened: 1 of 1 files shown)


ryan-blunden:
On mobile, the 3.4 blog post is still problematic because of the size of the code blocks but 3.5 is post fine.

ryan-blunden:
The 3.3 post still has a black bar but readability is not affected.

Screen Shot 2019-06-21 at 11 09 40 AM

ryan-blunden:
3.4 is the same as 3.3, in that the black bar is there but readability is not affected.

ScreenFlow

PR #138, opened 2019-06-22
Thorn case study and case study components

ryan-blunden:
ScreenFlow


christinaforney:
Where is this linked from in the navigation? I had to look at the file path to get to it. Will we be building a "case studies" top level page?

There are some title overflow issues on mobile. Perhaps use a media query to put image on top and quote below rather than keeping them side by side on mobile sizing:
image

The pdf you sent out had quite different formatting - is the intention for this to be a more simplified version of that?

ryan-blunden:
> Where is this linked from in the navigation? I had to look at the file path to get to it. Will we be building a "case studies" top level page?

It's not linked to from anywhere right now. First priority is to get this publicly accessible so we can share a link to it. Blog post will follow shortly and yeah, case studies index page is in the works.

> There are some title overflow issues on mobile.
Mobile styling will be fixed before merging.

> The pdf you sent out had quite different formatting - is the intention for this to be a more simplified version of that?

Yes, for now. Let each medium play to their strengths.

PR #141, opened 2019-07-01
add missing space and period

vanesa:
Noticed that there was a space and a period missing in the case study. Here are the fixes.


ryan-blunden:
I'll fix this along with other content changes in a new PR. Thanks for reporting!

PR #142, opened 2019-07-02
Thorn case study edits

ryan-blunden:
Home page changes

Screen Shot 2019-07-02 at 3 30 07 PM

Case study page changes

ScreenFlow


ryan-blunden:
@sqs: I've changed the copy in the CTA footer (the description text is from Adam's changes) and would like your input there specifically.

PR #176, opened 2019-07-26
add content for How Deep Do You Go talk

ijt:


ryan-blunden:
Thanks @ijt. 

You need to replace the `$LIVEBLOGGER` placeholders and set `published` to true.

ijt:
Done, PTAL.

PR #207, opened 2019-08-01
Add Convoy logo

ryan-blunden:
New style prevents logos wrapping beyond 1200px wide so we don't have the first row with 8 logos, and the second row with one. 

We may remove this once we have enough logos to fill both rows, presuming that we want two rows of logs on the home page.

ScreenFlow


ryan-blunden:
adjusted-styles

PR #209, opened 2019-08-02
Update pricing page to properly reflect search notificaitons

christinaforney:
We had a customer ask about notifications for saved searches and realized we should be more clear about what is available vs. not since we had to deprecate the Slack integration.


dadlerj:
Thanks!

PR #210, opened 2019-08-05
Better looking code snippets

ryan-blunden:
Adds padding to code blocks and changes to light text on a dark background, while inline code blocks remain the same.

Screen Shot 2019-08-05 at 4 19 54 PM


ryan-blunden:
> Seems fine. Is there any way we could get syntax highlighting into our Markdown? It makes a huge difference in readability, but I'm not sure what it would take to add that to our site.

I'll add this as an issue as yeah, it needs improving.

PR #218, opened 2019-08-13
Add pricing tier, update pricing basis from user accounts to MAUs

dadlerj:
Updated pricing:
1) Adds a new tier ("Enterprise Plus")
2) Converts all pricing from user-account-based to MAU-based
3) Converts all pricing from annual/paid-up-front to monthly subscriptions

image
image
image
image
image
image

### DIFF
--- a/website/src/pages/pricing.tsx
+++ b/website/src/pages/pricing.tsx
@@ -38,7 +38,7 @@ export default ((props: any) => (
                     <h4 className="font-weight-light">
                         Sourcegraph code search, code navigation,&nbsp;
                         <PricingFreeTierPopoverButton />
-                        are free for individuals and small teams (up to 20 users). <a href="https://docs.sourcegraph.com">Install</a> to get started today!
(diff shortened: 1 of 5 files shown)


website/src/pages/pricing.tsx
```diff
@@ -30,80 +34,71 @@ export default ((props: any) => (
...
+                                name="Enterprise"
+                                description="Enabling engineering and DevOps leaders to speed up the organization's software development and monitor risks."
+                                price="$29"
                                 features={[
-                                    { name: 'Code search', id: 'code-search' },
-                                    { name: 'Code navigation (definitions & references)', id: 'code-navigation' },
-                                    { name: 'Sourcegraph extensions', id: 'integrations' },
-                                    { name: 'Code host & editor integration', id: 'integrations' },
+                                    { name: 'Code review & pull request integration', id: 'code-review' },
+                                    { name: 'Code alerts & automation', id: 'code-alerts-automation' },
```
christinaforney: We should be careful about using 'automation' here if we are going to use that as the name for large scale code mod which we want in the top tier.

PR #234, opened 2019-09-13
Add blog post for NYT Crossword Session

aligg:


ryan-blunden:
@aligg Woo! First liveblog post for Strange Loop. This is excellent and I appreciate the detail.

Only thing I ask you change is to reduce the size of the images. Easiest way is to reduce the width to 740px, then save the quality between 60%-80%.

PR #240, opened 2019-09-13
add uptime 15364 days the computers of voyager

signalnine:


ryan-blunden:
Hey @signalnine, thanks for the great liveblog content.

It seems you've duplicated your content though, so fix that and we're good to go!

ryan-blunden:
Thanks for the fantastic liveblog content @signalnine

PR #247, opened 2019-09-14
Add post for apache-pulsar and telegraphy talk at strangeloop

sebasmurphy:
Post for Strangeloop talk `Strange Loop 2019 - Apache Pulsar and the Legacy of Telegraphy`

No slides|speak items were provided, so it's only text.


ryan-blunden:
Thanks @sebasmurphy for a great liveblog post

PR #251, opened 2019-09-15
strange loop liveblogs: ML for lipreading and ML for sign language

mchua:


ryan-blunden:
These are great! Thanks @mchua!

PR #262, opened 2019-09-17
Adding the Rhapsody in Zero Knowledge Proofs liveblog

prasincs:
The PR tries to wrap up the key points in the talk in a somewhat hierarchical and chronological manner. The order is close, but given the highly mathematical and domain specific nature of the talk, it's hard to present it without either making the text too long or changing the nature of the presentation. This likely needs another round of proofreading before it's ready to publish.


ryan-blunden:
Thanks for a great liveblog post @prasincs!

PR #263, opened 2019-09-17
Adding Strange Loop 2019 - RGB to XYZ: The Science and History of Color Live Blog

brittanyrw:


ryan-blunden:
@brittanyrw this is an awesome post! Thank you for liveblogging!

PR #288, opened 2019-10-21
Add "Edit this post" button

ryan-blunden:


website/src/pages/blog.tsx
```diff
@@ -69,6 +69,7 @@ export const pageQuery = graphql`
                     fields {
                         slug
                     }
+                    fileAbsolutePath
```
attfarhan: Why do you need to request `fileAbsolutePath` this in this file?

ryan-blunden: This was the easiest way to derive the blog post file name without adding a new field

ryan-blunden: Good catch, yeah, not needed.

PR #303, opened 2019-11-06
Added the Yelp tweet to the home page

ryan-blunden:
Screen Shot 2019-11-06 at 10 26 10 AM


dadlerj:
This looks nice, but I feel like we should treat the blog post as a case study, highlighted at the very top of the page, rather than buried in the tweet carousel. It'd be just like the Quantcast/Thorn case studies, but it would go to the Yelp eng blog page instead of a local one. What do you think?

vanesa:
I agree with what Dan, despite feeling a little wary of driving people out of our homepage. Can we host a copy of the blog post on our website?
Also: If you want to additionally add the tweet to the carousel, I would position in the middle, as it attracts the attention of the user more if it looks symmetrical.
image

dadlerj:
@ryan-blunden did you see my message?

ryan-blunden:
This is coming with the home page carousel.

PR #315, opened 2019-11-09
document engineering onboarding

sqs:
The handbook should document the steps to onboard an engineer. These can be just the eng-specific steps for now (we will add general onboarding steps in `handbook/people-ops/` soon and can cross-link them later).

Inspired by https://github.com/sourcegraph/infrastructure/pull/1723 (which should be closed and its contents, or some subset, moved here).


ryanslade:
I think @nicksnyder's comment in https://github.com/sourcegraph/infrastructure/pull/1723 is a great idea, pasted below:

Maybe all we need from an onboarding perspective (and which would be ok to document in our handbook) would be a set of onboarding "challenges/tasks" without providing the direct answers. e.g. "Make yourself an admin on sourcegraph.com and sgdev.org by updating the database directly. Here are some relevant doc links: [two links from above]". If the above documentation needs updates or could be clearer, then we should do that too.

sqs:
@nicksnyder friendly ping on this

nicksnyder:
Yep, this is on my todo list, just near the bottom. It will definitely bubble up when we get close to the next engineering hire start date if I haven't done it already by then :)

nicksnyder:
Going to open this as a fresh PR.

PR #320, opened 2019-11-10
enforce Slack 90-day message retention to encourage better documentation

sqs:
Discussions that occur on Slack (about a source-of-truth doc like an RFC, issue, PR, HubSpot record, etc.) are not discoverable by people looking at the source-of-truth doc. This causes people to:

- miss important context (because they have no way to know that discussion occurred on Slack)
- duplicate discussions and effort (e.g., often one person asks a question on Slack and someone else asks the same question on the RFC gdoc)
- consider the source-of-truth doc as not-quite-canonical (and trust it less, or invest less time in it) because they know there's likely crucial info scattered on Slack, too

To avoid this, I am proposing to configure our Slack workspace to delete messages in public after 90 days (in [Slack workspace settings > Data retention](https://sourcegraph.slack.com/admin/settings#data_retention)), and to communicate this to the team in the handbook. This will be a forcing function to everyone to update the source of truth after any significant Slack discussions.

Private channels and direct messages will remain "Keep all messages forever" because messages in those channels rarely pertain to source-of-truth docs anyway.

(GitLab has the [same policy](https://about.gitlab.com/handbook/communication/#slack).)

### DIFF
--- a/handbook/communication/index.md
+++ b/handbook/communication/index.md
@@ -64,9 +64,10 @@ See [example of deprecated Google Doc](https://docs.google.com/document/d/1M22s-
 
 ## Slack
 
-We use Slack for team chat, which means ephemeral conversations. Any important decisions or changes need to be reflected in the [source of truth](#sources-of-truth).
+We use Slack for team chat, which means ephemeral conversations. To enforce this, only 90 days of Slack activity in public channels is retained. Any important decisions or changes need to be reflected in the [source of truth](#sources-of-truth).
 
-See [common channels](team_chat.md).
+- [Common channels](team_chat.md)
+- [Historical archive of Slack messages prior to 2019-11-09](https://drive.google.com/file/d/1FUbOEsMM4fWRpxymgNHZCAssOPEFDelJ/view?usp=sharing): unzip and open `sourcegraph-slack-archive-to-20191109/index.html` to view. If you consult this, please be sure to add whatever information you learn to a source of truth so we reduce our reliance on this archive.
 
 ### Avoid private messages
 


slimsag:
Is there a way to archive prior discussions for searching somewhere?

Going forward, this policy certainly makes sense and knowing it is in place will definitely be a forcing function and I am in favor of it -- but that doesn't change the fact we have 3+yr old threads of value that I search back on and reference fairly frequently. If there's no way to avoid losing that.. I guess moving forward on this is still good, but this will be quite bad in many instances for me ("I am positive there was a discussion about exactly this topic but it's gone so I can't help you")

sqs:
@slimsag Yes, here is what I propose. When we make this policy effective, we will make a [*one-time* export of all Slack data from public channels](https://sourcegraph.slack.com/services/export) as of the current date and store it on a GCS bucket. Whoever does this will then either set up https://github.com/hfaran/slack-export-viewer or document how to do so locally for anyone who needs to search the archives. Each time someone searches the archives, they should be extra diligent about moving the information they learn into a source-of-truth doc somewhere.

slimsag:
Perfect!

sqs:
I made the dump of historical Slack messages in public channels and added the Google Drive link and instructions to this PR.

sqs:
> > Private channels and direct messages will remain "Keep all messages forever" because messages in those channels rarely pertain to source-of-truth docs anyway.

> Is this a "it isn't necessary" argument or is it a "this would be bad to do" argument. It seems like the former. One risk is that we create a perverse incentive for people to use private channels exactly for creating a searchable record in Slack. I propose enforcing this for private messages too.

"It isn't necessary." Private channels aren't a good searchable record because only a select few people would be able to find results in them. I don't think that (eg) the sales team will create a private Slack channel just so they have archives. That would cause friction when they need help from anyone else on the team, and it would obviously violate the spirit of the rule. If this happens, we can deal with it. :)

sqs:
@dadlerj Please review

sqs:
Enacted: 
image

PR #321, opened 2019-11-10
add guidelines for 1-1s

sqs:


sqs:
@nicksnyder I am going to merge this without review because it doesn't contain anything new except:

> To avoid diluting the meaning of and expectations for 1-1 meetings, avoid referring to a meeting that is not between a manager and direct report as a "1-1" meeting.

which I think you will agree with or at least be OK with.

nicksnyder:
lgtm

PR #327, opened 2019-11-11
WIP: document where customer contracts are stored (Dropbox)

sqs:
This is blocked on https://github.com/sourcegraph/about/pulls, which adds some more onboarding documentation, because we also need to include Dropbox account setup as a step in onboarding.

### DIFF
--- a/handbook/sales/index.md
+++ b/handbook/sales/index.md
@@ -3,6 +3,7 @@
 The Sales team represents us and our values to customers, bringing back dollars and feedback to help us grow.
 
 - [Onboarding](onboarding.md)
+- [Customer contracts in Dropbox](TODO)
 
 ## [Customer support](../support/index.md)
 


sqs:
@dadlerj Why do we store customer contracts on Dropbox not Google Drive? Can we switch to Google Drive? I can't seem to find my Dropbox credentials right now, which is a problem that I would not have if we just used Google Drive.

dadlerj:
No good reason, that's just where they've started and I've never felt the pain you're feeling. I do think we need an app that has desktop support (so I'll have to check GDrive's Mac client again, since it's been years), but I'm perfectly happy to switch otherwise.

sqs:
@dadlerj Can you please either add creating/linking a Dropbox account as a step to the handbook (and add my Dropbox account sqs@sourcegraph.com), or switch us from Dropbox to Google Drive? Also add a link to the top of the sales handbook section to the private Dropbox or Google Drive folder where customer contracts live.

dadlerj:
Closed in favor of https://github.com/sourcegraph/about/issues/346

PR #331, opened 2019-11-11
Simplify handbook index

nicksnyder:
Subpoints should either all be inlined or be removed entirely. I think the latter is better to avoid duplication (unless you want to keep it updated automatically somehow).

I did not audit the non-engineering sections.


sqs:
Your call for eng section. I think having the most frequently viewed pages in the index is useful personally.

PR #333, opened 2019-11-12
Add marketing team communication docs

dadlerj:


dadlerj:
@sqs I checked the [handbook guidelines](https://about.sourcegraph.com/handbook/usage#handbook-guidelines) but didn't see any hard rules on when a section deserves its own page (versus a section on an existing page). Curious if there's a preference here?

sqs:
@dadlerj i don’t have a strong feeling here either yet. If a preference emerges we can document it. Whatever works for now.

PR #340, opened 2019-11-13
Add link to GitLab's blog post

dadlerj:
Helpful to have this? We can move it down, but I think it's nice to make this discoverable.

image


christinaforney:
Thanks for adding this! I was thinking we should link to it in the second sentence of the first paragraph:

"Today, [GitLab announced its collaboration with Sourcegraph]() to provide..."

Then it doesn't take them away from the page before they get to the video. What do you think? I can push that change if you want to go with it.

dadlerj:
That's SO much better, good call!

christinaforney:
> That's SO much better, good call!

Yep! I worded that sentence intentionally for this exact purpose :)

PR #347, opened 2019-11-13
transition from remote-first to all-remote

sqs:
Proposes transitioning Sourcegraph from a remote-first company to an all-remote company. We briefly discussed this at company meeting on 2019-11-11 and then [on Slack in #general](https://sourcegraph.slack.com/archives/C02FSM7DU/p1573498881167500). This is the actual proposal to change this aspect of Sourcegraph (in line with [Changing or defining a process](https://about.sourcegraph.com/handbook/usage#changing-or-defining-a-process) in the handbook).

Remote work has been a part of Sourcegraph since very early on, and it has worked extremely well. More than half of our team is remote now, and we anticipate the proportion of remote teammates will grow because the population of talented people in SF is tiny compared to the population of talented people worldwide.

We transitioned from remote-friendly to remote-first earlier this year. That switch was successful. Several teammates who otherwise would have felt obligated to work from our SF office instead chose to move to a location they favor over SF or work from home despite being based in the SF Bay Area. Those people are happier and more productive because of it. The switch also helped us attract new talented remote candidates who prefer remote-first companies to remote-friendly companies (which are hit-or-miss, sometimes they claim to be remote-friendly but without the forcing function of remote-first let their communication practices privilege local folks).

Now that we are remote-first, why go all-remote?

1. Having an office causes many SF-based people to feel obligated to come into the office more than is optimal. They tell us they do this because they "don't want the office to feel empty" despite reporting that they would feel more productive and happier working from home more of those days. This is an admirable impulse, but we don't want people to feel guilty about not coming into the office and therefore choose a non-optimal arrangement.
   - Could we just *really* clearly communicate to these people that they aren't obligated to come into the office? We've tried that, but it hasn't changed their behavior (they still feel obligated, and people still feel bad that the office is empty).
   - Could we just get a smaller official office location? We could try that, and it'd probably make the situation better but not completely fix it. For people who like having an office, we instead suggest they coordinate working out of a private office(s) in a coworking space with other like-minded teammates, and we will reimburse that. That yields a very similar outcome to the people who prefer an office without affecting other teammates.
1. All-remote is more appealing to talented people worldwide than remote-first because it means they are on a level playing field with everyone else at the company (and not at a slight, possibly in perception only, disadvantage to local teammates). Being all-remote will help us continue to grow our team with world-class teammates.

What about people who prefer having an office with other teammates? Some SF-based teammates expressed this. See the "Could we just get a smaller office?" point above. You can coordinate with other SF-based teammates to get a private office(s) at a coworking space, and we will reimburse that. (The same, of course, would apply to any teammate or group of teammates anywhere in the world.)

Notes:

- We're going to document a lot more of our practices and policies for remote work, such as how new teammate onboarding changes when everyone is remote. These will come in future edits to the handbook.
- Now that we're all remote, we don't need to use the term "distributed" in preference to "remote".

### DIFF
--- a/handbook/communication/index.md
+++ b/handbook/communication/index.md
@@ -1,6 +1,6 @@
 # Communication
 
-We're an [all-remote](../../company/remote/index.md) company, with teammates from all around the world and (soon) no primary office. To make this work, we 
+We're an [all-remote](../../company/remote/index.md) company, with teammates from all around the world and (soon) no primary office. To make this work, we need to be deliberate about how we communicate.
 
 ## Sources of truth
 
--- a/handbook/communication/index.md
+++ b/handbook/communication/index.md
@@ -1,5 +1,7 @@
 # Communication
 
+We're an [all-remote](../../company/remote/index.md) company, with teammates from all around the world and (soon) no primary office. To make this work, we 
+
 ## Sources of truth
 
 These places are the source of truth for information at Sourcegraph. Information in these places is expected to be accurate and up-to-date:
@@ -94,3 +96,5 @@ Most meetings at Sourcegraph are video calls. We prefer [Zoom](https://zoom.us) 
 ## Writing
 
 1. Always use [ISO dates](https://en.wikipedia.org/wiki/ISO_8601#Calendar_dates) in all writing and legal documents because other formats [lead to online confusion](http://xkcd.com/1179/). Use `yyyy-mm-dd`, for example 2020-04-13, and never 04-13-2020, 13-04-2020, 2020/04/13, nor April 13, 2020. Even if you use an unambiguous alternative format, it is still harder to search for a date, sort on a date, and for other team members to know we use the ISO standard. For months use `yyyy-mm`, so 2020-01 for January 2020.
+
+
--- a/handbook/people-ops/index.md
+++ b/handbook/people-ops/index.md
@@ -9,7 +9,7 @@
 - [Spending company money](spending-company-money.md)
 - [Glossary](from-graphbook/glossary.md)
 - [Holidays](from-graphbook/holidays.md)
-- [Remote work](../../company/remote/index.md)
+- [All-remote work](../../company/remote/index.md)
 - [Teammate philosophy](teammate_philosophy.md)
 - [List of team members](../../company/team/index.md)
 
--- a/company/index.md
+++ b/company/index.md
@@ -2,3 +2,6 @@
 
 - [**about.sourcegraph.com**](https://about.sourcegraph.com) for information about Sourcegraph (the product and the company)
 - [Sourcegraph handbook](../handbook/index.md#company)
+- [Objectives and Key Results (OKRS)](okrs/index.md)
+- [All-remote](remote/index.md)
+- [Open source, open company](open_source_open_company.md)
--- a/company/remote/index.md
+++ b/company/remote/index.md
@@ -1,5 +1,28 @@
-# Remote
+# All-remote
 
-Sourcegraph is a remote-first company.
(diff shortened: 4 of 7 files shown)


handbook/communication/index.md
```diff
@@ -1,5 +1,7 @@
 # Communication
 
+We're an [all-remote](../../company/remote/index.md) company, with teammates from all around the world and (soon) no primary office. To make this work, we
```
nicksnyder: Incomplete sentence

handbook/people-ops/index.md
```diff
@@ -2,7 +2,7 @@
 
 - [Travel](travel.md)
 - [Onboarding](onboarding/index.md)
-- [Distributed work tips](https://github.com/sourcegraph/Graphbook/blob/master/Onboarding/Distributed%20work%20tips.md)
+- [Remote work tips](https://github.com/sourcegraph/Graphbook/blob/master/Onboarding/Distributed%20work%20tips.md)
```
nicksnyder: Rename the file too

handbook/people-ops/onboarding/index.md
```diff
@@ -2,8 +2,8 @@
 
 Some [onboarding docs](https://github.com/sourcegraph/Graphbook/tree/master/Onboarding) are in the Graphbook (which are in the process of being migrated to this handbook):
 
-- [Distributed teammates (US-based)](https://github.com/sourcegraph/Graphbook/blob/master/Onboarding/Distributed%20teammates%20(US%20based).md)
-- [Distributed teammates (non-US-based)](https://github.com/sourcegraph/Graphbook/blob/master/Onboarding/Distributed%20teammates%20(international).md)
+- [Remote teammates (US-based)](https://github.com/sourcegraph/Graphbook/blob/master/Onboarding/Distributed%20teammates%20(US%20based).md)
```
nicksnyder: Rename the file too

handbook/people-ops/onboarding/index.md
```diff
@@ -2,8 +2,8 @@
...
+- [Remote teammates (non-US-based)](https://github.com/sourcegraph/Graphbook/blob/master/Onboarding/Distributed%20teammates%20(international).md)
```
nicksnyder: Rename the file too

handbook/people-ops/index.md
sqs: @mercadon is moving these files into the handbook anyway very soon so not worth it.

nicksnyder: Ah, right, I missed that these weren't already in the handbook.

attfarhan:
I am curious about how perks and benefits are affected once we are fully remote. Some perks (food, transport budget) are currently not provided for off-site teammates. Will these benefits be removed completely? 

It was mentioned that there will be a budget for co-working spaces for teammates who prefer working outside of home, and in this case, a kitchen is not easily accessible, and there is still some transport involved, so I could see the case for still making these benefits available.

dadlerj:
@attfarhan yep, benefits would change to reflect that there no longer is an option that teammates can opt into that provides office/food/etc. We haven't collected quite enough data yet (on, e.g., market rates for coworking spaces) to finalize it, but benefits would become less location-dependent in this world.

beyang:
Regarding the food/perks question, what about rolling all work-related expenditures into some sort of standard stipend (or even just rolling this into salary)? I think it will be difficult/complex to describe a reimbursement policy for every individual meal. If there's a tax benefit to listing this as a perk, I'd favor that. Otherwise, I'd favor just bumping everyone's cash comp by a certain amount. What do folks think about that?

sqs:
As discussed at company meeting, we will move forward with this. There are still many details to figure out, but none that block this PR (which documents that we are transitioning to all-remote, not that we are currently all-remote).

PR #354, opened 2019-11-16
Fix company meeting time

unknwon:
The time is not the same as [the link it points to](https://about.sourcegraph.com/handbook/communication/company_meeting), nor does it reflect the current truth.

Not sure if this is intended but at least we have inconsistency in the handbook.


sqs:
Thanks!

PR #362, opened 2019-11-19
chore: add prettier-check command to force formatting

attfarhan:
Adds a `yarn prettier-check` command, and uses it instead of the current `yarn prettier` command in the build process to force adherence to prettier formatting.


sqs:
@attfarhan Can you please resolve the merge conflicts?

PR #369, opened 2019-11-20
fix: fix homepage js issues

dadlerj:


website/src/components/TestimonialCarousel.tsx
```diff
@@ -45,7 +45,7 @@ export const TestimonialCarousel: React.FunctionComponent<Props> = ({
                     <img src="/case-studies/quote.svg" className="testimonial-carousel__quote-icon mb-5" alt="" />
                     <Carousel controls={false}>
                         {testimonials.map(({ customer, logo, quote, author, cta }, i) => (
-                            <div className={`${slugify(customer).toLowerCase()} testimonial-carousel__testimonial`}>
+                            <div key={`testimonial-${i}`} className={`${slugify(customer).toLowerCase()} testimonial-carousel__testimonial`}>
```
ryan-blunden: What is the intention behind adding `key` here?

dadlerj: I believe react likes every element generated inside of loops to have unique keys (I get a console error when viewing this page locally). See https://reactjs.org/docs/lists-and-keys.html#keys

ryan-blunden: Good to know, thanks!

PR #365, opened 2019-11-20
Analytics additions

ebrodymoore:
Added hubspot data quality to 'sales' and created bizops/analytics documentation

PR #393, opened 2019-11-26
clean up BizOps section

sqs:
- Make the main BizOps handbook page consistent with the other main handbook pages for teams. It needs to explain what the team is.
- Fix links from the sales handbook pages for people to contact BizOps. Previously, these used incorrect URLs (which is a simple problem, https://github.com/sourcegraph/about/pull/365#issuecomment-558516775) and they referred to a GitHub team `@sourcegraph/bizops`. That is not a helpful way to tell people how to contact BizOps in general. It would only work for mentioning the BizOps team on GitHub issues/PRs (which is not the context in which the sales pages ask people to contact BizOps). I changed this to link to the main BizOps handbook page, which includes instructions on how to contact the BizOps team.
- Fix Markdown headers (see https://github.com/sourcegraph/about/pull/365/files#r350595870).

@ebrodymoore:

- [ ] There are 2 TODOs in this PR that you should address
- [ ] In general, we like to avoid acronyms/jargon ([I will document this in the handbook](https://github.com/sourcegraph/about/pull/394)). This would suggest calling the team "Business Ops" and the folder `business-ops`, not "BizOps" and `bizops`, respectively. If this is OK with you and doesn't change the meaning of the term in your opinion, please make this change. Otherwise, it is fine to leave it as-is.

### DIFF
--- a/handbook/bizops/index.md
+++ b/handbook/bizops/index.md
@@ -1,8 +1,14 @@
-### Analytics
+# BizOps
+
+The BizOps (Business Operations) team is responsible for <!-- TODO(ericbm): add brief summary for other teammates who don't know what BizOps means -->.
+
+To reach us, mention `@ericbm` in #sales or #marketing. <!-- TODO(ericbm): or create a new channel, or whatever you prefer -->
+
+## Analytics
 
 This page describes Sourcegraph's analytics function, our data sources, and how to use our data tools.
 
-## How to submit a data request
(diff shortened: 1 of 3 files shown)


sqs:
@ebrodymoore I am going to merge this now to fix the broken link errors (that also affect all other PRs to this repo), but please follow up on the 2 checkboxes.

PR #402, opened 2019-11-26
How to create a license key

chrisdelane:
Adding a section to define the next step after a deal is closed-won.


nicksnyder:
@dadlerj @juliasourceress what do you want to do with this PR?

dadlerj:
I'll own

PR #403, opened 2019-11-26
remove retrospective from release captain duties

beyang:
Companion of https://github.com/sourcegraph/sourcegraph/pull/6865


nicksnyder:
Should the release captain send a message that prompts/indicates that each team should do their own retrospective?

beyang:
> Should the release captain send a message that prompts/indicates that each team should do their own retrospective?

If each team is responsible for their own retrospective, would it not be better to leave them to handle it? Or is this something you want to standardize across all eng teams?

nicksnyder:
I do want teams to do retros, but release process probably isn't the right venue to enforce that anyway.

PR #411, opened 2019-11-27
engineering/commit_messages: Fix table indentation

rvantonder:


rvantonder:
The preview isn't accurate (it can't possibly have a row with just the three `---` so I can't see that this actually fixes the problem.

Screen Shot 2019-11-27 at 3 20 32 PM

PR #414, opened 2019-11-28
sales interview: add mock customer call info

sqs:
This describes (for us and for candidates) how we run mock customer calls during sales interviews.


sqs:
@chrisdelane asked:

> Will the mock customer call be for the candidate to complete to understand how they would approach a call? Or, are we doing it for the candidate as a dry run?

The former. It is so we can understand how they would run a call.

chrisdelane:
Thank you for posting this.

PR #419, opened 2019-12-02
Update Q4 Web OKRs outcome

lguychard:

PR #420, opened 2019-12-02
Separate KRs and KR results with full stops.

nicksnyder:
This makes them more readable.

See https://github.com/sourcegraph/about/pull/419#discussion_r352903788 for an example.

If approved I would update all existing OKRs to the new format before merging.

### DIFF
--- a/company/okrs/index.md
+++ b/company/okrs/index.md
@@ -44,18 +44,18 @@ We set OKRs for each quarter. The (rough) schedule for setting OKRs for an upcom
 
 ### Format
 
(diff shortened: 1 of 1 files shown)


christinaforney:
I just updated the product section in the Q4 OKRs to reflect a change that GitLab has adopted: https://github.com/sourcegraph/about/pull/421 I would suggest we adopt this format as well.

nicksnyder:
Even better! I think we should update our documentation and OKRs in a single commit (either my PR or yours).

christinaforney:
I can add it to mine if you like this strategy!

nicksnyder:
Go for it. I'll close this PR.

PR #421, opened 2019-12-02
Updating product OKRs to use new format and add current results

christinaforney:
GitLab changed their [format for OKRs](https://about.gitlab.com/company/okrs/#format) slightly to improve readability - I updated the product section to reflect that change as well.


nicksnyder:
This format is much easier to read! Can you update the handbook docs and (ideally) all existing KRs to follow this format too?

PR #422, opened 2019-12-02
2020-Q1 OKRs

christinaforney:
This is a DRAFT PR for our team to iterate on to start thinking about and building out OKRs for 2020.

All OKRs are subject to change until merged.


company/okrs/2020_q1.md
```diff
@@ -0,0 +1,69 @@
...
+
+1. **VP Marketing: Deliver sufficient MQLs to sales to reach [$N][N] net new ARR goal.**
+1. **VP Marketing: Create new category of developer tool.**
+1. **VP Marketing: Become the dominant provider of source code search.**
+
+## Product
+
+1. **Product: WIP Deliver on product roadmap**
+   1. Product KR: Roadmap communicates rough timeline for upcoming work.
+   1. Product KR: Build out per-team strategy RFCs to buffer roadmap.
```
sqs: @christinaforney 

- For per-team strategies, it would be nice for these to be more permanent and canonical. Can you add them to the handbook on the page for each team? For example, https://about.sourcegraph.com/handbook/engineering/distribution already has a vision statement, so putting a strategy RFC out would run the risk of diverging from that and from the linked Google Doc.

Also, the 2nd 2 KRs here are things to do, not results. Please reword them to describe an outcome (eg "Each team has ..." and "Each feature is ...").

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,69 @@
...
+   1. Product KR: Dogfooding each feature release to ensure quality UX.
+1. **Product: WIP Reduce product friction**
+   1. Product KR: Reduce time to "aha" (0 to search).
```
sqs: @christinaforney How are you defining time-to-aha? Either link to a dashboard (if available yet) or a doc describing how this will be measured. And is time-to-aha the same as "0-to-search" or are those 2 different things?

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,69 @@
...
+   1. Product KR: Reduce time 0 to deploy.
```
sqs: @christinaforney How are you defining time-to-deploy? Either link to a dashboard (if available yet) or a doc describing how this will be measured.

And how does this differ from the previous point (time-to-aha)? Is one for users and one for site admins?

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,69 @@
...
+   1. Product KR: Improve out of the box experience.
```
sqs: @christinaforney This is not a KR. It describes work to be done but not an outcome/result.

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,69 @@
...
+1. **Product: WIP Grow product team**
+   1. Product KR: Hire to fill position gaps (1 UX Designer, 1 PM).
+   1. Product KR: Create handbook section outlining product strategy and operation guide.
```
sqs: @christinaforney To be worded as a KR:

```suggestion
   1. Product KR: Handbook describes product strategy and processes.
```

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,69 @@
...
+   1. Product KR: Create scalable customer feature prioritization.
```
sqs: @christinaforney What does this mean? Also, reword as a KR and include the problem it is solving, such as "New product team members can understand and contribute to customer feature prioritization".

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,69 @@
...
+1. **VP Eng: Build a world class engineering team.**
+   1. VP Eng KR: Grow team to plan (+1 code intel, +2 distribution, +2 web, +1 core services).
+   1. VP Eng KR: Transition an existing teammate to be (or hire) an Engineering Manager of the distribution team.
+   1. VP Eng KR: Document how we do performance reviews and determine compensation for teammates in engineering roles.
+1. **VP Eng: Create a data driven planning process.**
+   1. VP Eng KR: Each item in a tracking issue has an estimate of time or size.
+   1. VP Eng KR: Each team commits to a written work plan for each release cycle that accounts for the work estimates and the team's capacity.
+   1. VP Eng KR: Each team completes >80% of the work that they committed to at the beginning of the release cycle each iteration.
+
+1. **Distribution: Resolve issues for critical customer deals and expansions**
```
sqs: @beyang What is at least 1 KR for this? How will we know if we succeeded?

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,69 @@
...
+1. **Distribution: Improve developer productivity**
+   1. Distribution KR: Lower build times to X
+   1. Distribution KR: Automate away release captain role
+   1. Distribution KR: Infrastructure for automated deploys and releases (no human in loop required by default)
+1. **Distribution: Build features that accelerate customers through pipeline**
```
sqs: @christinaforney @beyang This is similar to the product OKR "reduce product friction". Please chat and make sure your work is coordinated and you are working toward the same goals.

company/okrs/2020_q1.md
beyang: How about "Resolve all deal-blocking support requests for tier 1 and tier 2 customers"?

@dadlerj is there a way to tag support conversations as "deal-blocking" in Hubspot?
@juliasourceress are we continuing with our bucketing of customers into tiers 1, 2, and 3 this quarter?

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,69 @@
...
+1. **VP Business: Build a Customer Success organization that supports and helps us exceed our ARR goals.**
+   1. VP Business (Success) KR: Generate at least [$M][M] in net new ARR from existing accounts.
+   1. VP Business (Success) KR: Hire a Head of Customer Success who can own this function.
+   1. VP Business (Success) KR: Reach monthly NPS of 60 at all Tier 1 customers.
+   
+1. **VP Business: Make Sourcegraph a lean organization that uses budgets to guide decision making.**
+
+1. **VP Marketing: Deliver sufficient MQLs to sales to reach [$N][N] net new ARR goal.**
+1. **VP Marketing: Create new category of developer tool.**
+1. **VP Marketing: Become the dominant provider of source code search.**
```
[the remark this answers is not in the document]
adamfrankl2015: The 2020 objective is to be the #1 provider. So we need to implement metrics about how many devs use Sourcegraph, and how many devs search source code. So possibly the Q1 objective is to figure out how to measure this.

nicksnyder:
If we are not going to form a tree with the OKRs then the prefixes on the KRs are a bit redundant.

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,84 @@
...
+   1. Distribution KR: Surface information to site admins and managers to encourage expansion/upsell
+
+1. **Code intelligence: Make it easy for customers to set up LSIF**
+   1. Code intelligence KR: At least three customers are using LSIF on their primary repositories
+   1. Code intelligence KR: At least one customer is using LSIF across all of their code in languages for which there exists an LSIF indexer
+
+1. **Code intelligence: Make it clear where code intel results are coming from**
+   1. Code intelligence KR: Code intelligence UI (tooltips, references panel, etc.) indicates whether or not results are precise
+   
+1. **Code intelligence: Increase precision and recall of LSIF for nearest-commit**
```
nicksnyder: @chrismwendt This is a good objective statement on its own, but now there are 4 objectives and it would be nice to stick to no more than 3. Also, the "Make it clear where code intel results are coming from" objective above doesn't feel meaty enough to be its own objective.

What if we combined these objectives (leave KRs unchanged) and call the new objective something like "Increase user confidence in code intelligence provided by LSIF"?

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,84 @@
...
+   1. Code intelligence KR: Fall back to fuzzy if the file has been modified since the last LSIF upload
+   1. Code intelligence KR: Push LSIF data through diffs
```
nicksnyder: @chrismwendt Can you elaborate a little bit so that (1) it is clearer whether we have achieved this KR and (2) a less informed person can understand what this means.

company/okrs/2020_q1.md
```diff
@@ -80,5 +80,9 @@
    1. Code intelligence KR: Implement search within dependencies and/or dependents of current repo
    1. Code intelligence KR: Implement searching for defs/refs by symbol name
 
+1. **Web: Make users aware of advanced search features**
+   1. Web KR: Shorten time to first search
```
nicksnyder: @lguychard do we already have a metric tracking this? If not, then the KR should be to create a metric to start tracking this. Same question for the next KR.

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,86 @@
...
+1. **Distribution: Improve developer productivity**
+   1. Distribution KR: Lower build times to X
+   1. Distribution KR: Automate away release captain role
+   1. Distribution KR: Infrastructure for automated deploys and releases (no human in loop required by default)
+1. **Distribution: Build features that accelerate customers through pipeline**
+   1. Distribution KR: Shorten the time from "intent to try Sourcegraph" to "whole org is using Sourcegraph" (onboarding experience)
+   1. Distribution KR: Ensure new features (LSIF, GitLab Native Integration, Automation) are accessible and easily tryable by customers once they go GA
+   1. Distribution KR: Surface information to site admins and managers to encourage expansion/upsell
+
+1. **Code intelligence: Make it easy for customers to set up LSIF**
```
nicksnyder: @chrismwendt I think this is a better objective statement:

```suggestion
1. **Code intelligence: LSIF is adopted by customers**
```

We want customers to actually adopt LSIF. Making it easy to setup is just one part of that.

company/okrs/2020_q1.md
tsenart: Also, anything automation related?

christinaforney:
> If we are not going to form a tree with the OKRs then the prefixes on the KRs are a bit redundant.

I actually preferred the tree structure - it helped orient towards top level company goals and showed how the work the team is doing directly contributes. 

@sqs - what is your motivation to have each team broken into sections? Are we planning on keeping them this way or moving into a tree structure once everyone has contributed (ease with merge conflicts).

nicksnyder:
fwiw, I kind of like the lack of a tree structure. I sometimes felt the nesting was arbitrary and not meaningful. It also means that a team's OKRs are always adjacent to each other instead of potentially spread across the tree.

company/okrs/2020_q1.md
christinaforney: Updated. The intention is to make sure that we can scale tracking customer priorities as we increase the number of customers we have.

company/okrs/2020_q1.md
christinaforney: Consolidated to single KR of reducing the time from intent to rollout. @beyang - I essentially grabbed one of the distribution KRs where we had overlap since this is a broader issue to tackle from a product perspective. From what we talked about in the team syncs the distribution team will be taking a piece of this by focusing on improving the deployment/rollout process. Does this line up with your expectation from those conversations?

company/okrs/2020_q1.md
christinaforney: Updated, see below.

company/okrs/2020_q1.md
christinaforney: We discussed in team meetings and I have updated and commented above with the result.

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,86 @@
...
+   1. VP Eng KR: Transition an existing teammate to be (or hire) an Engineering Manager of the distribution team.
+   1. VP Eng KR: Document how we do performance reviews and determine compensation for teammates in engineering roles.
+1. **VP Eng: Create a data driven planning process.**
+   1. VP Eng KR: Each item in a tracking issue has an estimate of time or size.
+   1. VP Eng KR: Each team commits to a written work plan for each release cycle that accounts for the work estimates and the team's capacity.
+   1. VP Eng KR: Each team completes >80% of the work that they committed to at the beginning of the release cycle each iteration.
+
+1. **Distribution: Resolve issues for critical customer deals and expansions**
+1. **Distribution: Improve developer productivity**
+   1. Distribution KR: Lower build times to X
```
nicksnyder: @beyang I think we have reached diminishing marginal returns on build times. I think a different metric that would be more impactful would be time it take for a line of code to get deployed to sourcegraph.com after getting merged into master.

company/okrs/2020_q1.md
lguychard: @nicksnyder edited to add KR about creating metrics.

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,91 @@
...
+1. **Code intelligence: Make search better using precise code intel data**
+   1. Code intelligence KR: Implement search within dependencies and/or dependents of current repo
+   1. Code intelligence KR: Implement searching for defs/refs by symbol name
+
+1. **Web: Make users aware of advanced search features**
+   1. Web KR: Start tracking time to first search and search filters per query in pings
+   1. Web KR: Shorten time to first search
+   1. Web KR: Increase average number of search filters per query
+
+1. **Web: Automation is adopted by customers**
```
lguychard: @nicksnyder @christinaforney @mrnugget @tsenart @felixfbecker how would you feel about this automation OKR? (It should probably be for both Web and Core Services @tsenart)

mrnugget: > At least four customers are using campaigns in production

I don't know whether four is realistic or even within reach, but I like "customer using in production" :)

The one thought I had though is that ultimately this is not something we can influence, right? It's still up to a customer to decide to use. I don't know if that makes a difference, but at previous companies where we used OKRs one guideline was that we should only define Os and KRs that we could, theoretically, complete on our own.

> All campaign types have been used by customers in production

I don't know how useful that is (also: we _could_ just reduce the number of types and have one customer use the single one that's left) and I wonder whether this doesn't put "quantity" over "quality" in a way.

company/okrs/2020_q1.md
mrnugget: Chiming in to add a +1 to @nicksnyder's suggestion here 😄 

Another thought I had when seeing "Improve developer productivity" as objective: who owns alerting and flaky builds? I think getting rid of the alerting flakes and reducing the e2e test flakiness would probably be the biggest developer productivity boost at the moment.

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,91 @@
...
+1. **VP Marketing: Deliver sufficient MQLs to sales to reach [$N][N] net new ARR goal.**
+1. **VP Marketing: Create new category of developer tool.**
+1. **VP Marketing: Become the dominant provider of source code search.**
+
+## Product
+
+1. **Product: Deliver on product roadmap**
+   1. Product KR: Roadmap communicates rough timeline for upcoming work.
+   1. Product KR: Each feature is reviewed before release to ensure quality UX.
+1. **Product: Reduce product friction to accelerate customers through pipeline**
```
mrnugget: One recurring thought I had in the past week: what about documentation?

We _have_ a lot of documentation, but I think it's not the most navigatable docs page (navbar on the left doesn't match document structure, but lists subpages, which are sometimes listed, sometimees not, etc.) and it has rendering glitches and it's often a mix between "guide" and "reference".

If we look to some documentation pages that I would personally consider "best in class" (https://stripe.com/docs, https://www.terraform.io/docs/index.html) we can get a ton of inspiration for how to improve our documentation.

I do think there's a possibility that docs add needless "friction" and might cause customers to drop out of the pipeline.

tsenart: Huge 👍 to this.

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,114 @@
...
+   1. Core Services KR: Implement OR query support
+   1. Core Services KR: Implement NOT query support
+
+1. **Core Services: Existing customers are cared for**
+   1. Core Services KR: Each iteration work plan includes prioritised customer issues.
+   1. Core Services KR: Each team mate works on relevant customer issues every iteration.
+
+1. **Core Services: Automation is polished and adopted by customers**
+   1. Core Services KR: Write blog post about automation.
+   1. Core Services KR: At least four customers are running campaigns.
```
nicksnyder: @tsenart I think we should avoid duplication on this KR. See discussion on the web one above.

```suggestion
```

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,114 @@
...
+   1. Core Services KR: Implement stacking diffs.
+
+1. **Core Services: Structural search is polished and adopted by customers**
+   1. Core Services KR: Write blog post about structural search.
+   1. Core Services KR: At least four customers are running structural searches.
+   1. Core Services KR: Structural search can be ran at scale (~30k repositories).
```
nicksnyder: @tsenart make this measurable (choose X)

```suggestion
   1. Core Services KR: Structural searches over ~30k repositories complete in less than X seconds.
```

company/okrs/2020_q1.md
lguychard: > I don't know whether four is realistic

This was an arbitrary starting point. Open to edits on this :) My thinking was that this is something we could influence, through:
- Helping customers who are keen to try it out to set up a8n, create their first campaigns. We would probably get valuable, actionable feedback this way, although I know that a8n has already been demoed/discussed with customers a lot.  
- Identifying and solving blockers / feature gaps that are preventing customers from trying it out.

Do you have suggestions for other KRs that we could more directly make progress on, without being blocked on externalities, and that wouldn't amount to "build features X, Y, Z"?

> I don't know how useful that is

On second thought I think you're right, I'll remove this one.

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,114 @@
...
+1. **Web: Make users aware of advanced search features**
+   1. Web KR: Start tracking time to first search and search filters per query in pings
+   1. Web KR: Shorten time to first search
+   1. Web KR: Increase average number of search filters per query
+
+1. **Web: Automation is adopted by customers**
+   1. Web KR: At least four customers are using campaigns in production
+   1. Web KR: All campaign types have been used by customers in production
+
+1. **Core Services: Build features that accelerate customers through pipeline**
```
nicksnyder: @tsenart is O and the KRs are vague and don't seem to directly apply to this team. I think we could just delete this. This also helps reduce the number of OKRS for the team (5 seems like too many).

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,114 @@
...
+   1. Core Services KR: Shorten the time from "intent to try Sourcegraph" to "whole org is using Sourcegraph" (onboarding experience)
+   1. Core Services KR: Ensure new features (Automation, ACLs, etc) are accessible and easily tryable by customers once they go GA
+
+1. **Core Services: Search has feature parity with OpenGrok**
+   1. Core Services KR: Implement AND query support
+   1. Core Services KR: Implement OR query support
+   1. Core Services KR: Implement NOT query support
+
+1. **Core Services: Existing customers are cared for**
```
nicksnyder: @tsenart I think this is business as usual and doesn't need to be an ORK. This also helps reduce the number of OKRS for the team (5 seems like too many).

If anything, there could be a KR somewhere about documenting how this team does monthly planning (and these points would live in that documentation).

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,114 @@
...
+   1. Core Services KR: Each iteration work plan includes prioritised customer issues.
+   1. Core Services KR: Each team mate works on relevant customer issues every iteration.
+
+1. **Core Services: Automation is polished and adopted by customers**
+   1. Core Services KR: Write blog post about automation.
+   1. Core Services KR: At least four customers are running campaigns.
+   1. Core Services KR: Implement stacking diffs.
```
nicksnyder: @tsenart can you elaborate on what this means so that the average ready can better understand?

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,114 @@
+# 2020-Q1 OKRs
+
+## CEO
+
+1. **CEO: Grow ARR consistently**
+   1. CEO KR: [$N][N] net new ARR
+   1. CEO KR: Sales leadership owns revenue targets
+   1. CEO KR: Sales, engineering, and product have a successful process for prioritization
+1. **CEO: Aspirational, efficient all-remote team**
```
nicksnyder: Can you add a KR to make our handbook fast to load. It still isn't. I suspect it is slow after a deploy, which is not infrequent.

nicksnyder: @sqs

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,114 @@
...
+   1. Web KR: Shorten time to first search
+   1. Web KR: Increase average number of search filters per query
+
+1. **Web: Automation is adopted by customers**
+   1. Web KR: At least four customers are using campaigns in production
+   1. Web KR: All campaign types have been used by customers in production
+
+1. **Core Services: Build features that accelerate customers through pipeline**
+   1. Core Services KR: Shorten the time from "intent to try Sourcegraph" to "whole org is using Sourcegraph" (onboarding experience)
+   1. Core Services KR: Ensure new features (Automation, ACLs, etc) are accessible and easily tryable by customers once they go GA
```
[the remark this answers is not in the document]
rvantonder: Passing thought: as part of automation, launch N campaigns on the Sourcegraph codebase.

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,69 @@
...
+   1. Sales KR: Round robin new instances across team to collectively achieve a 15%+ response rate from Prospects who set up a new Sourcegraph instance.
+
+1. **VP Business: Build a Customer Success organization that supports and helps us exceed our ARR goals.**
+   1. VP Business (Success) KR: Generate at least [$M][M] in net new ARR from existing accounts.
+   1. VP Business (Success) KR: Hire a Head of Customer Success who can own this function.
+   1. VP Business (Success) KR: Reach monthly NPS of 60 at all Tier 1 customers.
+   
+1. **VP Business: Make Sourcegraph a lean organization that uses budgets to guide decision making.**
+
+1. **VP Marketing: Deliver sufficient MQLs to sales to reach [$N][N] net new ARR goal.**
```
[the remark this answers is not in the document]
adamfrankl2015: We don't have the data yet to create a detailed model--so I am assuming an across the board increase of 50% QoQ on all of our business metrics in order to get to 4-5x annual growth.

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,114 @@
...
+   1. Web KR: All campaign types have been used by customers in production
+
+1. **Core Services: Build features that accelerate customers through pipeline**
+   1. Core Services KR: Shorten the time from "intent to try Sourcegraph" to "whole org is using Sourcegraph" (onboarding experience)
+   1. Core Services KR: Ensure new features (Automation, ACLs, etc) are accessible and easily tryable by customers once they go GA
+
+1. **Core Services: Search has feature parity with OpenGrok**
+   1. Core Services KR: Implement AND query support
+   1. Core Services KR: Implement OR query support
+   1. Core Services KR: Implement NOT query support
```
rvantonder: I'm not sure it makes sense to break this down by operator. We should think about the expression language as one unit. The part that we'll have to expand is where and how this language applies (i.e., implementation that turns expressions into search functionality. In what order, if any, should we support file contents, file paths, repos, etc? AFAIK OpenGrok basically only works with expressions dealing with file content (and maybe file paths? not sure). More considerations to come in RFC...

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,114 @@
...
+   1. Core Services KR: Each iteration work plan includes prioritised customer issues.
+   1. Core Services KR: Each team mate works on relevant customer issues every iteration.
+
+1. **Core Services: Automation is polished and adopted by customers**
+   1. Core Services KR: Write blog post about automation.
+   1. Core Services KR: At least four customers are running campaigns.
+   1. Core Services KR: Implement stacking diffs.
+
+1. **Core Services: Structural search is polished and adopted by customers**
+   1. Core Services KR: Write blog post about structural search.
```
rvantonder: I am authoring a blog post like this right now, so as far the phrasing goes here, this will be "done", and won't really have any team level involvement. I'm all for this being a team effort, I just think the phrasing here needs to change to accommodate work items that others can latch onto.

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,86 @@
...
+1. **Distribution: Improve developer productivity**
+   1. Distribution KR: Lower build times to X
+   1. Distribution KR: Automate away release captain role
+   1. Distribution KR: Infrastructure for automated deploys and releases (no human in loop required by default)
+1. **Distribution: Build features that accelerate customers through pipeline**
+   1. Distribution KR: Shorten the time from "intent to try Sourcegraph" to "whole org is using Sourcegraph" (onboarding experience)
+   1. Distribution KR: Ensure new features (LSIF, GitLab Native Integration, Automation) are accessible and easily tryable by customers once they go GA
+   1. Distribution KR: Surface information to site admins and managers to encourage expansion/upsell
+
+1. **Code intelligence: LSIF is adopted by customers**
```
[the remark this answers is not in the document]
chrismwendt: Added a KR for tracking code intel perf.

company/okrs/2020_q1.md
rvantonder: An explicit perf goal like this is not something that I've had in mind immediately. I think it's a good goal. Just, as I'm currently the only person working on this, my question for this is "who is going to, or would like to, commit time to this effort this quarter?" I think that some discussion needs to cover this if it'll be a team level OKR.

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,69 @@
...
+
+1. **VP Business: Build a Customer Success organization that supports and helps us exceed our ARR goals.**
+   1. VP Business (Success) KR: Generate at least [$M][M] in net new ARR from existing accounts.
+   1. VP Business (Success) KR: Hire a Head of Customer Success who can own this function.
+   1. VP Business (Success) KR: Reach monthly NPS of 60 at all Tier 1 customers.
+   
+1. **VP Business: Make Sourcegraph a lean organization that uses budgets to guide decision making.**
+
+1. **VP Marketing: Deliver sufficient MQLs to sales to reach [$N][N] net new ARR goal.**
+1. **VP Marketing: Create new category of developer tool.**
```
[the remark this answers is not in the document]
adamfrankl2015: https://docs.google.com/document/d/14WKHDr2sVVXhCz9eBgMgPm4oYx9MT3gM7xhCrIlDpto/edit

company/okrs/2020_q1.md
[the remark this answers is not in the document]
adamfrankl2015: https://docs.google.com/document/d/14WKHDr2sVVXhCz9eBgMgPm4oYx9MT3gM7xhCrIlDpto/edit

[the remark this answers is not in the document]
adamfrankl2015: 56 MQLs is a 50% uplift from Q4.

company/okrs/2020_q1.md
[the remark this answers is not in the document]
adamfrankl2015: Our Q1 measurable result is 3 original pieces in the press.

company/okrs/2020_q1.md
rvantonder: To clarify: If the idea is that I have significant individual ownership over the bullets in this section, that's fine too, but then I'd rather condense it (because it feels slightly disproportionate in size to the KRs above), and /or propose other things I have in mind rather than `30K repos in X seconds`.

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,114 @@
...
+   1. Web KR: Start tracking time to first search and search filters per query in pings
+   1. Web KR: Shorten time to first search
+   1. Web KR: Increase average number of search filters per query
+
+1. **Web: Automation is adopted by customers**
+   1. Web KR: At least four customers are using campaigns in production
+   1. Web KR: All campaign types have been used by customers in production
+
+1. **Core Services: Build features that accelerate customers through pipeline**
+   1. Core Services KR: Shorten the time from "intent to try Sourcegraph" to "whole org is using Sourcegraph" (onboarding experience)
```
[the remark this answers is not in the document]
tsenart: This whole OKRs is here to account for the dependency that the distribution team may have on us to accomplish the same OKR. For me it's fine to have the distribution team own this general OKR and we'll support them in each iteration as needed with the changes required for whatever they try to accomplish, even though we don't have an explicit team OKR around this. WDYT?

company/okrs/2020_q1.md
tsenart: So, the Web team owns this?

company/okrs/2020_q1.md
tsenart: @rvantonder: This perf KR is not set in stone! It was a point for discussion. These are team OKRs though, not personal. It's OK if there's a main driver behind each KR, but I'd like us to think about whether what we're proposing here is achievable as a team, not as individuals.

Considering our other OKRs, do you feel that this would be too much (even with contributions from other team mates)?

company/okrs/2020_q1.md
tsenart: https://github.com/sourcegraph/about/pull/422#discussion_r365512415

company/okrs/2020_q1.md
tsenart: > @tsenart I think this is business as usual and doesn't need to be an ORK. This also helps reduce the number of OKRS for the team (5 seems like too many).

OK, I had a sense that it was important enough to warrant it's own KR, but I see you saying OKRs are more about new developments / achievements rather than maintenance and operations. Is that the case?

> If anything, there could be a KR somewhere about documenting how this team does monthly planning (and these points would live in that documentation).

That's on my plate, not the team's plate.

company/okrs/2020_q1.md
tsenart: I'll link to the RFC: https://docs.google.com/document/d/1m6WH47vl-DFynfEdjpGez2pj7bTO3MRFAPpDgEDlvFI/edit

company/okrs/2020_q1.md
tsenart: Please propose the KRs that you think make sense. I do think that the O is a good one to aspire to though.

company/okrs/2020_q1.md
tsenart: Not every KR will have balanced contribution from all team members. I think it's good to include this KR here, since it contributes so directly to the Objective.

company/okrs/2020_q1.md
mrnugget: > You can certainly influence it by focusing on building things that will unblock particular customers from actually using automation (and not working on things that don't directly contribute to us getting customers). I hope we can agree that the outcome we want as a company is to get customers on automation. Given that, it is just a matter of who should own that outcome.

Absolutely. "Influence" was probably the wrong word here. What I meant to say: even if we get four customers to say "yes, we'll purchase Automation and yes, we'll use it in production" in the end it's still possible for deal/rollout to be delayed due to external factors, such as customer not getting contract approved by finance department, etc.

I want the goals (1) to be as explicit as possible and (2) avoid measuring the wrong thing. If we have to add "well, what we mean with that is that we do everything _in our power_ to enable customers to adopt Automation" then I'd rather have "achieve $everything-in-our-power" as the key result. Things like: remove feature flag for Automation (meaning: get it polished!), write "getting started with Automation" guide, do 5 customer demos for Automation, record an Automation screencast, ...

> I think "don't build campaign types unless we can get them used by a customer in production" is a valid way to interpret this KR. Focus is good.

Same thing here. I don't think "All campaign types have been used by customers in production" is a key result that _we as engineers_ can check off. We can do exactly what you said and "don't build campaign types unless we can get them used by a customer in production", but I was hoping there's a better way to formulate it.

> Do you have suggestions for other KRs that we could more directly make progress on, without being blocked on externalities, and that wouldn't amount to "build features X, Y, Z"?

See my list in the paragraph above. I think we mainly need to close the feature gap (drafts, updating campaigns, stacking diffs, campaign timeline events, ...), polish (little UX polishes, _great_ UI for managing campaigns, stability, performance) and document it.

company/okrs/2020_q1.md
keegancsmith: Yeah I agree with deleting this given it is captured by other team OKRs. It would be nice to capture this idea, so we can point to it when trying to decide how to prioritize different bits of work. However, there is already plenty we are wanting to achieve so I am happy to leave this off and we work on it if we can. I trust our intuitions/conversations we have each month around this sort of topic.

company/okrs/2020_q1.md
lguychard: @tsenart I wouldn't see a problem with two teams working towards the same OKR. Would you?

company/okrs/2020_q1.md
nicksnyder: > OKRs are more about new developments / achievements rather than maintenance and operations. Is that the case?

Yes, ongoing work should be captured by KPIs (but we don't have those yet): https://about.gitlab.com/company/okrs/#okrs-are-what-is-different

company/okrs/2020_q1.md
nicksnyder: My proposal in https://github.com/sourcegraph/about/pull/422/files#r365303011 is that product should actually own this KR if we wanted to avoid duplication.

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,115 @@
...
+   1. Core Services KR: Implement OR query support
+   1. Core Services KR: Implement NOT query support
+
+1. **Core Services: Existing customers are cared for**
+   1. Core Services KR: Each iteration work plan includes prioritised customer issues.
+   1. Core Services KR: Each team mate works on relevant customer issues every iteration.
+
+1. **Core Services: Automation is polished and adopted by customers**
+   1. Core Services KR: Write blog post about automation.
+   1. Core Services KR: At least four customers are running campaigns.
```
mrnugget:
```suggestion
   1. Core Services KR: 3 different customers have together run at least 10 campaigns.
```

company/okrs/2020_q1.md
rvantonder: KRs, two ways to phrase this, roughly:
- Users can search for file contents using `AND`/`OR`/`NOT` operators 

or

- Sourcegraph supports a more expressive way to search file contents using `AND`/`OR`/`NOT` operators 

I'm wording with a restriction on "file contents" as the initial feature set. We want to expand this out to repos/file paths, but there are some "unknown unknowns" at this point and I don't know how much complexity that will import So: file contents is a good first start. 

Note that implementing AND/OR/NOT functionality will implicitly allow a 'bag of words' style search (so not only more expressive, but also a new search kind). So a direct result of this is a follow up KR that we support bag of words search on file contents.

company/okrs/2020_q1.md
rvantonder: Agree on "not every KR will be balanced balance". I'm just having trouble seeing how this phrasing invites N team mates contribute to this KR, where N > 1. One idea to reframe this is to encourage team mates generate technical documentation (blog posts, an examples section in the usage docs) that showcase our search functionality, particularly for structural search, or the new bag-of-words/opengrok variety from above, and then say 'we will generate N pieces of such technical documentation'.

company/okrs/2020_q1.md
rvantonder: Since there's a planned rework putting things under a more general 'search' objective, let's go from there. I think we can just drop this KR in the reworked section.

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,115 @@
...
+1. **Core Services: Search has feature parity with OpenGrok**
+   1. Core Services KR: Implement AND query support
+   1. Core Services KR: Implement OR query support
+   1. Core Services KR: Implement NOT query support
+
+1. **Core Services: Existing customers are cared for**
+   1. Core Services KR: Each iteration work plan includes prioritised customer issues.
+   1. Core Services KR: Each team mate works on relevant customer issues every iteration.
+
+1. **Core Services: Automation is polished and adopted by customers**
```
mrnugget: @lguychard @tsenart Here's the complete list of key results that we came up with in today's Core Services sync:

```
1. Core Services KR: Document Automation and write a blog post about it.
1. Core Services KR: Automation is enabled by default and not behind a feature flag.
1. Core Services KR: Run 3 user tests with Sourcegraph colleagues, collect issues and fix them
1. Core Services KR: 3 different customers have together run at least 10 campaigns.
```

Feel free to edit/remove/add to this.

I'm also still unsure how to get the features in this list or whether they should be in there at all, since we're going to work on them anyway.

company/okrs/2020_q1.md
```diff
@@ -88,28 +88,18 @@ Build out vision for product and run customer validation discussions
...
-   1. Core Services KR: Implement [stacking diffs](https://docs.google.com/document/d/1m6WH47vl-DFynfEdjpGez2pj7bTO3MRFAPpDgEDlvFI/edit).
-
-1. **Core Services: Structural search is polished and adopted by customers**
-   1. Core Services KR: Write blog post about structural search.
-   1. Core Services KR: At least four customers are running structural searches.
-   1. Core Services KR: Structural search can be ran at scale (~30k repositories).
+   1. Core Services KR: Document Automation and write a blog post about it.
+   1. Core Services KR: Automation is enabled by default and not behind a feature flag.
+   1. Core Services KR: Run 3 user tests with Sourcegraph colleagues, collect issues and fix them.
+   1. Core Services KR: 3 different customers have together run at least 10 campaigns.
```
mrnugget: Approve! 👍 

Tagging @lguychard here because Web & Core Services should be in sync here.

lguychard: The KRs look good. Re: ownership, what should we do here? Duplicating automation OKRs between web and core services does not make sense to me. These KRs are not core services-specific and web would be working towards them as well through our work on automation. What do you think of @nicksnyder's proposal to have product own this OKR (https://github.com/sourcegraph/about/pull/422/files#r365303011)?

mrnugget: > What do you think of @nicksnyder's proposal to have product own this OKR (https://github.com/sourcegraph/about/pull/422/files#r365303011)?

That makes a lot of sense to me. We'd still work towards it and help product achieve this, mostly by working on the things on the roadmap and by fixing the things from the user tests, etc.

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,105 @@
...
+   1. Code intelligence KR: At least one customer is using LSIF across all of their code in languages for which there exists an LSIF indexer
+   1. Code intelligence KR: Start tracking the performance (user interaction to result) of hover tooltips, go to definition, and find references
+
+1. **Code intelligence: Increase user confidence in code intelligence provided by LSIF**
+   1. Code intelligence KR: Code intelligence UI (tooltips, references panel, etc.) indicates whether or not results are precise
+   1. Code intelligence KR: Fall back to fuzzy if the file has been modified since the last LSIF upload
+   1. Code intelligence KR: Make use of LSIF data when the current line is not part of the diff since the last LSIF upload
+
+1. **Code intelligence: Make search better using precise code intel data**
+   1. Code intelligence KR: Implement search within dependencies and/or dependents of current repo
```
creachadair: There's a huge implementation cost difference between "dependencies" and "dependents". Is the latter intended to be scoped to a given customer's data? (If not it implies infrastructure I think we don't have yet)

company/okrs/2020_q1.md
```diff
@@ -88,28 +88,18 @@ Build out vision for product and run customer validation discussions
...
-
-1. **Core Services: Search has feature parity with OpenGrok**
-   1. Core Services KR: Implement AND query support
-   1. Core Services KR: Implement OR query support
-   1. Core Services KR: Implement NOT query support
-
-1. **Core Services: Existing customers are cared for**
-   1. Core Services KR: Each iteration work plan includes prioritised customer issues.
-   1. Core Services KR: Each team mate works on relevant customer issues every iteration.
+1. **Core Services: Search gains new capabilities which are adopted by customers**
```
rvantonder: Everything looks great here, except for the structural search 30K repo thing.

company/okrs/2020_q1.md
```diff
@@ -88,28 +88,18 @@ Build out vision for product and run customer validation discussions
...
+   1. Core Services KR: Support AND/OR/NOT operators on file contents.
+   1. Core Services KR: Write at least 2 blog posts about new search features.
+   1. Core Services KR: Structural search can be ran at scale (~30k repositories).
```
rvantonder: Everything for search looks great, but I still don't know about the structural search at scale thing. What does it mean, what runtimes are we looking for? I can say that already structural search will be 5x to 10x slower than regex, and this is reasonable because it does something more specialized. Structural search (or any 'analysis') will "run" at 30K repos but could take 'forever'. So without a time bound or idea around that I don't know what this KR means. I suggest we remove it.

nicksnyder: What if we changed this KR to be something like "add metrics for search latency per search type" (my assumption is that we don't track this from customer instances yet).

nicksnyder: Alternatively (although we should probably add metrics regardless), I think the point of this KR is to assert that structural search can actually be used at large customers. It wouldn't be useless, for example, if structural searches on 30k repos timed out after 60s.

I think it should be possible to formulate a KR like this:

```suggestion
   1. Core Services KR: Structural search over ~30k repositories returns results in less than X seconds.
```

Where X can be large enough to make you @rvantonder comfortable that this is possible to achieve, but low enough to be reasonable for a human to wait for with the context that they have never had this capability before (e.g. X = 30s might be ok).

A rough estimate would be fine here (e.g. if we expect this to be 10x slower than regex, then how fast are regex queries on 30k repos?).

rvantonder: > A rough estimate would be fine here (e.g. if we expect this to be 10x slower than regex, then how fast are regex queries on 30k repos?)

That's a good perspective. I think the trouble with doing this by number of repos is that it conflates speed of structural/regex search and the performance of the infrastructure around dealing with 30K repos, without saying how big those repos are. We can thumbsuck values for how many lines of text are in an average repo, but I think we can get closer to a metric for search functionality by saying "X KLOC per second" benchmarks (and this will also answer what the slow down is without talking about number of repos). If there is a scaling challenge unique to searching on 30K repos, then we should think about a metric that can track that.

nicksnyder: > I think the trouble with doing this by number of repos is that it conflates speed of structural/regex search and the performance of the infrastructure around dealing with 30K repos, without saying how big those repos are.

What if the OKR made a performance guarantee on structural search on k8s.sgdev.org (which is designed to replicate a large scale customer)? Regardless of the wording of this OKR, what we really what we want is "If current/future large customer X can use regular search, they should be able to use structural search" and by "use" I mean "searches complete in a reasonable amount of time, as judged by users". 

If we are having a hard time quantifying that, then lets change this KR to be about adding metrics so we can be more precise in the future.

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,105 @@
...
+   1. Web KR: Shorten time to first search
+   1. Web KR: Increase average number of search filters per query
+
+1. **Web: Automation is adopted by customers**
+   1. Web KR: At least four customers are using campaigns in production
+   1. Web KR: All campaign types have been used by customers in production
+
+1. **Core Services: Search gains new capabilities which are adopted by customers**
+   1. Core Services KR: Support AND/OR/NOT operators on file contents.
+   1. Core Services KR: Write at least 2 blog posts about new search features.
```
nicksnyder: @rvantonder @tsenart 

Are blog posts the right medium of success here? They are inherently ephemeral. Should we prefer docs (even if they are long form), instead?

rvantonder: I originally suggested this should be N pieces of technical documentation (includes usage docs, walkthrough examples, and blog posts). Likely in that order. IMO our existing search features really need some attention here (there is a lot of room for improvements on our search keywords reference and examples pages)

company/okrs/2020_q1.md
nicksnyder: Given the current state of alerts and e2e flakiness, and the correlation to being able to automate the release process, I think an explicit OKR around e2e test reliability would be nice.

company/okrs/2020_q1.md
rvantonder: Example, this exists on twitter https://twitter.com/slimsag/status/1214271545012416514 but not in our examples https://docs.sourcegraph.com/user/search/examples. We all use search, so an inclusive KR that everyone can contribute to here would be nice.

So perhaps `s/new search features/new and existing search features`

company/okrs/2020_q1.md
rvantonder: > What if the OKR made a performance guarantee on structural search on k8s.sgdev.org

I'd be onboard with that. The immediate question is the technical effort to just get a reliable signal for such a metric (benchmarking is hard, and if we do the whole end-to-end stack for a search query, there's bound to be jitter and other confounding variables). 

Thinking what this entails, it is more in the direction of 'change this KR to be about adding metrics', which would give us a clear signal to do the former. So :+1: for adding metrics.

company/okrs/2020_q1.md
nicksnyder: Agree with expanding the KR beyond just structural search and also beyond just blog posts.

company/okrs/2020_q1.md
tsenart: > I think the trouble with doing this by number of repos is that it conflates speed of structural/regex search and the performance of the infrastructure around dealing with 30K repos, without saying how big those repos are. 

This doesn't match my understanding exactly. Indexed search, for instance, allows search latency to **not** grow linearly with the number of repos mirrored by Sourcegraph.

Would there be ways to have that same sub-linear latency growth with structural search?

Challenges like repo size being uneven (i.e. work units aren't homogeneous) can be overcome — for instance by splitting repo archives into relatively equal sized parts and scaling infrastructure appropriately for parallel processing.

In the end we own the search problem end to end. As a user, what do you think should be the maximum amount of time you would wait to consider structural search "interactive", regardless of the number and individual size of repos?

> I'd be onboard with that. The immediate question is the technical effort to just get a reliable signal for such a metric (benchmarking is hard, and if we do the whole end-to-end stack for a search query, there's bound to be jitter and other confounding variables).

That's OK. What we want to optimize for is the user experience, so if there's jitter or other sources of slowness in the search path, we want to know about it and fix it. It's relatively easy to capture latency measurements over time with Prometheus, and we can always do before / after load tests, additionally.

company/okrs/2020_q1.md
tsenart:
```suggestion
   1. Core Services KR: Write at least 2 pieces of technical documentation (docs, walkthroughs, blog posts) about search functionality.
```

company/okrs/2020_q1.md
sqs: Yeah, this is really annoying. I investigated a bit and filed https://github.com/sourcegraph/about/issues/499. If I can't close that issue by the time OKRs are ready, then I will incorporate that into an OKR.

company/okrs/2020_q1.md
rvantonder: > Would there be ways to have that same sub-linear latency growth with structural search?

I understand the desire/vision here but can we even say that our unindexed regex search does not grow linearly by repos?

> As a user, what do you think should be the maximum amount of time you would wait to consider structural search "interactive", regardless of the number and individual size of repos?

I had a longish response to this, but I'm not going to answer it for sake of brevity. We can do that offline.

---

I think this can keep going back and forth, and in the interest of cutting that short, my overarching concern is "What does this KR mean?" If @nicksnyder' comment is accurate:

> I think the point of this KR is to assert that structural search can actually be used at large customers. 

and you think his suggestions are valid (I agree with them):

> What if the OKR made a performance guarantee on structural search on k8s.sgdev.org (which is designed to replicate a large scale customer)? 

> What if we changed this KR to be something like "add metrics for search latency per search type" (my assumption is that we don't track this from customer instances yet).

> we should probably add metrics regardless

then I feel that we need to establish and set a baseline and subsequent OKRs can iterate on that baseline. We can establish a baseline against, e.g., k8s. So this KR could be establishing a baseline performance for search along N parameters. I think this could apply to existing search generally, unless it's been done already or we believe it doesn't apply because "existing search is fast enough \<handwave\>"

---

Speaking very frankly, a much more impactful commitment than any of this would be introducing search pagination that gives us _more control_ to _display results quickly_ since it mimics a kind of streaming capability rather than optimizing our batch-like search path. If there is a mutual willingness from @slimsag and core to help accelerate that effort then we should allocate engineering to that instead.

rvantonder: After catching up @tsenart and I agree that we'll revise this to emphasize "time to first result" for search. We'll stipulate that there should be a 99th percentile latency/time to return results within N seconds. This encompasses a wide possibility of strategies, including adding metrics, optimizing our end-to-end path (e.g., gitserver fetches and caches) and adding control knobs for displaying results faster with, e.g., pagination / streaming.

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,105 @@
...
+   1. Web KR: Increase average number of search filters per query
+
+1. **Web: Automation is adopted by customers**
+   1. Web KR: At least four customers are using campaigns in production
+   1. Web KR: All campaign types have been used by customers in production
+
+1. **Core Services: Search gains new capabilities which are adopted by customers**
+   1. Core Services KR: Support AND/OR/NOT operators on file contents.
+   1. Core Services KR: Write at least 2 pieces of technical documentation (docs, walkthroughs, blog posts) about search functionality.
+   1. Core Services KR: Structural search can be ran at scale (~30k repositories).
```
rvantonder:
```suggestion
   1. Core Services KR: Ensure time-to-first-result for search requests in the 99th percentile latency of 2s, up to any number of repositories.
```

@tsenart are your thoughts to have this for 'search' generally, or structural search only? I think with the addition of AND/OR/NOT which will demand more compute or multiple search passes, it is worth adding this cap this for any search request. Also, since basically my first day at Sourcegraph I've been unhappy that it takes >2s to list all repos for a simple query like `repo:github.com/rvantonder` or `repo:github.com/tsenart` on sourcegraph.com [and it still takes >2s](https://sourcegraph.com/search?q=repo:github.com/rvantonder&patternType=literal) so I really think the other forms of search can use some attention.

tsenart: > are your thoughts to have this for 'search' generally, or structural search only?

Search, generally.

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,105 @@
...
+   1. Distribution KR: Infrastructure for automated deploys and releases (no human in loop required by default)
+1. **Distribution: Build features that accelerate customers through pipeline**
+   1. Distribution KR: Shorten the time from "intent to try Sourcegraph" to "whole org is using Sourcegraph" (onboarding experience)
+   1. Distribution KR: Ensure new features (LSIF, GitLab Native Integration, Automation) are accessible and easily tryable by customers once they go GA
+   1. Distribution KR: Surface information to site admins and managers to encourage expansion/upsell
+
+1. **Code intelligence: LSIF is adopted by customers**
+   1. Code intelligence KR: At least three customers are using LSIF on their primary repositories
+   1. Code intelligence KR: At least one customer is using LSIF across all of their code in languages for which there exists an LSIF indexer
+   1. Code intelligence KR: Start tracking the performance (user interaction to result) of hover tooltips, go to definition, and find references
```
tsenart: I believe there's urgency to fixing our end to end user perceived latency so I'd like to suggest we aim for more than just start measuring it this quarter (as we're doing for search in core services).

```suggestion
   1. Code intelligence KR: 99th percentile of end to end latency of hover tooltips, go to definition, and find references is below N seconds.
```

Framing it like this ensure we focus on user perceived latency and it may require you to split this KR into multiple ones due to the potentially differing complexities of making each of those user interactions faster. We'll probably need things like https://github.com/sourcegraph/sourcegraph/issues/1423 to measure it.

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,105 @@
...
+
+1. **Code intelligence: Increase user confidence in code intelligence provided by LSIF**
+   1. Code intelligence KR: Code intelligence UI (tooltips, references panel, etc.) indicates whether or not results are precise
+   1. Code intelligence KR: Fall back to fuzzy if the file has been modified since the last LSIF upload
+   1. Code intelligence KR: Make use of LSIF data when the current line is not part of the diff since the last LSIF upload
+
+1. **Code intelligence: Make search better using precise code intel data**
```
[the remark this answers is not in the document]
efritz: Can we leave it as a slush objective in the case that the 99th percentile dashboard shows that we've already hit our target?

That or it's something we could pull back in.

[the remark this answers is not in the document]
nicksnyder: We can always add OKRs later in the iteration if we finish everything else. As of right now, enhancing search doesn't feel like a priority. @christinaforney Are there customers waiting on this? The roadmap doesn't contain any information.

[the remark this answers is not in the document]
creachadair: I agree that we should remove it for now, and add it back later if we have space. Until we have a good idea of how many users are using it, how well it performs, and how reliable it is, it is probably a little early to talk about improving it. (Though I fully expect we will come back to this topic regardless)

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,105 @@
...
+Build out vision for product and run customer validation discussions
+
+## Engineering
+
+1. **VP Eng: Build a world class engineering team.**
+   1. VP Eng KR: Grow team to plan (+1 code intel, +2 distribution, +2 web, +1 core services).
+   1. VP Eng KR: Transition an existing teammate to be (or hire) an Engineering Manager of the distribution team.
+   1. VP Eng KR: Document how we do performance reviews and determine compensation for teammates in engineering roles.
+1. **VP Eng: Create a data driven planning process.**
+   1. VP Eng KR: Each item in a tracking issue has an estimate of time or size.
```
nicksnyder: This is an implementation detail of the KR below

```suggestion
```

company/okrs/2020_q1.md
christinaforney: We should find an owner for improving the docs (See [RFC 102](https://docs.google.com/document/d/1rmNEiyU-gP97BYVwZWVWo7G94jeRAtrQrk4VZGlcBic/edit))

company/okrs/2020_q1.md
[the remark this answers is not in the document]
efritz: ~If we are successful in Q1~ this would be a great Q2 OKR.

FTFY.

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,114 @@
...
+
+1. **VP Marketing: Deliver sufficient MQLs to sales to reach [$N][N] net new ARR goal.**
+1. **VP Marketing: Create new category of developer tool.**
+1. **VP Marketing: Become the dominant provider of source code search.**
+
+## Product
+
+1. **Product: Deliver on product roadmap**
+   1. Product KR: Roadmap communicates rough timeline for upcoming work.
+   1. Product KR: Each feature is reviewed before release to ensure quality UX.
```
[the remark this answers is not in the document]
christinaforney: Updated, great suggestion!

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,91 @@
...
+1. **Product: Reduce product friction to accelerate customers through pipeline**
+   1. Product KR: Shorten onboarding time of "intent to try Sourcegraph" to "whole organization is using Sourcegraph".
+   1. Product KR: Shorten product feedback cycle on new features.
+1. **Product: Build the product and design team to support engineering and ARR growth**
+   1. Product KR: Grow team to plan (+1 UX Designer, +1 PM).
+   1. Product KR: Handbook describes product strategy and processes.
+   1. Product KR: Customer feature prioritization process scales with team and customer growth.
```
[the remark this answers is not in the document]
christinaforney: Removed to focus on building team and creating/documenting process.

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,114 @@
...
+1. **VP Marketing: Create new category of developer tool.**
+1. **VP Marketing: Become the dominant provider of source code search.**
+
+## Product
+
+1. **Product: Deliver on product roadmap**
+   1. Product KR: Roadmap communicates rough timeline for upcoming work.
+   1. Product KR: Each feature is reviewed before release to ensure quality UX.
+1. **Product: Reduce product friction to accelerate customers through pipeline**
+   1. Product KR: Shorten onboarding time of "intent to try Sourcegraph" to "whole organization is using Sourcegraph".
```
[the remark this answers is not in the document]
christinaforney: Removed both to focus on building team and creating/documenting process.

company/okrs/2020_q1.md
lguychard: @nicksnyder @christinaforney @tsenart since web and core services should be in sync here, what is the final takeaway here? Should this be listed as a "split ownership" OKR (`**Core services, web**`), or should it be owned by product as suggested by @nicksnyder in https://github.com/sourcegraph/about/pull/422/files#r365303011 ?

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,97 @@
...
+
+1. **VP Eng: Build a world class engineering team.**
+   1. VP Eng KR: Grow team to plan (+1 code intel, +2 distribution, +2 web, +1 core services).
+   1. VP Eng KR: Transition an existing teammate to be (or hire) an Engineering Manager of the distribution team.
+   1. VP Eng KR: Document how we do performance reviews and determine compensation for teammates in engineering roles.
+1. **VP Eng: Create a data driven planning process.**
+   1. VP Eng KR: Each team commits to a written work plan for each release cycle that accounts for the work estimates and the team's capacity.
+   1. VP Eng KR: Each team completes >80% of the work that they committed to at the beginning of the release cycle each iteration.
+
+1. **Distribution: Resolve issues for major customer**
```
sqs:
```suggestion
1. **Distribution: Resolve issues for major customers**
```

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Sales and marketing
+
+1. **Sales (Closing Deals): Achieve [$N][N] net new ARR.**
+   1. Have each AE close 3 Tier 1 deals (>=$100K ARR) and 10 non-Tier 1 deals, 2 of which are from their own outbound outreach efforts.
+   1. Get 70% of companies on a free trial to convert to paid customers.
+1. **Sales (Pipeline Optimization): Scale our sales pipeline to 50+ deals per AE, of which 60% of deals are engaged and actively evaluating Sourcegraph.**
+   1. Round robin new instances across team to collectively achieve a 15%+ response rate from Prospects who set up a new Sourcegraph instance.
+
+1. **VP Business: Build a Customer Success organization that supports and helps us exceed our ARR goals.**
```
sqs: @dadlerj @juliasourceress This is pending our discussion of Customer Engineering

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,103 @@
...
+   1. Team members rely on the handbook for cross-functional info and processes
+   1. The handbook is fast to browse and search
+
+## Sales and marketing
+
+1. **Sales (Closing Deals): Achieve [$N][N] net new ARR.**
+   1. Have each AE close 3 Tier 1 deals (>=$100K ARR) and 10 non-Tier 1 deals, 2 of which are from their own outbound outreach efforts.
+   1. Get 70% of companies on a free trial to convert to paid customers.
+1. **Sales (Pipeline Optimization): Scale our sales pipeline to 50+ deals per AE, of which 60% of deals are engaged and actively evaluating Sourcegraph.**
+   1. Round robin new instances across team to collectively achieve a 15%+ response rate from Prospects who set up a new Sourcegraph instance.
```
sqs: @juliasourceress New instances' site admins may be reached out to first by Customer Engineering not Sales if we decide to do that. So this is pending the Customer Engineering discussion.

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,103 @@
...
+   1. Sales, engineering, and product have a working process for prioritization and delivery
+1. **CEO: Aspirational, efficient all-remote team**
+   1. Roles and responsibilities are documented and maintained in handbook
+   1. Team members rely on the handbook for cross-functional info and processes
+   1. The handbook is fast to browse and search
+
+## Sales and marketing
+
+1. **Sales (Closing Deals): Achieve [$N][N] net new ARR.**
+   1. Have each AE close 3 Tier 1 deals (>=$100K ARR) and 10 non-Tier 1 deals, 2 of which are from their own outbound outreach efforts.
```
sqs: @juliasourceress These sales OKRs should reflect the sales plan you are working on. Should we consider these pending until the plan is ready?

company/okrs/2020_q1.md
christinaforney: The indirect ownership here feels off to me. This is a broader question beyond just Automation planning and as we continue to grow our cross-functional teams will require more collaboration. This almost feels like there is a need to have an Automation specific OKR that is owned by individuals on core and web. I know this is a bit of a regression back to having an Automation team, but this is still a big enough focus that requires two teams of effort.

Thoughts on creating a single Automation objective that is owned by both the core and web teams? Both teams would review the OKR in team meetings.

nicksnyder: I also thought about just creating an automation OKR, and since you mentioned it, I think it is a pragmatic solution for now. It avoids duplication in the OKRs and there is a set of people who clearly own it.

> Isn't that the same as both team having the same OKR? It's just how we display it that changes, right?

Yes.

nicksnyder: I pushed two changes to address this:

https://github.com/sourcegraph/about/pull/422/commits/84805a2796011dedbca29d222f5db46484a778a3
https://github.com/sourcegraph/about/pull/422/commits/a725aaedfd2611a650d24ee2a69f9c5cda05e58c

company/okrs/2020_q1.md
dadlerj: 👍 

@sqs in case you didn't see it, I updated my OKRs and left a note on this one in a separate PR at https://github.com/sourcegraph/about/pull/513 (apologies if I used the wrong process here?)

company/okrs/2020_q1.md
dadlerj: 👍 

@sqs in case you didn't see it, I updated my OKRs and left a note on this one in a separate PR at https://github.com/sourcegraph/about/pull/513 (apologies if I used the wrong process here?)

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,107 @@
...
+
+### Distribution
+
+1. **Resolve issues for major customers**
+   1. Ensure that no technical issue blocks a deal for any tier 1 or tier 2 customer.
+1. **Improve developer productivity**
+   1. Automate away release captain role and adopt continuous releases (no human in loop required by default).
+   1. e2e and regression test suite is 100% reliable (no flakiness) when run in CI.
+1. **Understand where customers falloff during setup process**
+   1. Time to important admin setup milestones (i.e. instance started, admin account created, >1 reporitories added) are tracked in [pings](https://docs.sourcegraph.com/admin/pings).
```
nicksnyder:
```suggestion
   1. Time to important admin setup milestones (i.e. instance started, admin account created, >1 repositories added) are tracked in [pings](https://docs.sourcegraph.com/admin/pings).
```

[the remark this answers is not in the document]
dadlerj: We already capture all of this

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,107 @@
...
+   1. e2e and regression test suite is 100% reliable (no flakiness) when run in CI.
+1. **Understand where customers falloff during setup process**
+   1. Time to important admin setup milestones (i.e. instance started, admin account created, >1 repositories added) are tracked in [pings](https://docs.sourcegraph.com/admin/pings).
+   1. 50th and 90th percentile time from account creation to important user onboarding milestones (i.e. first search, first hover tooltip, first go-to-definition, first find references) are tracked in [pings](https://docs.sourcegraph.com/admin/pings).
+
+### Code intelligence
+
+1. **LSIF is adopted by customers**
+   1. At least three customers have > 50% of code intel queries powered by LSIF (excluding fallback references)
+   1. At least one customer has > 75% of code intel queries powered by LSIF (excluding fallback references)
```
nicksnyder: @efritz I would like to see a more aggressive % here mitigated by the fact that denominator only includes code intel requests for languages that we have an indexer for (I think we would need to tweak/add a metric in your rfc).
```suggestion
   1. At least one customer has > 95% of code intel queries on languages that we have an LSIF generator for powered by LSIF (excluding fallback references).
```

efritz: > on languages that we have an LSIF generator for

We can't capture that unless we can return language stats in pings.

nicksnyder: Eh, I think the tracking for my suggestion above would annoyingly explode the data that we would need to collection. How about we just update the % to be more aggressive :)

```suggestion
   1. At least one customer has > 90% of code intel queries powered by LSIF (excluding fallback references)
```

nicksnyder: > We can't capture that unless we can return language stats in pings.

We could capture it without knowing language stats (client hard codes which languages we have lsif generator for and then we capture aggregate counts of fuzzy code intel requests for only those language), but as I said above I don't think that extra tracking is worth the effort or complexity.

company/okrs/2020_q1.md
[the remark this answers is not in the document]
nicksnyder: Yeah after I thought about this more I suspect that our current tracking is sufficient here (assuming that we track the binary state of whether an instance has reached each state).

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,107 @@
...
+### Distribution
+
+1. **Resolve issues for major customers**
+   1. Ensure that no technical issue blocks a deal for any tier 1 or tier 2 customer.
+1. **Improve developer productivity**
+   1. Automate away release captain role and adopt continuous releases (no human in loop required by default).
+   1. e2e and regression test suite is 100% reliable (no flakiness) when run in CI.
+1. **Understand where customers falloff during setup process**
+   1. Time to important admin setup milestones (i.e. instance started, admin account created, >1 reporitories added) are tracked in [pings](https://docs.sourcegraph.com/admin/pings).
+   1. 50th and 90th percentile time from account creation to important user onboarding milestones (i.e. first search, first hover tooltip, first go-to-definition, first find references) are tracked in [pings](https://docs.sourcegraph.com/admin/pings).
```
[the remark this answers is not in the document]
nicksnyder:
```suggestion
```

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,107 @@
...
+   1. Each team commits to a written work plan for each release cycle.
+   1. Each team completes >80% of the work that they committed to at the beginning of the release cycle each iteration.
+
+### Distribution
+
+1. **Resolve issues for major customers**
+   1. Ensure that no technical issue blocks a deal for any tier 1 or tier 2 customer.
+1. **Improve developer productivity**
+   1. Automate away release captain role and adopt continuous releases (no human in loop required by default).
+   1. e2e and regression test suite is 100% reliable (no flakiness) when run in CI.
```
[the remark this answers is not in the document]
beyang: Core services owning for now. Will remove from here.

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,106 @@
...
+
+1. **Build a world class engineering team.**
+   1. Grow team to plan (+1 code intel, +2 distribution, +2 web, +1 core services).
+   1. Transition an existing teammate to be (or hire) an Engineering Manager of the distribution team.
+   1. Document how we do performance reviews and determine compensation for teammates in engineering roles.
+1. **Create a data driven planning process.**
+   1. Each team commits to a written work plan for each release cycle.
+   1. Each team completes >80% of the work that they committed to at the beginning of the release cycle each iteration.
+
+### Distribution
```
beyang: @nicksnyder updated following our conversation in Distribution sync today

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,106 @@
...
+
+1. **Resolve issues for major customers**
+   1. Ensure that no technical issue blocks a deal for any [tier 1 or tier 2 customer](https://sourcegraph.looker.com/dashboards/132).
+1. **Improve developer productivity**
+   1. Automate away release captain role and adopt continuous releases (no human in loop required by default).
+1. **Improve installation and site admin onboarding process**
+   1. Track time to important admin setup milestones (i.e. instance started, admin account created, >1 repositories added) are tracked in [pings](https://docs.sourcegraph.com/admin/pings).
```
nicksnyder: @beyang Dan said that we already track (or can track this based on current pings): https://github.com/sourcegraph/about/pull/422#discussion_r371519877

If so (can you confirm?) I think we can delete this KR.

```suggestion
```

company/okrs/2020_q1.md
nicksnyder: Thanks! I left one comment below.

company/okrs/2020_q1.md
ebrodymoore: We track _some_ things. Here's an example of major milestones we already have data on https://sourcegraph.looker.com/looks/508. @beyang whether or not you leave it as a KR, it'd probably be worth it to chat and make sure we can get you the data you need.

company/okrs/2020_q1.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Engineering
+
+### VP Eng
+
+1. **Build a world class engineering team.**
+   1. Grow team to plan (+1 code intel, +2 distribution, +2 web, +1 core services).
+   1. Transition an existing teammate to be (or hire) an Engineering Manager of the distribution team.
+   1. Document how we do performance reviews and determine compensation for teammates in engineering roles.
+1. **Create a data driven planning process.**
```
nicksnyder:
```suggestion
1. **Identify long term owners for new engineering functions**
  1. Hire someone to own security now and who has the potential to lead a team in the future.
  1. Hire someone to own internal developer productivity now and who has the potential to lead a team in the future.
1. **Create a data driven planning process.**
```

christinaforney:
@sqs, @juliasourceress, @adamfrankl2015 - I know we are still waiting on OKRs from Marketing and Sales, but it is becoming difficult for the engineering teams to track this since they aren't published. Can we mark the Marketing and Sales OKRs as WIP and merge this, then follow up in separate PRs? I can do this for you if I get a thumbs up on this :)

sqs:
@christinaforney Yes, please do it! That would be a huge help. Thank you.

PR #428, opened 2019-12-04
Adding step for implementing RFCs

christinaforney:
Adding this to start the conversation about putting RFCs into action. Something that has been brought up by the team ([one example](https://sourcegraph.slack.com/archives/CMMTWQQ49/p1575362741306300)) is that once an RFC is being worked on, there is no place to track what parts of that RFC have been implemented, are in progress, or that we might want to prioritize at a later time. 

As a PM it is also hard for me to keep track of all the moving pieces that have been implemented and not. This creates a space to link together all the work that is done for an RFC where the work is being done (and not in a google doc). 

Only RFCs that are going to be worked on, or are starting to be worked on should have tracking issues created to avoid a proliferation of GitHub issues that are not currently prioritized in the project roadmap.

### DIFF
--- a/handbook/communication/rfcs/index.md
+++ b/handbook/communication/rfcs/index.md
@@ -18,6 +18,7 @@ Each RFC has a status that is in the title of the RFC (e.g. "RFC 1 WIP: Title").
 | REVIEW | The RFC is ready to be reviewed. The RFC explicitly lists whose approvals are required and a requested timeline for those approvals. |
 | APPROVED | All comment threads are resolved and the RFC has been approved by all required approvers. Just because an RFC is approved doesn't mean it will definitely be implemented (e.g. priorities may change, or new information might be discovered during implementation or code review that causes a change in the plan). It is in the author's best interest to avoid surprises at code review time by ensuring the RFC has a sufficient level of detail and has approval from all relevant stakeholders. |
 | ABANDONED | There are no plans to move forward with this RFC. The particular reason is communicated in the metadata section of the RFC. For example, the RFC may have failed to get the necessary approvals, it may be been superseded by another RFC, priorities may have changed, or we may not have resources to work on this RFC in the foreseeable future. |
+| IN PROGRESS | This RFC is currently being worked on and has been migrated to a GitHub issue and broken down (as needed) into sub-issues.
 | IMPLEMENTED | This RFC has been implemented. |
 
 A prose description of the status appears in the [metadata](#RFC-structure) of an RFC.
@@ -110,6 +111,10 @@ Sometimes there is information relevant to an RFC, but that information can't be
 - If a material amount of the content is non-public then it is ok to make the RFC only accessible to the Sourcegraph team.
 
 Only Sourcegraph teammates are able to see the revision history of RFCs (because edit access is required for that). This means if sensitive information is found in the document all you need to do is remove that information. Comments that accidentally contain sensitive info can be deleted.
+
+## Implementing RFCs
+
+When the team decides to implement an APPROVED RFC, a tracking issue is created in GitHub with as-needed sub-issues created to break down the work required to implement the full RFC. Teams may choose to implement only a portion of an RFC, and this enables us to capture the work that has not yet been completed for later prioritization.
 
 ## References
 


handbook/communication/rfcs/index.md
```diff
@@ -18,6 +18,7 @@ Each RFC has a status that is in the title of the RFC (e.g. "RFC 1 WIP: Title").
 | REVIEW | The RFC is ready to be reviewed. The RFC explicitly lists whose approvals are required and a requested timeline for those approvals. |
 | APPROVED | All comment threads are resolved and the RFC has been approved by all required approvers. Just because an RFC is approved doesn't mean it will definitely be implemented (e.g. priorities may change, or new information might be discovered during implementation or code review that causes a change in the plan). It is in the author's best interest to avoid surprises at code review time by ensuring the RFC has a sufficient level of detail and has approval from all relevant stakeholders. |
 | ABANDONED | There are no plans to move forward with this RFC. The particular reason is communicated in the metadata section of the RFC. For example, the RFC may have failed to get the necessary approvals, it may be been superseded by another RFC, priorities may have changed, or we may not have resources to work on this RFC in the foreseeable future. |
+| IN PROGRESS | This RFC is currently being worked on and has been migrated to a GitHub issue and broken down (as needed) into sub-issues.
```
nicksnyder: I am fine adding an IN PROGRESS status, but I don't think RFCs universally need to be translated to GitHub issues.

```suggestion
| IN PROGRESS | This RFC is currently being worked on.
```

handbook/communication/rfcs/index.md
```diff
@@ -111,6 +112,10 @@ Sometimes there is information relevant to an RFC, but that information can't be
 
 Only Sourcegraph teammates are able to see the revision history of RFCs (because edit access is required for that). This means if sensitive information is found in the document all you need to do is remove that information. Comments that accidentally contain sensitive info can be deleted.
 
+## Implementing RFCs
+
+When the team decides to implement an APPROVED RFC, a tracking issue is created in GitHub with as-needed sub-issues created to break down the work required to implement the full RFC. Teams may choose to implement only a portion of an RFC, and this enables us to capture the work that has not yet been completed for later prioritization.
```
nicksnyder: This is not universally true because we use RFCs for non-engineering and product things. I think this is an implementation detail of the RFC and the team implementing it, and doesn't need to be documented here.

TBH, I think the pain we are experiencing now in automation is an artifact of the rough planning process and high variance in the level of detail on the published RFCs. No other team seems to have this issue so I think it is premature to add more process to our documentation.

tsenart: > TBH, I think the pain we are experiencing now in automation is an artifact of the rough planning process and high variance in the level of detail on the published RFCs. 

What @christinaforney proposed above, then, may be something good for us to experiment with for automation work going forward, before resolving to make it team wide policy.

@mrnugget: This proposal is coming in part due to how you felt recently that there were quite a few loose ends on status of previous RFCs. You brain-dumped all of those loose ends into https://github.com/sourcegraph/sourcegraph/issues/7010 which definitely helped bring clarity into the situation.

What do you think about this proposal for automation work? Would you suggest some changes or something else entirely?

mrnugget: I don't think creating a tracking issue for every RFC is necessary. I think the problem is due to this:

> we use RFCs for non-engineering and product things

Some RFCs are not like others. Some of them are easy to keep track of. Example: [RFC 30 - Zoekt Horizontal Scaling](https://docs.google.com/document/d/18w8T_KzYxQye8wg1g01QpMOX4_ERTtbOxMBRYaOEkmk/edit#heading=h.rf2d3v2oo71l), and that contains multiple steps of implementation.

I think for me the biggest problem comes from meta RFCs that are like mini-roadmaps or taskslists, except that they're never checked off. Examples:

- [RFC 20: Monitoring existing campaigns](https://docs.google.com/document/d/1UY9B_kLlwRtYj-fuv7XZS1-Mu99Czx9Ojm7sgEmoQIA/edit#heading=h.t1nv3cdlfjqp)
- [RFC 36: Campaign Creation (public milestone, splitted up into multiple technical milestones)](https://docs.google.com/document/d/1DKWy2zC6_rDZzoPS7ASUpvnVDK3Zj6cf_I9Riq9KNu0/edit#heading=h.b674i7gxjses)

What these two share is that they split up the included requirements into "required" and "stretch-goals/nice-to-haves". But when planning/implementing/updating the slides or roadmap, the RFC is synonymous with the "required" things, and the stretch goals are left behind.

As I wrote on Slack, I don't have a better solution I can propose right now. But I think that maybe these meta-RFCs shouldn't be RFCs but instead still part of something like [a8n sequence of milestones](https://docs.google.com/document/d/1TDsjrCy55UTZA_NyofVssnBotTyPP6Hvbrsp__aUCfM/edit#heading=h.go9qqwdnhiyu) and then the things we know we want to tackle will be extracted into technical milestones.

Or, as proposed, instead of RFCs they're "feature" tracking issues on GitHub with taskslists and, if required, technical RFCs can be written for a subset of the tasks.

nicksnyder: > As I wrote on Slack, I don't have a better solution I can propose right now. But I think that maybe these meta-RFCs shouldn't be RFCs but instead still part of something like a8n sequence of milestones and then the things we know we want to tackle will be extracted into technical milestones.

100% agree. I don't think these should have been RFCs, they were more vision documents.

christinaforney: I think some of the challenges of the Search UI features could have been overcome with better sub-issues on the RFCs as well. I like that it creates a step for the implementer to think about and break down the problem into implementation actions. 

We can always qualify this section to be for development RFCs that are being implemented by engineering. If this really is just an implementation detail, it is a detail that has been missed on several occasions. I had discussed this idea with @tsenart in our 1:1, so wanted to make sure to write it down. I'm happy to pull this if it really isn't helpful in this context.

christinaforney: Should we abandon this change and try this out with Automation (where it makes sense) as @tsenart suggested to start?

mrnugget: That sounds good to me.

tsenart: > Should we abandon this change and try this out with Automation (where it makes sense) as @tsenart suggested to start?

:+1:

PR #429, opened 2019-12-04
Update Q4 Core Services OKR outcomes

tsenart:


company/okrs/2019_q4.md
```diff
@@ -1,7 +1,7 @@
 # 2019-Q4 OKRs
 
 1. **CEO: Net new ARR**
-   1. **Sales (Closing Deals): Reach [$N](https://docs.google.com/document/d/1clukaACTCjPLnmhEttJXcOwadxHEGYK24Znn0HUQGB0/edit#bookmark=kix.n8t17z6iyawc) net new ARR in Q4 by fully onboarding 2 new AEs who will close 4 Tier 1 deals (>=$100K ARR) and 11 non-Tier 1 deals.**
+   1. **Sales (Closing Deals): Reach [\$N](https://docs.google.com/document/d/1clukaACTCjPLnmhEttJXcOwadxHEGYK24Znn0HUQGB0/edit#bookmark=kix.n8t17z6iyawc) net new ARR in Q4 by fully onboarding 2 new AEs who will close 4 Tier 1 deals (>=\$100K ARR) and 11 non-Tier 1 deals.**
```
christinaforney: I was having trouble with this markup too - my prettier auto-format kept adding these backslashes, but they end up showing up in our handbook pages when I was checking locally. How do we want to handle this?

tsenart: Hmm, is there another Markdown renderer we could use that supports this @sqs?

sqs: Not that I'm aware of. The Markdown formatter we use in docsite and in the product (https://github.com/russross/blackfriday) is the only good one available for Go when I last checked. We use some advanced features of it, including walking the Markdown AST, so it would not be easy to switch to another one (and very few Markdown libs expose the AST anyway).

I see the same problem being mentioned in Hugo at https://discourse.gohugo.io/t/backslash-rendered-to-html-in-escaped-dollar-sign/15531, which has a workaround (adding a comment to disable prettier).

christinaforney: Let's give this a shot :) 

```suggestion
   <!-- prettier-ignore -->
   1. **Sales (Closing Deals): Reach [$N](https://docs.google.com/document/d/1clukaACTCjPLnmhEttJXcOwadxHEGYK24Znn0HUQGB0/edit#bookmark=kix.n8t17z6iyawc) net new ARR in Q4 by fully onboarding 2 new AEs who will close 4 Tier 1 deals (>=$100K ARR) and 11 non-Tier 1 deals.**
```

nicksnyder: Or use a git client (e.g. https://desktop.github.com/) that allows you to commit only a portion of the diff for a file :)

christinaforney: I guess a piece of clarification that I am wondering about: Do we want to follow the style enforcements from prettier for our markdown files? It's easy enough to turn off auto-formatting as well. I'm just not clear on what the expectations are. See recent handbook PRs with initial commits of formatting fixes and then second commits with changes: #431 and #430

sqs: No, the handbook markdown does not use or enforce prettier. If your editor runs it on all files, that’s your call, but you should avoid that if it’s confusing people or introducing other unrelated changes into your PR.

In my experience, prettier for markdown is not worth it because it means you can’t make edits from the web (without running into minor violations that add a ton of time to the process).

nicksnyder: 1. Do we want to enforce prettier as a build step => No for the reasons Quinn mentioned.
2. Is it nice to keep things consistently formatted => I think this is good and seems to be what happens by default given the prettier config (or lack thereof) in this repo.

If we want to change (2) then I think we need to adjust the prettier config in this repo.

@felixfbecker as a prettier expert, do you have any advice?

felixfbecker: Here's the Prettier issue for this: https://github.com/prettier/prettier/issues/6213

I personally find Prettier very valuable in markdown, but if this causes problems for now, I would recommend adding `"prettier.requireConfig": true` to your VS Code user settings. This will make Prettier only run in repos that have a prettier config checked in (not this repo, but our main repo).

sqs: Thanks, @felixfbecker. This PR adds that VS Code setting so that people with the VS Code prettier extension should not see this undesired behavior anymore on handbook Markdown files. https://github.com/sourcegraph/about/pull/435

PR #430, opened 2019-12-04
Document reviewer friendly force pushing policy in PRs

tsenart:
- First commit is the result of VS Code auto formatting the Markdown file.
- The second is the actual change, so please review that commit only.

### DIFF
--- a/handbook/engineering/code_reviews.md
+++ b/handbook/engineering/code_reviews.md
@@ -67,6 +67,7 @@ Do:
   - Describe any remaining open questions that you are seeking feedback on.
 - Make it clear who's review(s) you are waiting for.
 - Politely remind your reviewer that you are waiting for a review if they haven't responded after one business day.
+- Don't force push commits that erase the latest reviewed commit or any commit prior to that. This makes it easy for reviewers to pick up where they left off in the previous round.
 
 ## What makes an effective code review?
 
--- a/handbook/engineering/code_reviews.md
+++ b/handbook/engineering/code_reviews.md
@@ -67,7 +67,7 @@ Do:
   - Describe any remaining open questions that you are seeking feedback on.
 - Make it clear who's review(s) you are waiting for.
 - Politely remind your reviewer that you are waiting for a review if they haven't responded after one business day.
-- Don't force push commits that erase the latest reviewed commit or any commit prior to that. This makes it easy for reviewers to pick up where they left off in the previous round.
+- If you need to work on your PR after a first round of reviews, try not to force push commits that erase the latest reviewed commit or any commit prior to that. This makes it easy for reviewers to pick up where they left off in the previous round.
 
 ## What makes an effective code review?
 


handbook/engineering/code_reviews.md
```diff
@@ -67,6 +67,7 @@ Do:
   - Describe any remaining open questions that you are seeking feedback on.
 - Make it clear who's review(s) you are waiting for.
 - Politely remind your reviewer that you are waiting for a review if they haven't responded after one business day.
+- Don't force push commits that erase the latest reviewed commit or any commit prior to that. This makes it easy for reviewers to pick up where they left off in the previous round.
```
mrnugget: Can we add that prior to requesting reviews or having recevied a review it's fine to clean up commits by force pushing?

I think this sounds much more general than intended. As I understood it, this rule should help avoid a situation where a PR goes through multiple rounds of reviews and the reviewers lose track of what they already looked at after the PR author force-pushed changes in between rounds. I think the rule is good to avoid _that situation_, but in other situations it's, IMHO, fine to force push for whatever reasons.

tsenart: @mrnugget: Sure thing, please suggest how you'd like to word it.

handbook/engineering/code_reviews.md
```diff
@@ -42,31 +42,32 @@ It should be obvious to any other engineer on the team from the PR description a
...
-    - Discuss alternative implementations that you considered but didn't pursue (and why).
-    - Describe any remaining open questions that you are seeking feedback on.
+- Create meaningful PR title and description that communicates **what** the PR does and **why** (the **how** is the diff).
+  - Include links to relevant issues (e.g. "closes #1234", "helps #5678").
+  - Include a screenshot, GIF, or video if you are making user facing changes.
+  - Discuss alternative implementations that you considered but didn't pursue (and why).
+  - Describe any remaining open questions that you are seeking feedback on.
 - Make it clear who's review(s) you are waiting for.
 - Politely remind your reviewer that you are waiting for a review if they haven't responded after one business day.
+- Don't force push commits that erase the latest reviewed commit or any commit prior to that. This makes it easy for reviewers to pick up where they left off in the previous round.
```
mrnugget:
```suggestion
- If you need to work on your PR after a first round of reviews, try not to force push commits that erase the latest reviewed commit or any commit prior to that. This makes it easy for reviewers to pick up where they left off in the previous round.
```

handbook/engineering/code_reviews.md
mrnugget: https://github.com/sourcegraph/about/pull/430/files#r353808172

handbook/engineering/code_reviews.md
nicksnyder: I think @mrnugget wording is clearer.

felixfbecker:
I think there is a use case for rebasing and force pushing even if already reviewed - whenever the author thinks that leads to an overall easier review (e.g. new commits change the approach significantly that it will be easier to review the full change from scratch than incrementally, or the original commits were not rebased for commit-by-commit review but as the PR progresses it increasingly makes sense to do so). I feel uneasy about putting into the handbook that doing so is prohibited, i.e. take a way from authors the power to make this call.

handbook/engineering/code_reviews.md
slimsag: I like this better because it also includes `try not to` which implies there are some exceptional cases one could run into in which it is OK to do so.

slimsag:
@felixfbecker I agree and was worried about the same thing initially, what do you think about https://github.com/sourcegraph/about/pull/430#discussion_r353928741 ?

unknwon:
@felixfbecker:
> I think there is a use case for rebasing and force pushing even if already reviewed - whenever the author thinks that leads to an overall easier review (e.g. new commits change the approach significantly that it will be easier to review the full change from scratch than incrementally, or the original commits were not rebased for commit-by-commit review but as the PR progresses it increasingly makes sense to do so).

I think the author of the PR would comment to notify the reviewers anyway in case of major refactor and ask them to do a full review, or even propose a new PR, i.e. force-push is not the only way, so to speak.

> I feel uneasy about putting into the handbook that doing so is prohibited, i.e. take a way from authors the power to make this call.

As [@slimsag commented](https://github.com/sourcegraph/about/pull/430#discussion_r353928741), and I agree that the wording suggested by @mrnugget only encourages teammates to not use force-push but use it whenever necessary.

PR #434, opened 2019-12-05
Adding reimbursable expenses and budgets

mercadon:


handbook/people-ops/spending-company-money.md
```diff
@@ -1,3 +1,67 @@
 # Spending company money
+Please see below for a list of reimbursable expenses (all in USD).
```
dadlerj: There are probably some high-level guidelines to add at the top in subsequent bullets (using https://about.gitlab.com/handbook/spending-company-money/ as a guide). A few that seem relevant, and a few additions:

- You are free to purchase these items and expense them without asking permission **as long as it's in the interest of the company**. When in doubt, please contact @sourcegraph/people-ops for clarification before the purchase.
- You likely won't need every single item on this list—please use your best judgment and buy when you need them. Remember, Sourcegraph doesn't have unlimited funds—spend this money as if it was your own.
  - For example, if you intend to [work out of a coworking space](#Ongoing-expenses) most of the time, you may not need a fully equipped home office as well.
- These are average amounts, and should be treated as guidelines, not hard limits. If you expect to significantly exceed the listed amount on any purchase, please notify @sourcegraph/people-ops so we can decide if we need to update this guide.
- If you spend more than $1,000 on any given item, it will be considered property of the company.

... Anything else?

dadlerj: >- These are average amounts, and should be treated as guidelines, not hard limits. If you expect to significantly exceed the listed amount on any purchase, please notify @sourcegraph/people-ops so we can decide if we need to update this guide.

My bullet above may actually be incorrect, looking at the numbers below. Were they intended to be maximum amounts, or averages?

handbook/people-ops/spending-company-money.md
```diff
@@ -1,3 +1,67 @@
...
+**Laptop stand:** $40
+
+**Desk:** $500
+
+**Ergonomic chair:** $200
+
+
+
+## Ongoing expenses
+**Software:** case by case, please get your manager’s approval.
```
dadlerj: I'd rather say: 

>Please with the appropriate team and manager to ask if we already have a subscription or licenses to the software. If not, please get @sourcegraph/people-ops's approval for software or a subscription that would cost more than $500 per year.

dadlerj: Also, let's change the header here to "Software/SaaS/cloud"

handbook/people-ops/spending-company-money.md
```diff
@@ -1,3 +1,67 @@
...
+
+**Internet/WiFi plan:** $60/month
```
dadlerj: Should we make people expense this, or just add it as a default like the cell phone plan? I vote the latter to avoid having to find the receipt every month. And let's reduce to $50, given that.

handbook/people-ops/spending-company-money.md
```diff
@@ -1,3 +1,67 @@
...
+
+**Cell phone plan:** $75/month (no need to expense, this is automatically added onto your pay).
+
+**Coworking/office space:** TBD
```
dadlerj: I have some benchmarks for this one, but I'm actually leaning towards using a "average cost" approach here, rather than a maximum... I don't want to penalize people in SF/NYC/Tokyo/London vs. other locations, so I sort of like the "here's an average, we trust you to find the right deal" approach. What's your take?

handbook/people-ops/spending-company-money.md
```diff
@@ -1,3 +1,67 @@
...
+
+**Moving (moving, packing, shipping, etc.):** $1000/year
+
+**Wellness:** $85/month
+
+
+
+## Other expenses:
+
+**Laptop repairs:** if your laptop requires repairs, we cover up to $500 a year on repairs. If necessary repairs exceed the yearly budget please reach out to PeopleOps for approval.
```
dadlerj: This doesn't imply that the teammates own the laptop, does it? I can't recall our current policy here.

handbook/people-ops/spending-company-money.md
```diff
@@ -1,3 +1,67 @@
...
+
+**Laptop upgrade:** if your laptop is 4 years or older, you are eligible for a new laptop (same budget as new laptop applies).
+
+**Travel expenses:** please see our page on travel budgets. 
+
+
+## How to submit Expenses:
```
dadlerj: This section should just be a link to the page I added a while back! https://about.sourcegraph.com/handbook/people-ops/expenses

handbook/people-ops/spending-company-money.md
slimsag: My preference would be the latter FWIW.

nicksnyder: Agree.

handbook/people-ops/spending-company-money.md
ggilmore: I don’t understand why there should be a cap on repair costs for our work laptop?

handbook/people-ops/spending-company-money.md
tsenart: To be noted that adding to salary means it becomes taxable income, rather than expensable.

handbook/people-ops/spending-company-money.md
tsenart: That sounds fair @dadlerj.

handbook/people-ops/spending-company-money.md
```diff
@@ -1,3 +1,67 @@
 # Spending company money
+Please see below for a list of reimbursable expenses (all in USD).
+
+## One-time expenses
+**Laptop (engineers):** $3500
```
tsenart: Do note that prices of laptops vary substantially across locations and markets (notable case: Apple). Having a fixed USD price limit for everyone for these things isn't ideal and would result in some degree of unwanted unfairness I believe.

handbook/people-ops/spending-company-money.md
```diff
@@ -1,3 +1,67 @@
...
+
+**Keyboard:** $130
+
+**Mouse or trackpad:** $80 or $130
+
+**USB-C Adapter:** $70
+
+**HDMI/monitor cable:** $15
+
+**Laptop stand:** $40
```
tsenart: What would it take to give more flexibility to team mates rather than specifying amounts for each possible hardware component they might need in their work environment? Imagine that team mate 1 doesn't want or need and laptop stand nor a desk, but would rather have a larger monitor. Team mate 2 wants the opposite. Is there a reason we can't accommodate these preferences?

handbook/people-ops/spending-company-money.md
```diff
@@ -1,3 +1,67 @@
...
+
+
+
+## Other expenses:
+
+**Laptop repairs:** if your laptop requires repairs, we cover up to $500 a year on repairs. If necessary repairs exceed the yearly budget please reach out to PeopleOps for approval.
+
+**Laptop upgrade:** if your laptop is 4 years or older, you are eligible for a new laptop (same budget as new laptop applies).
+
+**Travel expenses:** please see our page on travel budgets.
```
tsenart: We should link here.

handbook/people-ops/spending-company-money.md
nicksnyder: Expensing is a pain and time consuming. If we intend to compensate people for internet because it is necessary for the job, then it doesn't make sense to have the whole company waste time every month uploading a receipt to Expensify. I think this is less about tax efficiency (if that is what you are worried about) and more about communicating "your job requires broadband internet access and we compensate for that"

handbook/people-ops/spending-company-money.md
nicksnyder: For engineers, I would rather define this as minimum specs (e.g. CPU, memory, disk space) which would circumvent @tsenart's fairness issue. Engineers would be free to upgrade beyond minimum specs at their own expense. If this proposal sounds good I can propose what those minimum specs would be.

I have no opinion about how this should work for non-engineers.

handbook/people-ops/spending-company-money.md
mercadon: The majority of the amounts are based on what we currently spend on these items, so I was thinking of them as maximums (unless there needed to be an exception).

handbook/people-ops/spending-company-money.md
mercadon: Agreed!

handbook/people-ops/spending-company-money.md
mercadon: @dadlerj Correct, it would still be Sourcegraph property. @ggilmore I should clarify, what I meant here was that it would be $500/year for repairs without needing approval, but anything above that would require additional approval before expensing. Will edit that to be more clear :)

handbook/people-ops/spending-company-money.md
mercadon: It's definitely harder to define for engineers (since certain specs here are more important), so I'm okay with making this spec-based for engineers.

nicksnyder:
```suggestion

**Laptop (engineers):** (Choose one)
  - 13" MacBook Pro, 2.8 GHz 4-core, 16 GB, 512 GB SSD
  - 16" MacBook Pro, 2.6 GHz 6‑core, 16 GB, 512 GB SSD
```

tsenart: Are we requiring team mates to use Macs?

nicksnyder: No. I updated my suggestion.

handbook/people-ops/spending-company-money.md
```diff
@@ -1,3 +1,67 @@
 # Spending company money
+Please see below for a list of reimbursable expenses (all in USD).
+
+## One-time expenses
+**Laptop (engineers):** $3500
+
+**Laptop (non-engineers):** $2000
+
+**Laptop coverage:** $380
```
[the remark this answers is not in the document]
mercadon: Yep!

handbook/people-ops/spending-company-money.md
felixfbecker: @nicksnyder your suggestion appears to be a significant downgrade though (-$1100), not sure if that was intentional. The configuration linked is $2399, vs $3500 of what we currently spend ([if I understand @mercadon correctly here](https://github.com/sourcegraph/about/pull/434#discussion_r355606217)).
As a data point, as I am typing this comment my SSD is full with 605GB (it's probably `node_modules` 😆 ) and I have 29GB of RAM in use, both more than the spec now in the PR but under $3500 total if configured.

nicksnyder: Money isn't what is important here (which is why we removed all references to money). What is important is that engineers have a sufficiently powerful machine to develop our application. I based these initial specs on my laptop (I had to start somewhere), which has served me well but I haven't been doing coding for the last ~2 years. I started with a less powerful laptop when I was hired and asked to upgrade because I noticed it was impacting my productivity. If engineers need 32 GB and a 1TB disk to effectively develop our product, then we can definitely change these specs. I'll send out a survey to collect more data from the team.

FWIW I have a 500GB SSD and am using about half of it. In the past 3 years of having this laptop, I think I have cleaned up unused items on my disk once.

handbook/people-ops/spending-company-money.md
felixfbecker: @tsenart's point is still valid though, isn't it? Back of the envelope calculation: In Germany you get the highest possible income tax of 42% if you earn anything over €54,950/year. $50 are €45 according to Google (not considering Veem's markup). So real money available to pay the broadband bill after taxation is €26. A basic 50 MBit/s at Deutsche Telekom cost €40/month, or 45€ for 100MBit/s for those all-company Zoom meetings. So then it wouldn't really fulfill the intend, unless it's higher.

handbook/people-ops/spending-company-money.md
eseliger: The current 16" mbp has 1TB per default

handbook/people-ops/spending-company-money.md
```diff
@@ -4,9 +4,8 @@ Please see below for a list of reimbursable expenses (all in USD).
 ## One-time expenses
 
 **Laptop (engineers):** (Choose one)
-  - [13" MacBook Pro, 2.8 GHz 4-core, 16 GB, 512 GB SSD](https://www.apple.com/shop/buy-mac/macbook-pro/13-inch-space-gray-256gb-2.4ghz-quad-core-processor-with-turbo-boost-up-to-4.1ghz#)
-  - [16" MacBook Pro, 2.6 GHz 6‑core, 16 GB, 512 GB SSD](https://www.apple.com/shop/buy-mac/macbook-pro/16-inch-space-gray-512gb-2.6ghz-6-core-processor#)
-  - Any other laptop that meets minimum requirements (4-cores, 16 GB, SSD) and is not more expensive than the pre-approved laptops above.
+  - [16" MacBook Pro, 2.4 GHz (5 GHz turbo) 8‑core, 32 GB, 1 TB SSD](https://www.apple.com/shop/buy-mac/macbook-pro/16-inch-space-gray-1tb-2.3ghz-8-core-processor#)
```
eseliger: maybe we add a note that apple does not deep link to config upgrades (which are implied by the link label though?) Noemi and I had that problem first and almost got a wrong one :)

nicksnyder: I marked the upgrades. There will be less back and forth now since engineers will either get the standard mac laptop (as defined here) or a custom (presumably non-mac) one.

PR #437, opened 2019-12-05
releases: fix broken variable expansion + simplify commands

slimsag:
The `git checkout` commands didn't work in Bash as-is, and since this is not hand-written anymore it seems fine to not use variables at all.


sqs:
@slimsag Any blockers to merging this?

PR #438, opened 2019-12-07
Add Continuous Releasability Contract

beyang:
This adds a page to the handbook defining a *Continuous Releasability Contract*, which holds that (1) `master` should always be releasable to customers and (2) we use feature flags for every new feature.

One side effect of abiding by the Releasability Contract is that end-of-cycle QA should become a formality, rather than a period during which we expect to catch a non-zero number of release-blocking issues. In order for this to happen, certain teams may have to do a more thorough QA pass *before* merging/enabling new features in `master` than they have in the past.

I believe this is just a written statement of our existing policy w.r.t to PRs and the state of `master` branch. However, since that policy was not written down and our team has grown since we decided on it, I am not sure that every teammate is aware of it. I'd therefore like everyone on the engineering team to sign off on this.

Initial reviewers, please add your teammates as reviewers after you approve this PR.

Original discussion in [RFC 86](https://docs.google.com/document/d/1mbnLHBhEPT2c8ttoxQaQRstivW0JIbKHKtjg-ZRr5WM/edit). This is in service of OKR `1.vi.c.a` ("Major/minor and patch releases take 1 hour (including general QA but excluding customer-specific testing")

### DIFF
--- a/handbook/engineering/continuous_releasability.md
+++ b/handbook/engineering/continuous_releasability.md
@@ -1,3 +1,5 @@
+# Continuous releasability
+
 We practice **continuous releasability**. This is a variant of [continous
 delivery](https://en.wikipedia.org/wiki/Continuous_delivery) adapted to a monthly release cadence
 that serves the needs of customers running their own Sourcegraph instances.
--- a/handbook/engineering/continuous_releasability.md
+++ b/handbook/engineering/continuous_releasability.md
@@ -4,7 +4,7 @@ We practice **continuous releasability**. This is a variant of [continous
 delivery](https://en.wikipedia.org/wiki/Continuous_delivery) adapted to a monthly release cadence
 that serves the needs of customers running their own Sourcegraph instances.
 
-### Continuous Releasability Contract
+## Continuous Releasability Contract
 
 * `master` branch is always releasable
 * Every new feature includes a feature flag
--- a/handbook/engineering/continuous_releasability.md
+++ b/handbook/engineering/continuous_releasability.md
@@ -44,7 +44,7 @@ Feature flags achieve 2 things:
   Releasability Contract. E.g., if we discover a critical bug in a new feature 1 day before release,
   we can immediately disable it to make the release deadline.
 
-### Exceptions
+## Exceptions
 
 In some cases, it may be prohibitively expensive to preserve releasability. For example, a major
 change made the indexed search backend horizontally scalable. Maintaining strict releasability of
--- a/handbook/engineering/continuous_releasability.md
+++ b/handbook/engineering/continuous_releasability.md
@@ -28,7 +28,7 @@ that the revision can be released to customers. In particular,
 * **Up-to-date Regression test suite**: Every key workflow has a regression test.
   * This means adding/updating regression tests when you add/update workflows.
 
-**A feature flag is required for every new feature**
+### A feature flag is required for every new feature
 
 The preferred feature flag mechanism is the
 [`experimentalFeatures`](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@2b90ec5006f6879193d9a0fd2d2493bc6e061004/-/blob/schema/site.schema.json#L47:6)
--- a/handbook/engineering/continuous_releasability.md
+++ b/handbook/engineering/continuous_releasability.md
@@ -1,0 +1,62 @@
+We practice **continuous releasability**. This is a variant of [continous
+delivery](https://en.wikipedia.org/wiki/Continuous_delivery) adapted to a monthly release cadence
+that serves the needs of customers running their own Sourcegraph instances.
+
+### Continuous Releasability Contract
+
+* `master` branch is always releasable
+* Every new feature includes a feature flag
+
+**`master` is always releasable**
+
+Every change merged into `master` must preserve the releasability of `master`. *Releasable* means
+that the revision can be released to customers. In particular,
+
+* **Up-to-date changelog and docs**: The changelog and documentation reflect the current feature set.
+  * Docs must be updated as soon as a feature is merged and enabled.
+  * You can merge a feature without yet updating the docs if it is behind a disabled feature flag.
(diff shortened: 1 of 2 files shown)


keegancsmith:
> Currently I often try to split up my feature work into multiple smaller PRs which are easier to review, rebase and merge. Does this policy incentive us to do the opposite and make bigger PRs? With a bigger PR you only have to to the manual QA process once, run the testsuite once, etc.

I don't think it does. What it does do is encourage the use of feature flags. This is something I haven't seen us to across the board (although we often hack in an environment variable in core-services to switch on a feature, instead of using experimentalFeatures config).

LGTM

beyang:
> Currently I often try to split up my feature work into multiple smaller PRs which are easier to review, rebase and merge. Does this policy incentive us to do the opposite and make bigger PRs? With a bigger PR you only have to to the manual QA process once, run the testsuite once, etc.

Echoing what @keegancsmith said, as long as the change doesn't change behavior visible to the end user, QA isn't necessary. One might argue that in order to be 1000% sure, you should run the regression test suite for *every* PR (because even PRs that don't intend to change end-user functionality could inadvertently do so), but for now, that decision is at the discretion of the PR author.

handbook/engineering/continuous_releasability.md
```diff
@@ -0,0 +1,62 @@
+We practice **continuous releasability**. This is a variant of [continous
+delivery](https://en.wikipedia.org/wiki/Continuous_delivery) adapted to a monthly release cadence
+that serves the needs of customers running their own Sourcegraph instances.
+
+### Continuous Releasability Contract
+
+* `master` branch is always releasable
+* Every new feature includes a feature flag
+
+**`master` is always releasable**
```
nicksnyder: So we can link directly to it

```suggestion
### `master` is always releasable
```

handbook/engineering/continuous_releasability.md
```diff
@@ -0,0 +1,62 @@
...
+  * Simple features should be tested by someone other than the author.
+  * Complex or scale-sensitive changes should be tested by multiple people on one of the dogfood instances.
+* **No regressions**: The regression test suite passes.
+  * The regression test suite is not run in CI, because it is time-consuming and somewhat flaky.
+  * The author should manually run the regression test suite before merging, *unless* they are
+    certain the change does not break the regression test suite AND introduces no new behavior.
+* **Up-to-date Regression test suite**: Every key workflow has a regression test.
+  * This means adding/updating regression tests when you add/update workflows.
+
+**A feature flag is required for every new feature**
```
nicksnyder:
```suggestion
### A feature flag is required for every new feature
```

handbook/engineering/continuous_releasability.md
```diff
@@ -0,0 +1,62 @@
...
+* Every new feature includes a feature flag
+
+**`master` is always releasable**
+
+Every change merged into `master` must preserve the releasability of `master`. *Releasable* means
+that the revision can be released to customers. In particular,
+
+* **Up-to-date changelog and docs**: The changelog and documentation reflect the current feature set.
+  * Docs must be updated as soon as a feature is merged and enabled.
+  * You can merge a feature without yet updating the docs if it is behind a disabled feature flag.
```
unknwon: What about a feature that isn't necessarily have a feature flag but with explicit configuration option has default "off" value? Should we add feature flag in this case too?

beyang: This seems like a reasonable exception to the rule. What we care most about is ease of disabling. In this case, to disable the buggy feature, we could just remove the option altogether.

beyang:
> I think we also need to have a way (by a checklist in PR template or a bot) to remind the PR author check the bullet points listed in the "master is always releasable" section, i.e. update changelog and docs, QA, regressions test. These checks aren't necessarily are merge blockers w.r.t. our practice, but the PR author should own checking them off.

I will follow up with another PR for updating the PR template (there was some discussion about this in the RFC).

hadrian-git:
If I understood correctly, it looks like the `site.schema.json` will soon get very large, is that correct? If so, has there been discussion into updating the way experimental features are declared so they can be loaded from separate files? (or from a DB for live experimentation)

handbook/engineering/continuous_releasability.md
nicksnyder: All that matters is that we can turn the feature off. Config value is fine if it makes sense for us to allow users to disable the feature in perpetuity.

handbook/engineering/continuous_releasability.md
```diff
@@ -0,0 +1,64 @@
...
+* **QA**: Every feature has been QA'd by an appropriate set of users in an appropriate environment.
+  * Simple bug fixes should be tested at the very least by the author.
+  * Simple features should be tested by someone other than the author.
+  * Complex or scale-sensitive changes should be tested by multiple people on one of the dogfood instances.
```
christinaforney: This was an idea I discussed with @sqs - the author would demo their change to product to validate documentation for that feature as well as usability. My concern is with becoming a bottleneck, though my expectation is that this would be a quick process assuming there aren't any issues with the change. Curious to hear thoughts/concerns/other ideas.

```suggestion
  * Complex or scale-sensitive changes should be tested by multiple people on one of the dogfood instances.
  * Changes that impact the user and new product features should be demoed with Product to verify workflows, documentation, and usability.
```

uwedeportivo:
> > Currently I often try to split up my feature work into multiple smaller PRs which are easier to review, rebase and merge. Does this policy incentive us to do the opposite and make bigger PRs? With a bigger PR you only have to to the manual QA process once, run the testsuite once, etc.
> 
> Echoing what @keegancsmith said, as long as the change doesn't change behavior visible to the end user, QA isn't necessary. One might argue that in order to be 1000% sure, you should run the regression test suite for _every_ PR (because even PRs that don't intend to change end-user functionality could inadvertently do so), but for now, that decision is at the discretion of the PR author.

On top of that you can always have a longer lived feature branch and work the PR branches off it and in the end merge the feature branch into master. This keeps the PRs small.

PR #442, opened 2019-12-08
blog post "Code navigation in GitHub pull requests"

sqs:
This is a blog post that describes an existing feature but makes it really clear/simple how and why to use it. It also should help us rank higher in https://www.google.com/search?q=code+navigation+github+pull+requests and similar queries.

I will record a video to go along with this, but no need to block publishing on the video.

### DIFF
--- a/blogposts/code-navigation-in-github-pull-requests.md
+++ b/blogposts/code-navigation-in-github-pull-requests.md
@@ -1,0 +1,52 @@
+---
+title: Code navigation in GitHub pull requests
+author: Quinn Slack
+publishDate: 2019-12-09T10:00-08:00
+tags: [blog]
+slug: code-navigation-in-github-pull-requests
+heroImage: https://storage.googleapis.com/sourcegraph-assets/code-navigation-in-github-pull-requests-typescript-hover-hero.png
+published: true
+---
+
+Code navigation helps you review code in GitHub pull requests more quickly and effectively.
+
+- **Hovers** show you documentation and type information for symbols.
+- **Go to definition** jumps you to where a function (or other symbol) is defined.
+- **Find references** lists everywhere a function (or other symbol) is called or used.
+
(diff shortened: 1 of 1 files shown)


blogposts/code-navigation-in-github-pull-requests.md
```diff
@@ -0,0 +1,52 @@
...
+
+- What does the function do, and are the arguments correct? Hover over it to see its documentation and signature.
+- Does the function expect its inputs to be sanitized? Go to its definition and find out by reading the source code.
+- Are we already calling this function? If so, are there any helper functions we could reuse? Any usage patterns we should follow for consistency? Find references to the function to get these answers.
+
+## Set it up
+
+You'll need to use Sourcegraph because [GitHub's code navigation](https://help.github.com/en/github/managing-files-in-a-repository/navigating-code-on-github) doesn't work in pull requests. GitHub has no public plans to add support, but if it does, switching should be easy because Sourcegraph's underlying [code intelligence](https://docs.sourcegraph.com/user/code_intelligence) for [20+ languages](https://sourcegraph.com/extensions?query=category%3A%22Programming+languages%22) is open source, uses open standards (including Microsoft's LSP and [LSIF](https://docs.sourcegraph.com/user/code_intelligence/lsif)), and is [vendor neutral](https://about.gitlab.com/blog/2019/11/12/sourcegraph-code-intelligence-integration-for-gitlab/).
```
dadlerj: Is "Microsoft's LSP and LSIF" the correct way to refer to these?

sqs: Yes, I am speaking to people who might worry that when GitHub adds native support for this, their work on Sourcegraph setup would be for nil. Mentioning that they are Microsoft standards increases the likelihood that GitHub would also adopt them and the switching cost would be low.

PR #444, opened 2019-12-08
document why we require checks to pass before merging PRs

sqs:


handbook/engineering/index.md
```diff
@@ -54,3 +54,9 @@ Sourcegraph has a lot of repositories!
 
 - [Business repositories](https://github.com/sourcegraph?utf8=%E2%9C%93&q=repo-type-business&type=&language=)
 - [Customer repositories](https://github.com/sourcegraph?utf8=%E2%9C%93&q=repo-type-customer&type=&language=)
+
+## Misc.
+
+This point lives here for now:
+
+- We require passing checks on GitHub PRs before merging (and don't allow direct pushes to master). Sometimes it's nice to push without waiting for checks (such as for docs-only changes), but this is outweighed by the downside that people too often accidentally merged changes that broke the build. We intend to allow certain kinds of changes (such as docs-only changes) to bypass the full build pipeline so the checks pass quickly in those cases.
```
keegancsmith: we already do that for docs-only changes. We run the mininum checks required (Which is just prettier). However, it is pretty slow due to yarn. We are partially working on that in https://github.com/sourcegraph/sourcegraph/issues/6770

nicksnyder:
```suggestion
- We require passing checks on GitHub PRs before merging (and don't allow direct pushes to master). Sometimes it's nice to push without waiting for checks (such as for docs-only changes), but this is outweighed by the downside that people too often accidentally merged changes that broke the build. Certain kinds of low risk changes (e.g., documentation only changes) may only run a subset of the build pipeline so that checks pass quickly in those cases.
```

PR #445, opened 2019-12-09
Document why we don't share feedback after interviews

nicksnyder:


handbook/people-ops/hiring.md
```diff
@@ -17,3 +17,15 @@ We use [Google Hire](https://hire.withgoogle.com/t/sourcegraphcom/hiring/landing
...
    - All interviewers submit written feedback independently before being able to see feedback from other interviewers.
    - We maintain a consistent hiring decision process for all candidates.
+
+## Sharing feedback
+
+We don't share specific reasons or feedback if we decide not to move forward in our hiring process with a candidate.
+
+We can only interview and hire a finite number of people in a given amount of time, so we move forward with only a few of the best candidates at each stage of our hiring pipeline.
+
+Ultimately, each hiring decision is a complex judgement call based on many inputs.
```
dadlerj:
```suggestion
Ultimately, each hiring decision is a complex one based on many inputs.
```
"judgement call" to me implies lack of decision structure/framework (even though there is always a bit of judgement)

sqs:
What spurred this? The section you added doesn't really explain why we don't. Did you intend for it to explain it, or just state it?

nicksnyder:
> What spurred this?

Two candidates asked for feedback in the last week (and others have asked for it in the past).

At the very least I want to document the policy. It would be nice to cite reasons too, but frankly I was having trouble articulating the reasons so I stopped short of that.

Issue #446, opened 2019-12-10
Document why we don't list specific numbers for experience in job descriptions


sqs:
A candidate asked (on Twitter DM @srcgraph):

> Hey Sourcegraph!
>
> My name is $NAME, and I'm interested in your code intelligence position. I'm considering applying because I'm looking for remote opportunities outside $COUNTRY.
>
> I was just wondering how experienced candidates are you looking for? I've finished my Master's thesis in August, and I've been working as a software engineer for 2 years now.

My reply:

> Hey there, we don’t have any requirement on specific length of experience. We have team members who’ve joined straight from high school/university or after 10+ years of working at somewhere like Google, and all kinds can be successful here. Looking forward to seeing your application!

We should document this for people who read (eg) https://github.com/sourcegraph/careers/blob/master/job-descriptions/software-engineer-code-intelligence.md and wonder why it doesn't list a certain number of years.

nicksnyder:
Where do you think this documentation should live? In the job description itself?

sqs:
Just a section somewhere in the engineering section of the handbook so we can link people to it (and so that team members and others who peruse the handbook internalize this knowledge).

PR #450, opened 2019-12-11
Document why we prefer to avoid requiring a specific numbers for experience in job descriptions

nicksnyder:
closes https://github.com/sourcegraph/about/issues/446

I am proposing putting this in our general documentation, which is perhaps controversial.

The only current job description that diverges from this documentation is [Account Executive](https://github.com/sourcegraph/careers/blob/master/job-descriptions/account-executive.md) which lists "5+ years mid-market/enterprise SaaS sales experience".

Possible resolutions:
- (1) If "5+ years mid-market/enterprise SaaS sales experience" is normal/standard for this kind of role (time in role is in fact important than other measurable skills), then EITHER
    - (a) Adjust wording to "We prefer to avoid" but leave it as general documentation.
    - (b) Move this documentation to the engineering section of the handbook.
- (2) If not (1) then update the AE job description to remove "5+ years mid-market/enterprise SaaS sales experience" in favor of more skill based qualifications, or move it to "Nice-to-have".

Thoughts?


nicksnyder:
I chose (1a) as a solution and am going to merge this because I think it is valuable to document.

PR #459, opened 2019-12-14
docs: update security page

beyang:


beyang:
> In general, Do you think the “security practices” content doesn’t belong here (perhaps in a separate handbook doc?) or that it’s just not ready yet to go in here?

Feel free to add any "security practices" content here, though in the long term, you're right, it feels like something that should be part of the handbook (this page feels more market-y).

PR #476, opened 2019-12-23
New events page to list conferences, meetups, etc

aileenrose:
This is the first draft of a new page.  We will need to rearrange top navigation, possibly introduce a main nav category: Resources  that will include Documentation and Blog, as listed in the Footer.
An alternate category instead of Resources is Learn.  This category could be A/B tested.


ryan-blunden:
Have you considered using Bootstrap cards with an image to make it more visually appealing? 

I'd also recommend adding a description for each event, e.g., who is speaking and about what, or that we're sponsoring.

aileenrose:
> Have you considered using Bootstrap cards with an image to make it more visually appealing?
> 

Yes, this is the first version to get something up quickly.

> I'd also recommend adding a description for each event, e.g., who is speaking and about what, or that we're sponsoring.

First let's get an internal doc about all the committed events. Maybe revive this spreadsheet:
https://docs.google.com/spreadsheets/d/1FLJEjTaRVAUReZ9LSA4VpT2qaL0h_y9fUa38X-WD8jQ/edit#gid=384923095

Let's review the data points needed, this is suggested
Event name, Date, Location, Speaker, Type [Meetup|Conference|Webinar], Short description < 200 characters, link to image
Target audience, Goal, Challenges and Comments will be for internal use only.

I would love help with someone who could write a script that can pull this data via Google API and publish a json file of this info.

ryan-blunden:
You can add [TSConf Japan](https://www.tsconf.jp/2020/) to this list (@felixfbecker  presenting on fully type-safe Web Workers).

ryan-blunden:
> I would love help with someone who could write a script that can pull this data via Google API and publish a json file of this info.

Considering this data won't change once an event is entered, I would keep it simple by embedding the JSON representation of this data in the events page instead.

ryan-blunden:
> Yes, this is the first version to get something up quickly.

Ok, no worries.

website/src/pages/events.tsx
```diff
@@ -0,0 +1,73 @@
...
+        }
+    }
+
+    public render(): JSX.Element | null {
+        const desc =
+            'Meet Sourcegraph at these conferences and meetups around the world.'
+        return (
+            <Layout location={this.props.location}>
+                <div className="Events bg-white text-dark">
+                    <Helmet>
```
ryan-blunden: See https://github.com/sourcegraph/about/blob/master/website/src/pages/case-studies/lyft-monolith-to-microservices.tsx#L8 for an example of not needing Helmet to add page metadata

ryan-blunden:
DId you mean to add a link to this page in the footer nav?

aileenrose:
> DId you mean to add a link to this page in the footer nav?

yes, will do

website/src/pages/events.tsx
aileenrose: oh yeah, addressed this in latest commit!  thanks.

PR #512, opened 2020-01-17
handbook/testing: add permissions note for Percy

rvantonder:
Looks like we don't have a dedicated section for Testing yet. I'm only adding the Percy/permissions part to this for now: we want https://about.sourcegraph.com/search?q=percy to have this info. [Slack reference](https://sourcegraph.slack.com/archives/CMMTWQQ49/p1579300085026700?thread_ts=1579300004.026500&cid=CMMTWQQ49).

### DIFF
--- a/handbook/engineering/testing.md
+++ b/handbook/engineering/testing.md
@@ -6,3 +6,7 @@ This file documents how we test code at Sourcegraph.
 ## Percy
 
 We use [Percy](https://percy.io/) to detect visual changes in Sourcegraph features. You may need permissions to update screenshots if your feature introduces visual changes. Post a message in #dev-chat that you need access to Percy, and someone will add you to our organization (you will also receive an invitation via e-mail). Make sure you are logged in with your GitHub credentials.
+
+## Other
+
+[Documentation for running tests in sourcegraph/sourcegraph](https://sourcegraph.sgdev.org/sourcegraph/sourcegraph/-/blob/doc/dev/testing.md).
--- a/handbook/engineering/testing.md
+++ b/handbook/engineering/testing.md
@@ -3,6 +3,6 @@
 This file documents how we test code at Sourcegraph.
 
 
-### Percy
+## Percy
 
 We use [Percy](https://percy.io/) to detect visual changes in Sourcegraph features. You may need permissions to update screenshots if your feature introduces visual changes. Post a message in #dev-chat that you need access to Percy, and someone will add you to our organization (you will also receive an invitation via e-mail). Make sure you are logged in with your GitHub credentials.
--- a/handbook/engineering/testing.md
+++ b/handbook/engineering/testing.md
@@ -1,0 +1,9 @@
+# Testing
+
+This file documents how we test code at Sourcegraph. See [permissions](#permissions) for help on accessing our testing tools. 
+
+## Permissions
+
+### Percy
+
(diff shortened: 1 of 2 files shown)


rvantonder:
These changes don't show up in the [netlify preview](https://deploy-preview-512--sourcegraph.netlify.com/handbook/engineering), any idea why they wouldn't?

handbook/engineering/testing.md
```diff
@@ -0,0 +1,9 @@
+# Testing
+
+This file documents how we test code at Sourcegraph. See [permissions](#permissions) for help on accessing our testing tools.
```
nicksnyder: I don't think we need links within the document like this, especially when the document is so small.

```suggestion
This file documents how we test code at Sourcegraph.
```

handbook/engineering/testing.md
```diff
@@ -0,0 +1,9 @@
+# Testing
+
+This file documents how we test code at Sourcegraph. See [permissions](#permissions) for help on accessing our testing tools. 
+
+## Permissions
+
+### Percy
```
nicksnyder:
```suggestion
## Percy
```

I think Percy would be the top level item in this document (and maybe a permissions item under it instead, but I don't think there is enough content to warrant that).

handbook/engineering/testing.md
```diff
@@ -0,0 +1,9 @@
+# Testing
+
+This file documents how we test code at Sourcegraph. See [permissions](#permissions) for help on accessing our testing tools. 
+
+## Permissions
```
nicksnyder:
```suggestion
```

rvantonder:
@nicksnyder hmmm good point. Searching on **docs.sourcegraph.com** finds details about percy: https://docs.sourcegraph.com/search?q=percy&v=. I'm not sure what the right move is, the prompt behind this PR is to have a search for "percy" work on about.sourcegraph.com.

handbook/engineering/testing.md
```diff
@@ -0,0 +1,8 @@
+# Testing
+
+This file documents how we test code at Sourcegraph.
+
+
+## Percy
+
+We use [Percy](https://percy.io/) to detect visual changes in Sourcegraph features. You may need permissions to update screenshots if your feature introduces visual changes. Post a message in #dev-chat that you need access to Percy, and someone will add you to our organization (you will also receive an invitation via e-mail). Make sure you are logged in with your GitHub credentials.
```
nicksnyder: Hmm...so docs.sourcegraph.com is for product docs so developer related documentation doesn't really belong there (e.g. we should perhaps move it out of the docs directory, but you don't necessarily need to do this yourself).

I agree that the handbook should have *some* page that talks about testing, but it also feels wrong to include docs that are hyper specific to sourcegraph/sourcegraph (which is what https://sourcegraph.sgdev.org/sourcegraph/sourcegraph/-/blob/doc/dev/testing.md is).

How about we just link to it from here. I think that would resolve my concerns for now.

```suggestion
We use [Percy](https://percy.io/) to detect visual changes in Sourcegraph features. You may need permissions to update screenshots if your feature introduces visual changes. Post a message in #dev-chat that you need access to Percy, and someone will add you to our organization (you will also receive an invitation via e-mail). Make sure you are logged in with your GitHub credentials.

## Other

[Documentation for running tests in sourcegraph/sourcegraph](https://sourcegraph.sgdev.org/sourcegraph/sourcegraph/-/blob/doc/dev/testing.md).
```

rvantonder: Works for me!

PR #516, opened 2020-01-20
Add jargon file

efritz:
We should have a place that defines some common terms that are either unknown during onboarding or unknown when members of different teams are communicating.

### DIFF
--- a/handbook/communication/jargon.md
+++ b/handbook/communication/jargon.md
@@ -1,0 +1,24 @@
+# Jargon file
+
+This page is a collection of terms and phrases (roughly categorized) which have a specific internal meaning. This list should be treated as a living document and should be updated when new terms come into use, when old terms are given a new or altered meaning, and when old terms are met with confusion (e.g during onboarding).
+
+Some terms in this document may only be useful for internal communication. For concepts which are also user-facing, the public description should be used over any alternate internal description.
+
+## General terminology
+
+This section contains terminology that was defined outside of and is not specific to Sourcegraph, but is used to describe Sourcegraph features and technology.
+
+### Code intelligence
+
+- **Language Servers** are a long-running program that performs semantic analysis over code and answers code intelligence queries (e.g. find definitions, find refrences, get hover text) from clients.
+- **Language Server Protocol (LSP)** is the RPC-based protocol used between Language Servers and clients.
+- **Language Server Index Format (LSIF)** is a specification by Microsoft that captures a subset of data returned by queries to a language server. It allows queries to be answered without running a language server.
+
+## Sourcegraph-specific
+
+This section contains terminology that is defined within Sourcegraph or is specific to a Soucegraph feature or technology.
+
+### Code intelligence
+
+- **Precise code intelligence** (find definitions, find references, get hover text) is provided by semantic analysis over source code (currently LSIF or LSP). Answers provided by code intelligence can assumed to be correct.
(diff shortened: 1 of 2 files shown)


sqs:
https://about.sourcegraph.com/handbook/people-ops/from-graphbook/glossary

If you didn’t find that doc when you searched, I fear the next person also wouldn’t find this new jargon doc. I would prefer defining terms inline and linking, and avoiding jargon where possible, to making a glossary. If you really feel it’s necessary, then try it, but be open to removing it if nobody consults it.

efritz:
> https://about.sourcegraph.com/handbook/people-ops/from-graphbook/glossary

I feel like this glossary has a different purpose than a jargon file, but I'm open to discussion on this point. This page specifically says it's developer lingo for non-developers and is specifically nested in people-ops (where I wouldn't necessarily have looked as an engineer).

This list seems like a crash course on how engineers talk to each other with terms that are pretty well-established in the domain. But we _don't_ yet (as far as I know) have a similar glossary for terms that we've come over time to attribute a specific meaning to that a non-Sourcegrapher might not intuitively know.

The impetus for this change was that I saw when talking to a new engineer that we were trying to discuss the same feature but we were calling them different things and ended up talking past each other (precise code intel/basic code intel (our terms) vs indexed and unindexed code intel (their terms)). Also, my first entire week I just kept nodding when people used the term LSIF because I figured that was something that I should have already known and was going to look up later.

Maybe this isn't the right place for such a document, but I do want to address the term.

sqs:
@efritz I understand your point, but I am still skeptical. (GitLab also experienced the same problem I am warning about: https://gitlab.com/gitlab-com/www-gitlab-com/issues/1731.) It's fine to merge and try it out, but I just want to lay this out beforehand so we can revisit it in a month or so and see if it's working.

efritz:
I'll continue to entertain this PR if and only if other teams have a similar issue. If it's only code intel, then obviously our team needs to document things in the correct place and be able to point to it during onboarding.

handbook/communication/jargon.md
```diff
@@ -0,0 +1,24 @@
...
+- **Language Server Index Format (LSIF)** is a specification by Microsoft that captures a subset of data returned by queries to a language server. It allows queries to be answered without running a language server.
+
+## Sourcegraph-specific
+
+This section contains terminology that is defined within Sourcegraph or is specific to a Soucegraph feature or technology.
+
+### Code intelligence
+
+- **Precise code intelligence** (find definitions, find references, get hover text) is provided by semantic analysis over source code (currently LSIF or LSP). Answers provided by code intelligence can assumed to be correct.
+- **Search-based code intelligence** (find definitions, find references, get hover text) is provided by the basic code intelligence extension and is powered by code search. It uses heuristics (file extensions, import parsing) to prune the result set of obviously incorrect search results. This type of code intelligence can have many false positive results. This type of code intelligence used to be referred to internally as _basic_ or _fuzzy code intelligence_.
```
uwedeportivo: s/many false positive/false positive

dadlerj: Our docs use "basic code intelligence": https://docs.sourcegraph.com/user/code_intelligence

I believe at one point we tried to replace this with "fuzzy" though.

Worth noting that for this type of jargon/definition, our style guide often provides guidance as well (that's where I go by default for any questions about product naming conventions): https://about.sourcegraph.com/handbook/communication/style_guide

sqs:
@efritz Do you plan to merge this? If not, please close it.

PR #518, opened 2020-01-20
docs: consolidate security policy

nicksnyder:
This PR does three things:
1. Remove a link to HackerOne since we don't use that.
2. Document _when_ we issue rewards. My proposal is after we have confirmed the issue is fixed.
3. I found 3 different security related READMEs and I think there should be only 1, so I consolidated the content (see also https://github.com/sourcegraph/sourcegraph/pull/7906).


docs/security.md
```diff
@@ -36,10 +50,23 @@ Sourcegraph supports HTTPS encryption when deployed on-premises.
...
+- Tools such as Dependabot and GitHub security advisories to identify security vulnerabilities in our code and in dependencies
+
+All development laptops have encrypted hard drives.
+
 ## Bug reports
 
 If you think that you have found a security issue, please email us at <a href="mailto:security@sourcegraph.com">security@sourcegraph.com</a>. Please do not publicly disclose the issue until we’ve addressed it.
 
-We provide monetary rewards, up to $10,000, for open bounties. This is determined based on the percentage of users impacted, the likelihood of encountering the vulnerability under normal use of the site, and the severity of potential service disruption or data leakage.
+We provide monetary rewards, up to $10,000, for reporting security issues. This is determined based on the percentage of users impacted, the likelihood of encountering the vulnerability under normal use of the site, and the severity of potential service disruption or data leakage. Bounties will be awarded after the issue is confirmed fixed.
```
tsenart:
```suggestion
We provide monetary rewards, up to $10,000, for reporting security issues. This is determined based on the percentage of users impacted, the likelihood of encountering the vulnerability under normal use of the product, and the severity of potential service disruption or data leakage. Bounties will be awarded after the issue is confirmed fixed.
```

PR #534, opened 2020-01-29
handbook eng: simulate k8s admin security restrictions

uwedeportivo:
describe how to simulate managing a sourcegraph instance in k8s with restricted admin powers subject to pod security policies

### DIFF
--- a/handbook/engineering/distribution/k8s_admin_custom_policy.md
+++ b/handbook/engineering/distribution/k8s_admin_custom_policy.md
@@ -49,44 +49,53 @@ Save this yaml into `nonroot-policy.yaml` and add the policy to the cluster:
 kubectl apply -f nonroot-policy.yaml 
 ```
 
+From here on out we will do almost everything in a namespace. This is not strictly necessary (it works in the default namespace too)
+but it is cleaner and we're migrating to namespaces anyways.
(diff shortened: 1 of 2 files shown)


handbook/engineering/distribution/k8s_admin_custom_policy.md
```diff
@@ -0,0 +1,92 @@
...
+```
+
+First we need a pod security policy that matches as close as possible the restrictions at the customer site. For example
+(this is just an example, tweak as necessary):
+
+```yaml
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: foopolicy
```
nicksnyder: Can we avoid using `foo` here an elsewhere? It makes it seem less realistic. If you were actually going to configure a policy for our cluster, what would you call it?

uwedeportivo: fixed

uwedeportivo:
> Is this something that only distribution teammates will need to do?

Probably. Distribution needs to test that our configs work without requiring unreasonable privileges.

uwedeportivo:
i'm going to update this with namespaces before merging

PR #546, opened 2020-02-05
define IARR, new/expansion IARR

sqs:
I propose we use IARR (defined in this change) instead of "net new ARR" as we have been saying. This is to avoid confusion; "new ARR" can be interpreted as "incremental ARR" or "new logo ARR", which are 2 very different things. Now that we are distinguishing between new and expansion ARR much more than in the past, this confusion is more likely to occur.

Note (for the curious): GitLab uses the terms "new IACV" and "growth IACV" (https://about.gitlab.com/handbook/sales/#annual-contract-value-acv). We use the term "expansion" instead of "growth". If anyone has a preference for "growth" instead, speak up.

### DIFF
--- a/handbook/sales/index.md
+++ b/handbook/sales/index.md
@@ -31,6 +31,20 @@ The Sales team represents us and our values to customers, bringing back dollars 
 
 Annual Recurring Revenue (ARR) is the dollar value of contracted recurring revenue in a (normalized) one-year period.
 
+### IARR
+
+Incremental [ARR](#arr) (IARR) is the change in ARR from one period to another.
+
+### New IARR
+
+New [IARR](#iarr) is IARR from *new customers* (i.e., organizations that were **not** existing customers at the beginning of the period). See also [expansion IARR](#expansion-iarr).
+
+### Expansion IARR
+
+Expansion [IARR](#iarr) is IARR from *existing customers* (i.e., organizations that were already customers at the beginning of the period).
+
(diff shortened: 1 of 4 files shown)


handbook/sales/index.md
```diff
@@ -27,6 +27,20 @@ The Sales team represents us and our values to customers, bringing back dollars
...
+
+### New IARR
+
+New [IARR](#iarr) is IARR from *new customers* (i.e., organizations that were **not** existing customers at the beginning of the period). See also [expansion IARR](#expansion-iarr).
+
+### Expansion IARR
+
+Expansion [IARR](#iarr) is IARR from *existing customers* (i.e., organizations that were already customers at the beginning of the period).
+
+If within a single period a new customer signs a contract which then grows in ARR before the end of the period, the total ending ARR is all considered [new IARR](#new-iarr), not expansion IARR. For example, if Acme Corp signs a $100k contract on February 3 and then the contract expands to $200k on March 5, all $200k would be considered new IARR for Q1.
```
dadlerj: Random note, but we use a mix of capital and lower case characters for currency/number notations (i.e., the "k", "m", and "b"s on dollar amounts): https://sourcegraph.com/search?q=repo:%5Egithub.com/sourcegraph/about%24+%5C%24%5B0-9%5C.%5D%2B%5BkKmMbB%5D+lang:markdown+&patternType=regexp&case=yes

Over time I've noticed that you default to lower case, so I've proposed this in https://github.com/sourcegraph/about/pull/548

sqs: @dadlerj Thanks, I updated some more mentions elsewhere but kept those that were historical.

sqs:
@juliasourceress Does this look good to you?

PR #548, opened 2020-02-05
Clarify number suffixes in style guide

dadlerj:


handbook/communication/style_guide.md
```diff
@@ -152,6 +152,16 @@ For consistency, all examples should use the following names (as appropriate).
 - Prefer "call site" (not "callsite")
 - Prefer "tooltips" (not "tool-tips")
 
+### Numbers
+
+Use shorthand suffixes for shortening numbers in the thousands ("k"), millions ("m"), billions ("b"), and above. Prefer lower case suffixes.
```
creachadair: These suggestions are fine when there is a currency mark, as in the examples below. When the numbers in question are not currency, however, but unlabelled values, these suffixes are ambiguous (e.g., 100k could mean 100x10^3 or 100x2^10, depending on the units at play). And in that case "b" is no longer the correct SI label for 10^9.

In some cases the distinction doesn't matter, but since the goal is clarity I think we need to tease this apart.

One simple fix: Change the title of this section to "Currency". Then I think these suggestions are fairly consistent (though I still think the capitalization guideline is excessive).

handbook/communication/style_guide.md
```diff
@@ -152,6 +152,16 @@ For consistency, all examples should use the following names (as appropriate).
 - Prefer "call site" (not "callsite")
 - Prefer "tooltips" (not "tool-tips")
 
+### Numbers
```
dadlerj:
```suggestion
### Currency
```

PR #560, opened 2020-02-06
Update people ops index

dadlerj:


dadlerj:
@mercadon bump on this

PR #571, opened 2020-02-10
Update travel policy and add new team

dadlerj:


dadlerj:
FYI @mercadon some updates to travel budgets

PR #579, opened 2020-02-13
add HubSpot and email logging step to sales onboarding

sqs:
To avoid problems like https://sourcegraph.slack.com/archives/C0B2RU51Q/p1581598734414000?thread_ts=1581542364.391200&cid=C0B2RU51Q.


sqs:
@kpsource Thanks - I didn't know that doc existed. I will add a link to it from the handbook. @juliasourceress at some point you should consolidate these 2 docs so there is just 1 (ideally all content is in the handbook, or else the handbook links to the single Google doc).

PR #580, opened 2020-02-13
Mock Calls

kpsource:
Adding in an event for week 2 of sales onboarding

### DIFF
--- a/handbook/sales/onboarding/index.md
+++ b/handbook/sales/onboarding/index.md
@@ -77,6 +77,8 @@ See also [Sales onboarding logistics](https://docs.google.com/document/d/1un9fFP
 
 ## Week 2
 
+- Mock calls
+  - Schedule internal mock calls to practice discovery questiions and use cases
 - Checkpoint
   - Complete a 5-minute live product demo
 


handbook/sales/onboarding/index.md
```diff
@@ -77,6 +77,8 @@ See also [Sales onboarding logistics](https://docs.google.com/document/d/1un9fFP
 
 ## Week 2
 
+- Mock calls
+  - Schedule internal mock calls to practice discovery questiions and use cases
```
sqs:
```suggestion
  - Schedule internal mock calls to practice discovery questions and use cases
```

juliasourceress: @kpsource awesome, thanks! Could you also add set up time with various team members to hear their pitch of Sourcegraph and our value? (Prior to doing own mock pitches). I'd get a distribution engineer, someone in marketing, and @sqs 's pitch in. We can also get these recorded so the step would be to watch it if this gets too repetitive/time intensive for the team.

kpsource: @juliasourceress Yes!

PR #582, opened 2020-02-13
Add manager perspective of peer feedback

nicksnyder:
I was a bit confused as to what to write in the summary box, so I have been using it to offer my perspective on the peer feedback the person received. I don't think this needs to be a separate step though, so I am proposing adding a third question instead.

### DIFF
--- a/handbook/people-ops/review-cycles.md
+++ b/handbook/people-ops/review-cycles.md
@@ -30,9 +30,9 @@ Review cycles take place in February and August. They begin the first week of ea
 - Manager review: you will write a review of your manager.
 - Peer reviews: you will write reviews of any peers that selected you to write feedback for them.
 
-### Manager feedback + summaries (9 days)
+### Manager feedback (9 days)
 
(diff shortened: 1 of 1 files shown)


handbook/people-ops/review-cycles.md
```diff
@@ -59,3 +59,4 @@ Once the reviews are complete for each of your direct reports, you’ll be able
 
 - Describe this person’s impact at Sourcegraph and since the last review cycle and how that impact compares to your expectations.
 - What are this person’s biggest opportunities for personal development and growth moving forward?
+- What are your thoughts on the peer feedback that this person received?
```
dadlerj: This is a bit of an odd question to me — I feel like peer feedback should be factored in to a summary along with the manager's feedback. That said, I'm fine removing the extra step, I'm just not sure an additional question is necessary here... It feels like the two questions above this should factor that feedback in.

nicksnyder: Ok, I wasn't sure if it should or shouldn't. I am fine removing this question and just assuming that these boxes should factor in peer feedback. If so, the summary question feels particularly redundant and I think we could be more explicit about the fact that peer feedback should be considered in the manager's review.

dadlerj: Agreed! Sorry, didn't mean for you to close the PR. I agree that removing the question makes sense, and being more clear, like you said.

PR #590, opened 2020-02-18
Add BigQuery/ping documentation

ebrodymoore:


handbook/engineering/adding_ping_data.md
```diff
@@ -23,8 +23,19 @@ You will be asked the following questions:
...
       - Pings documentation: https://docs.sourcegraph.com/admin/pings
       - The Site-admin > Pings page, e.g.: https://sourcegraph.com/site-admin/pings
-3. Request help from the Business Operations team to add the new field(s) to BigQuery and Looker. 
+3. Determine if any transformations/ETL jobs are required, and if so, add them to the [script](https://console.cloud.google.com/storage/browser/_details/sg-analytics-data/dataflow/pipelines/udf/transform.js?project=telligentsourcegraph&authuser=0&angularJsUrl=%2Fstorage%2Fbrowser%2F_details%2Fsg-analytics-data%2Fdataflow%2Fpipelines%2Fudf%2Ftransform.js%3Fproject%3Dtelligentsourcegraph%26authuser%3D1).
+4. Open a PR to change [the schema](https://github.com/sourcegraph/analytics/tree/master/BigQuery%20Schemas) with Business Operations (EricB and Dan) as approvers. Keep in mind:
+	- Check the data types sent in the JSON match up with the BigQuery schema (e.g. a JSON '1' will not match up with a BigQuery integer). 
+	- Every field in the BigQuery schema should be `"mode": "NULLABLE"`. There will be instances on the older Sourcegraph versions that will not be sending new data fields, and this will cause pings to fail.
```
dadlerj: I'd rephrase this as "no fields should be non-nullable" (though I don't know what the actual term in BigQuery is for this). The reason is that there is a third option, which is fine to use, called "REPEATED" — where it accepts a list of values of any length (so it can't be `NULL`, but it can be `[]` or an empty list).

handbook/engineering/adding_ping_data.md
```diff
@@ -23,8 +23,19 @@ You will be asked the following questions:
...
+
+
+## Changing the BigQuery schema
+
+1. Use the merged JSON to update the schema in the test table `bq --project_id=telligentsourcegraph update --schema schema.json sourcegraph_analytics.update_checks_backup`
```
dadlerj:
```suggestion
1. Use the merged JSON to update the schema in the test table `bq --project_id=$PROJECT update --schema $SCHEMA_FILE $DATASET.$TABLE`, replacing `$PROJECT` with the project ID, `$SCHEMA_FILE` with the path to the schema JSON file generated above, and `$DATASET.$TABLE` with the dataset and table name, separated by a dot. 
```

dadlerj: Actually, now that I've read this and the number below, perhaps this change isn't exactly correct... Perhaps have a description of this command above the bullets, and then bullet 1 can be "Run it on a test table and ensure the schema matches expectations" and bullet 2 can be "Run it on the production table".

handbook/engineering/adding_ping_data.md
```diff
@@ -23,8 +23,19 @@ You will be asked the following questions:
...
+2. Once the test is complete and the changes are ready to be implemented, run the same command on the table `bq --project_id=telligentsourcegraph update --schema schema.json sourcegraph_analytics.update_checks`
+
+To get the current schema for any reason, use `bq --project_id=telligentsourcegraph --format=prettyjson show sourcegraph_analytics.update_checks > schema.json`
```
dadlerj: Update this to use variables as well

handbook/engineering/adding_ping_data.md
creachadair: BigQuery is built on top of protobuf (specifically proto2), so what it _really_ means is `required`. It's safe to convert nullable (`optional`) to repeated, but not vice versa; and of course `required` is forever.

PR #593, opened 2020-02-19
Add antiflicker for AB testing with Optimize

aileenrose:
Adding script for running A/B tests only on home page.

### DIFF
--- a/website/src/pages/index.tsx
+++ b/website/src/pages/index.tsx
@@ -228,4 +228,4 @@ export default ((props: any) => (
             </Jumbotron>
         </div>
     </Layout>
-)) as React.FunctionComponent < any >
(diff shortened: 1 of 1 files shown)


aileenrose:
Ran Optimize diagnostics and passed tests.

website/src/pages/index.tsx
```diff
@@ -222,4 +228,4 @@ export default ((props: any) => (
             </Jumbotron>
         </div>
     </Layout>
-)) as React.FunctionComponent<any>
+)) as React.FunctionComponent < any >
```
ryan-blunden:
```suggestion
)) as React.FunctionComponent<any>
```

This should be reverted.

PR #598, opened 2020-02-20
Let's start tracking clicks on signin and signup

aileenrose:
I added onclick events to Signup and Signin in the navigation.  This will help with optimizing content i.e. A/B testing.


aileenrose:
No longer required since we can use GTM for eventlogging.  See #616 https://github.com/sourcegraph/about/pull/616

PR #608, opened 2020-02-24
Update GitLab code intel announcement content

ryan-blunden:
The GitLab announcement is no longer "new" so move it down the page and move the tweets up.


aileenrose:
Looking at the heatmap, I would recommend removing the tweets section altogether ...adding it the /product page under the customer logos (replacing the two quotes from Lyft and Uber.
https://app.crazyegg.com/sharing/a057ba49cd171f99

We are making more changes to the content for the March 3 rollout.   I could remove it then.  Updating the heading for the GitLab video is important, so I will approve.

PR #616, opened 2020-02-26
Remove Eventlogger since GTM can manage tracking and logging events

aileenrose:
Dismantled Eventlogger, removing all tracking of click elements from various pages.  Google Tag Manager GTM can easily be configured to track and log any interactivity on a web page.  BTW, using Sourcegraph made it easy to find all the Eventlogger references.


ryan-blunden:
> Google Tag Manager GTM can easily be configured to track and log any interactivity on a web page

Does this mean that it **is** configured to track and log the same events that are removed in this PR?

PR #623, opened 2020-03-02
Add a mission statement for Campaigns

mrnugget:
In the past couple of weeks I've been thinking about Campaigns, what it is and isn't, where it should go and what we, as engineers, can do to support that.

Call it an engineering vision or a mission statement, I'm not set on a name, and it's still rough and I want to polish it (with everybody's input, of course).

I do think it's valuable to have something like this in place, because it helps us make priority decisions, especially from an engineering perspective. Ever since I came up with these analogies (see PR) for myself, it got a lot easier for me to decide what to work on and what not.

---

I just put this in `index.md` because I couldn't find a better place to put it, but I'm happy to move it.

### DIFF
--- a/handbook/engineering/campaigns.md
+++ b/handbook/engineering/campaigns.md
@@ -11,7 +11,7 @@ When I write an AWS Lambda function I want to focus on which requests it receive
 
 When I deploy a static site on Netlify I want to focus on my site — its content and design — and not think about where it's deployed, how to get new SSL certificates, how to install dependencies to run the static site generator on a server, how to preview the site in a pull request.
 
-**Campaigns allow users to focus on finding code that needs to be changed and changing it by running code**.
+**Find code that needs to be changed and change it by running code**.
 
 Campaign users don't have to worry about:
 
--- a/handbook/engineering/campaigns.md
+++ b/handbook/engineering/campaigns.md
@@ -5,12 +5,6 @@ Developer documentation: https://docs.sourcegraph.com/dev/campaigns_development
 
 ## Mission Statement
 
-Campaigns are to large-scale code changes what Netlify is to static site generation and AWS Lambda is to handling HTTP requests.
-
-When I write an AWS Lambda function I want to focus on which requests it receives and what response to send out. I don't want to worry about which server to run it on, how to scale it, secure it, add logging, keep track of its usage.
-
-When I deploy a static site on Netlify I want to focus on my site — its content and design — and not think about where it's deployed, how to get new SSL certificates, how to install dependencies to run the static site generator on a server, how to preview the site in a pull request.
-
 **Find code that needs to be changed and change it by running code**.
 
 Campaign users don't have to worry about:
@@ -22,3 +16,9 @@ Campaign users don't have to worry about:
 * Create pull requests from patches
 * Draft, keep track of and update a large number of pull requests
 * Re-running code when the base branch changes
+
+Campaigns are to large-scale code changes what Netlify is to static site generation and AWS Lambda is to handling HTTP requests.
+
+When I write an AWS Lambda function I want to focus on which requests it receives and what response to send out. I don't want to worry about which server to run it on, how to scale it, secure it, add logging, keep track of its usage.
+
+When I deploy a static site on Netlify I want to focus on my site — its content and design — and not think about where it's deployed, how to get new SSL certificates, how to install dependencies to run the static site generator on a server, how to preview the site in a pull request.
--- a/handbook/engineering/campaigns.md
+++ b/handbook/engineering/campaigns.md
@@ -7,7 +7,7 @@ Developer documentation: https://docs.sourcegraph.com/dev/campaigns_development
 
 **Find code that needs to be changed and change it by running code**.
 
-Campaign users don't have to worry about:
+Users can focus on the changes to their code because Campaigns solve the following problems for you: 
 
 * Finding the correct repositories in which to run code
 * Fetching the newest version of each repository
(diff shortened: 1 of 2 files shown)


tsenart:
I love the analogies! LGTM

handbook/engineering/campaigns.md
```diff
@@ -0,0 +1,23 @@
+# Campaigns
+
+User-facing documentation: https://docs.sourcegraph.com/user/campaigns
+Developer documentation: https://docs.sourcegraph.com/dev/campaigns_development
+
+## Mission Statement
+
+Campaigns are to large-scale code changes what Netlify is to static site generation and AWS Lambda is to handling HTTP requests.
```
eseliger: like it, but `large-scale code changes` would omit our second proposed use case that doesn't generate `code-changes`. I'm not against this though, just a side-note

mrnugget: Good point! In one sense this use case can be part of the list: users don't have to worry about their changes staying up to date with ever updating code.

The other sense of that use case is still under-defined, so I don't think it's necessary to include it _yet_. WDYT?

eseliger: absolutely, just wanted to kick that thought off so we are on the same page that it might be a use-case that campaigns will pick up in the future :)

mrnugget: Added another point!

mrnugget:
Little update: with the help of @adamfrankl2015 we now have a punchied version of what Campaigns allow a user to do:

> **Find code that needs to be changed and change it by running code**.

---

@nicksnyder and @felixfbecker pointed out that this version here does not include the to-be-built functionality of Campaigns that allows users to "enforce invariants" in their codebase. I refrained from putting that in, because to me it still contains a lot of question marks and I don't think I can come up with a version that's adequate, giving us enough input to help with day-to-day decision making. But I do think that Adam's version ("Find code that needs to be changed") also includes that, in a certain sense, and the bullet point I added also touches on the "rerunning" aspect of it.

I'd merge this PR, keep this in the back of my head and once I can come up with something better, I'll edit this.

felixfbecker:
One thought I had since our discussion in the sync is the "by running code" part seems like an unnecessary restriction - while we could argue that everything is essentially running code at one layer or another, it's not necessarily how I'd think about the problem. For example, I wouldn't think of a comby search+replace as "running code" or invoking a refactor code action

mrnugget:
> One thought I had since our discussion in the sync is the "by running code" part seems like an unnecessary restriction - while we could argue that everything is essentially running code at one layer or another, it's not necessarily how I'd think about the problem. For example, I wouldn't think of a comby search+replace as "running code" or invoking a refactor code action

That was _exactly the point_ of this 🙂 One of the things that kicked this of was [this message by @sqs](https://sourcegraph.slack.com/archives/CMMTWQQ49/p1582209486256700?thread_ts=1582034798.229000&cid=CMMTWQQ49):

> Customers will definitely want to use Docker directly and not just any high-level APIs for changing code (eg “upgradeNpmDependency foo 1.3.5”). I’m very confident of that. It’s one of the first questions they ask in all of my presentations in the last 2 weeks and all have been reassured by me saying we’re not trying to come up with a magic code changer, we’re letting you write commands and then we scale that up to all your code. 

One of the reasons for this PR is to document this: we just run Docker/commands for you and you don't have to worry about the plumbing,no "magic code changer".

And at the moment and for the foreseeable future we _only_ offer running Docker containers and user-supplied commands. We don't offer "comby search+replace" and we don't offer "refactor code action": you can run a Docker `comby` container or an `eslint` command, but for us it's just running what the user supplied.

And if we leave that out and make what Campaigns is about even more generic, to have no "unnecessary restriction", we might run in danger of saying nothing at all, losing all practical benefit in decision helping that this document might provide.

handbook/engineering/campaigns.md
```diff
@@ -0,0 +1,24 @@
+# Campaigns
+
+User-facing documentation: https://docs.sourcegraph.com/user/campaigns
+Developer documentation: https://docs.sourcegraph.com/dev/campaigns_development
+
+## Mission Statement
+
+**Find code that needs to be changed and change it by running code**.
+
+Campaign users don't have to worry about:
```
christinaforney:
```suggestion
Users can focus on the changes to their code because Campaigns solve the following problems for you: 
```

handbook/engineering/campaigns.md
```diff
@@ -0,0 +1,24 @@
...
+
+* Finding the correct repositories in which to run code
+* Fetching the newest version of each repository
+* Running code in each repository
+* Turn the result into patches
+* Create pull requests from patches
+* Draft, keep track of and update a large number of pull requests
+* Re-running code when the base branch changes
+
+Campaigns are to large-scale code changes what Netlify is to static site generation and AWS Lambda is to handling HTTP requests.
```
christinaforney: I think we can actually explain this a little bit on top of the analogy. 

```suggestion
Netlify and AWS Lambda solve difficult, repeatable problems for developers, removing overhead and enabling them to focus on the problems they are solving. In that regard, Campaigns are to large-scale code changes what Netlify is to static site generation and AWS Lambda is to handling HTTP requests.
```

handbook/engineering/campaigns.md
```diff
@@ -0,0 +1,24 @@
...
+
+When I write an AWS Lambda function I want to focus on which requests it receives and what response to send out. I don't want to worry about which server to run it on, how to scale it, secure it, add logging, keep track of its usage.
+
+When I deploy a static site on Netlify I want to focus on my site — its content and design — and not think about where it's deployed, how to get new SSL certificates, how to install dependencies to run the static site generator on a server, how to preview the site in a pull request.
```
christinaforney: It feels like we should extend these to also talk about what we want from campaigns directly, and in the same fashion as the other two analogies. I'm not attached to my proposed wording, so please edit to make it better :) 

```suggestion
When I deploy a static site on Netlify I want to focus on my site — its content and design — and not think about where it's deployed, how to get new SSL certificates, how to install dependencies to run the static site generator on a server, how to preview the site in a pull request.

When I create a campaign I want to focus on the specific change I want to make across all of the code at my organization. I don't want to worry about all of the overhead associated with execution, code hosts, and management of all things listed above. 
```

PR #628, opened 2020-03-03
Edit Campaigns page and refine Vision/Mission

mrnugget:
After the discussions yesterday I think it makes sense to include what
we _don't_ do. See my comment here: https://github.com/sourcegraph/about/pull/623#issuecomment-593560930

I'm still not sure about the "Vision"/"Mission" naming and this still
does not include the things that might at some point come in the future
or anticipate things that might change the vision (like
linters/diagnostics/etc.), but I wanted to add this so that we have this
written down somewhere.

As always, we can still change it.

### DIFF
--- a/handbook/engineering/campaigns.md
+++ b/handbook/engineering/campaigns.md
@@ -3,11 +3,13 @@
 User-facing documentation: https://docs.sourcegraph.com/user/campaigns
 Developer documentation: https://docs.sourcegraph.com/dev/campaigns_development
 
-## Mission Statement
+## Vision
 
 **Find code that needs to be changed and change it by running code**.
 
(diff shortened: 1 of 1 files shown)


handbook/engineering/campaigns.md
```diff
@@ -17,10 +19,18 @@ Users can focus on the changes to their code because Campaigns solve the followi
 * Draft, keep track of and update a large number of pull requests
 * Re-running code when the base branch changes
 
+Users provide the code to make the change, we provide the plumbing to turn it into a large-scale change and monitor its progress.
+
+* We take care of all the bits and pieces that would rob users of their time and that are not essential to the change they want to make.
```
ryanslade:
```suggestion
* We take care of all the details that would rob users of their time and that are not essential to the change they want to make.
```

PR #641, opened 2020-03-05
Move duplicated HubSpot code into HubSpot component

ryan-blunden:


ryan-blunden:
>  I'm not seeing it being used anywhere after adding the new component.

That's where the **Talk to a product specialist** links go to.

PR #642, opened 2020-03-05
Add newsletter signup form to footer

ryan-blunden:
Screen Shot 2020-03-05 at 4 45 52 PM


ryan-blunden:
I think it's ok to merge so long as it's not detrimental to the design, but agree that it doesn't stand out and could be improved. @aileenrose What do you think?

aileenrose:
> Could we put this in a box instead of subtly showing it with the same background as the footer navigation? When I first saw it, I thought it was just some legal disclaimer and didn't realize it was a CTA to sign up for the newsletter.

@vanesa - we can optimize the design over time.  I would like to create additional CTAs on the blog landing page - which needs a bit of updating.  :)

ryan-blunden:
@aileenrose Can we redirect form submissions to a page on our site, as at the moment, it redirects to the ebook landing page. 

Or better yet, is there a way we can submit without the user being redirected anywhere?

PR #649, opened 2020-03-11
eng: how to set up HTTPS for local instance

uwedeportivo:
Describes how to run a local Sourcegraph behind a HTTPS endpoint.

### DIFF
--- a/handbook/engineering/https_dev_instance.md
+++ b/handbook/engineering/https_dev_instance.md
@@ -6,53 +6,35 @@
 
 ## Method One
 
-> Super easy to install but makes a roundtrip to an outside (AWS server) for what is basically a local request.
-> On the other hand you can invite your co-workers to use your instance by sharing the ngrok.io URL.
-> This is useful for shared debugging, demos and presentations.
-
-This method uses [ngrok](https://ngrok.com/).
-
-1. Follow [instructions](https://ngrok.com/download) for downloading and setting up ngrok.
-1. Start your instance (`dev/start.sh` or `enterprise/dev/start.sh`)
-1. Run this command: `ngrok http 3080`
-
-Output will look something like this:
-
-```text
-ngrok by @inconshreveable                                                                                                            
-                                                                                                                                                                                                                        
-Session Status                online                                                                                                                                                                                    
-Account                       xxxxxx (Plan: xxxxx)                                                                                                                                                                 
-Version                       2.3.35                                                                                                                                                                                    
-Region                        United States (us)                                                                                                                                                                        
-Web Interface                 http://127.0.0.1:4040                                                                                                                                                                     
-Forwarding                    http://xxxxxx.ngrok.io -> http://localhost:3080                                                                                                                                         
-Forwarding                    https://xxxxxxx.ngrok.io -> http://localhost:3080                                                                                                                                        
-                                                                                                                                                                                                                        
-Connections                   ttl     opn     rt1     rt5     p50     p90                                                                                                                                               
-                              0       0       0.00    0.00    0.00    0.00                                                                                                                                              
-```
-
-You can use the exposed HTTPS endpoint to access your instance.
-
-> SECURITY NOTE: The ngrok URLS are by definition accesible from the outside internet. Even though the URLs
-> are hard to guess anyone that guesses the URL has access to your Sourcegraph instance. It is recommended to
-> kill the ngrok proxy when you are done and not let it run for long periods of time.
-
-## Method Two
-
-> A little more involved but works without outside internet.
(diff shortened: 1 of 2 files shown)


uwedeportivo:
Many thanks to @eseliger for the initial instructions on which this is based.

handbook/engineering/https_dev_instance.md
```diff
@@ -0,0 +1,173 @@
...
+> NOTE: On a Mac installation this method requires one restart of the machine to bounce the DNS resolving machinery
+> of the operating system. The `/etc/resolver` entry will otherwise be ignored until the next restart. 
+
+This method uses:
+
+* [dnsmasq](http://www.thekelleys.org.uk/dnsmasq/doc.html)
+* [mkcert](https://github.com/FiloSottile/mkcert)
+* [caddy2](https://caddyserver.com/)
+
+### Installing and setting up dnsmasq
```
keegancsmith: FYI you don't need dnsmasq to map a single address to an ip. Just edit `/etc/hosts`.

```shellsession
$ grep dev /etc/hosts
127.0.0.1 sourcegraph.example.dev
127.0.0.1 phabricator.example.dev
127.0.0.1 gitlab.example.dev
```

handbook/engineering/https_dev_instance.md
```diff
@@ -0,0 +1,173 @@
...
+
+* Generate a certificate:
+
+```shell script
+mkcert -cert-file sourcegraph.crt -key-file sourcegraph.key sourcegraph.test
+``` 
+
+The Caddy2 reverse proxy will use this certificate.
+
+### Installing and setting up Caddy2
```
felixfbecker: Afaik we are already using `nginx` in dev, which means that is already running and everyone already has it installed:

https://github.com/sourcegraph/sourcegraph/blob/c983276311fc4ef4222914bdf6a43b7ce7aa0d56/dev/Procfile#L10

https://github.com/sourcegraph/sourcegraph/blob/master/dev/nginx.conf

Would it be possible to configure nginx in the above conf file to always use HTTPS instead of requiring manual installation of another webserver?

handbook/engineering/https_dev_instance.md
```diff
@@ -0,0 +1,173 @@
+# How to run an HTTPS reverse proxy for a Sourcegraph instance on a local dev machine
+
+> NOTE: These instructions have been verified on a Mac running Catalina (the most common dev machine configuration).
+> It should fundamentally work the same on Linux with commands tailored accordingly. All the big pieces work
+> on both Linux and Mac.
+
+## Method One
```
felixfbecker: `ngrok` has really bad latency for me, so I'd suggest that "Method Two" should be the recommended approach by swapping the two

handbook/engineering/https_dev_instance.md
uwedeportivo: yeah true :-). i guess with dnsmasq you do it once and then any .test you use resolve to 127.0.0.1. i'll add the /etc/hosts as a shorter alternative, thanks

handbook/engineering/https_dev_instance.md
uwedeportivo: it's possible, i'll look into it.

handbook/engineering/https_dev_instance.md
uwedeportivo: okay, will do

felixfbecker:
Btw, a this will also make it possible to test code intelligence more easily directly from github.com, sourcegraph.com etc pointing to local servers which was previously not possible due to mixed protocols 🎉

handbook/engineering/https_dev_instance.md
uwedeportivo: ok i looked into it. that nginx serves a purpose (it better be, duh :-)). anyways it's a proxy chain like this: webpack-dev:3080 -> nginx:3081 -> sourcegraph-frontend:3082

i added another server block in the nginx.conf and rerouted the traffic. 

it works. i'm a little bit uncomfortable with it because it very much reminds me of a rube goldberg contraption.

i'll add it to the doc as an option.

handbook/engineering/https_dev_instance.md
```diff
@@ -0,0 +1,220 @@
...
+#### Option two for DNS: Using dnsmasq
+
+* `brew install dnsmasq`
+* Edit `/usr/local/etc/dnsmasq.conf` to add this entry:
+
+```text
+address=/test/127.0.0.1
+```
+
+* Add an entry to `/etc/resolver`:
```
felixfbecker: I tried following this but I get
```
tee: /etc/resolver/test: No such file or directory
```

PR #650, opened 2020-03-11
Add outline of code review cycle

ryanslade:


nicksnyder:
I have no problem documenting this, but I am curious if there is something that motivated adding this documentation?

ryanslade:
I thought that it was the responsibility of the reviewer to resolve a comment once the author had fixed it. So instead of resolving myself I was leaving a comment saying it was fixed and treating the reviewer resolving it as their way of acknowledging that they had seen the fix.

PR #669, opened 2020-03-13
Removed top 5 hospitality company logo

ryan-blunden:
I think we have enough logos now that this "Top 5 hospitality company" logo is not necessary.


ebrodymoore:
Yessss

dadlerj:
Did you intend to close without merging @ryan-blunden ?

ryan-blunden:
@dadlerj Yeah, the thinking is we keep adding to the number of logos at this stage which means keeping this one.

PR #676, opened 2020-03-17
core-services: encourage high level status updates

keegancsmith:
This is a distillation of a discussion core-services has had on slack/syncs
recently.


keegancsmith:
(proof that not every doc item I put on my TODO list is always backlogged)

keegancsmith:
If anyone has a newer suggestion of a status update they/someone else wrote I'd be happy to update that link.

PR #701, opened 2020-03-25
Update cheatsheet command for port-forward Jaeger

unknwon:
Sometimes `--namespace prod` is required.


unknwon:
Merging as-is, happy to address any post-merge comments!

PR #714, opened 2020-03-28
Blog refactor

ryan-blunden:
Mainly markup changes to rely more on Bootstrap styles. As a result, the blog index page displays thumbnails consistently and is more mobile-friendly.

### DIFF
--- a/website/src/pages/go.tsx
+++ b/website/src/pages/go.tsx
@@ -4,23 +4,23 @@ import * as React from 'react'
 import BlogPosts from '../components/BlogPosts'
 import Layout from '../components/Layout'
 export default class GoList extends React.Component<any, any> {
-    public componentDidMount(): void {
-        const script = document.createElement('script')
-        script.src = '//js.hsforms.net/forms/v2.js'
-        const hubspot = document.getElementById('gopherConForm')
-        hubspot.appendChild(script)
-        script.addEventListener('load', () => {
-            ;(window as any).hbspt.forms.create({
-                portalId: '2762526',
-                formId: '9cfe9111-fe45-45f5-873a-9f4925affdfd',
-                target: '#gopherConForm',
-            })
-        })
-    }
+    // public componentDidMount(): void {
+    //     const script = document.createElement('script')
+    //     script.src = '//js.hsforms.net/forms/v2.js'
+    //     const hubspot = document.getElementById('gopherConForm')
+    //     hubspot.appendChild(script)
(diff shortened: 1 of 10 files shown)


website/src/pages/go.tsx
```diff
@@ -4,23 +4,23 @@ import * as React from 'react'
...
+    //     const hubspot = document.getElementById('gopherConForm')
+    //     hubspot.appendChild(script)
+    //     script.addEventListener('load', () => {
+    //         ;(window as any).hbspt.forms.create({
+    //             portalId: '2762526',
+    //             formId: '9cfe9111-fe45-45f5-873a-9f4925affdfd',
+    //             target: '#gopherConForm',
+    //         })
+    //     })
+    // }
```
rvantonder: Is there a reason this is commented out but not deleted?

website/src/pages/strange-loop.tsx
```diff
@@ -4,23 +4,23 @@ import * as React from 'react'
...
+    //     const hubspot = document.getElementById('strangeLoopForm')
+    //     hubspot.appendChild(script)
+    //     script.addEventListener('load', () => {
+    //         ;(window as any).hbspt.forms.create({
+    //             portalId: '2762526',
+    //             formId: 'b4b2b911-0515-4a76-bfc0-2f4ca384bdbc',
+    //             target: '#strangeLoopForm',
+    //         })
+    //     })
+    // }
```
rvantonder: Flagging as commented out.

rvantonder:
I remember some overflow issues with blog posts on mobile and it's possible to reproduce. Don't know if this PR is meant to solve it, so just flagging since you asked me to have a look. 

Example, see release blog post 3.4:

Screen Shot 2020-03-30 at 9 22 59 PM

website/src/pages/go.tsx
ryan-blunden: Because it will be re-used once (if) we host the liveblog for GopherCon again.

ryan-blunden: I'll fix.

website/src/pages/strange-loop.tsx
ryan-blunden: Since removed.

ryan-blunden:
> I remember some overflow issues with blog posts on mobile and it's possible to reproduce.

Thanks for this, I couldn't remember which post it was that caused the issue.  To fix that post would require content changes so I'm ok with this PR not addressing that.

PR #727, opened 2020-03-31
Update code intel team's 2020 Q1 OKR progress

efritz:


company/okrs/2020_q1.md
```diff
@@ -57,12 +57,16 @@
 1. **LSIF is adopted by customers**
    1. At least three customers have > 50% of code intel queries powered by LSIF (excluding fallback references)
```
nicksnyder: What is the status on these first two KRs?

creachadair: As far as I know we have not hit either (1) or (2) here.

nicksnyder: We shouldn't remove KRs, but I do want there to be an update here. The update can be "no progress, we are pursuing this by making LSIF work on open source (LINK TO RFC)" or whatever is accurate

creachadair: I added forward references to the notes I added below. Would you like to see more detail than this?

PR #732, opened 2020-04-01
decisions sheet for recording and revisiting decisions

sqs:
When making decisions at Sourcegraph, it's important to:

1. Gather input from people with relevant knowledge.
1. Decide quickly and precisely.
1. Communicate the decision and the rationale clearly.
1. Review the decision with the benefit of hindsight to understand what we got right or wrong.

These things become more important as we grow, so we're trying to be more deliberate about decision-making. To do so, we're trying out recording these decisions in [**Decisions (Google Sheet)**](https://docs.google.com/spreadsheets/d/1CVdqgSYyukW_eMlfKUDSRZpT9xzYHRDMH5NYbZRVLzQ/edit).

### DIFF
--- a/handbook/communication/decisions.md
+++ b/handbook/communication/decisions.md
@@ -1,0 +1,34 @@
+# Making decisions
+
+When making decisions at Sourcegraph, it's important to:
+
+1. Gather input from people with relevant knowledge.
+1. Decide quickly and precisely.
+1. Communicate the decision and the rationale clearly.
+1. Review the decision with the benefit of hindsight to understand what we got right or wrong.
+
+These things become more important as we grow, so we're trying to be more deliberate about decision-making.
+
+## Decisions sheet
+
+> This is optional and experimental. We'll be iterating on this. Try it and post your feedback in #handbook on Slack!
+
+[**Decisions (Google Sheet)**](https://docs.google.com/spreadsheets/d/1CVdqgSYyukW_eMlfKUDSRZpT9xzYHRDMH5NYbZRVLzQ/edit)
+
+1. This sheet for decisions that are:
+   - high risk (difficult or costly to unwind, or with big/unknown long-term implications); and
+   - not able to be reached by consensus within 5 minutes.
+1. When a decision needs to be made, add a row.
+   - Even if the decision hasn't been made yet (but needs to be made soon), it's still useful to add a row with as many cells filled out as possible.
+   - For information on each column, see the column header notes. We will document these in the handbook once we find a set of descriptions that work well.
+1. The recommender is responsible for adding a `Revisit decision: ...` item to the relevant weekly meeting agenda for the decision when its revisit date is reached.
+
+## Related reading
+
+- [How Coinbase makes decisions](https://medium.com/@barmstrong/how-we-make-decisions-at-coinbase-cd6c630322e9)
+- [Bain's RAPID decision-making tool](https://www.bain.com/insights/rapid-tool-to-clarify-decision-accountability/)
+  - R: Recommend (the person who describes a problem and recommends a decision)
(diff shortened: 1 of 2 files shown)


handbook/communication/decisions.md
```diff
@@ -0,0 +1,34 @@
+# Making decisions
+
+When making decisions at Sourcegraph, it's important to:
+
+1. Gather input from people with relevant knowledge.
+1. Decide quickly and precisely.
+1. Communicate the decision and the rationale clearly.
+1. Review the decision with the benefit of hindsight to understand what we got right or wrong.
```
dadlerj, line 8:
```suggestion
1. Review the decision with the benefit of hindsight to understand what we got right or wrong.
1. Ensure clear owners are responsible for following through on the decision with the appropriate context on prioritization and timing.
```

handbook/communication/decisions.md
```diff
@@ -0,0 +1,34 @@
...
+
+These things become more important as we grow, so we're trying to be more deliberate about decision-making.
+
+## Decisions sheet
+
+> This is optional and experimental. We'll be iterating on this. Try it and post your feedback in #handbook on Slack!
+
+[**Decisions (Google Sheet)**](https://docs.google.com/spreadsheets/d/1CVdqgSYyukW_eMlfKUDSRZpT9xzYHRDMH5NYbZRVLzQ/edit)
+
+1. This sheet for decisions that are:
```
dadlerj, line 18:
```suggestion
1. This sheet is for decisions that are:
```

handbook/communication/decisions.md
```diff
@@ -0,0 +1,34 @@
...
+   - high risk (difficult or costly to unwind, or with big/unknown long-term implications); and
+   - not able to be reached by consensus within 5 minutes.
+1. When a decision needs to be made, add a row.
+   - Even if the decision hasn't been made yet (but needs to be made soon), it's still useful to add a row with as many cells filled out as possible.
+   - For information on each column, see the column header notes. We will document these in the handbook once we find a set of descriptions that work well.
+1. The recommender is responsible for adding a `Revisit decision: ...` item to the relevant weekly meeting agenda for the decision when its revisit date is reached.
+
```
dadlerj, line 25: Any thoughts on this convention?

Note I added some conditional formatting if the row has "Yes" marked in that column.

```suggestion
1. When a decision is implemented (i.e., the `Implemented` column is marked as `Yes`), hide the row in the Google sheet.
```

christinaforney: To start with and as we're experimenting I'd prefer we don't hide them since the point is to surface these decisions more. In the future, could we move the implemented items to a new sheet tab named "Implemented" so that they aren't being hidden, but are still moved out of the way so active decisions are top of focus?

dadlerj: Sgtm

juliasourceress:
I think it's also important to note that we want to be more deliberate about decision making *and* about tracking which decisions come up and still need to be made, perhaps in the future, or perhaps after an additional conversation if one such conversation wasn't enough - but this way important decisions do not fall through the cracks for lack of being captured and attended to.

sqs:
@juliasourceress sounds good, can you propose some wording for that?

handbook/communication/decisions.md
```diff
@@ -0,0 +1,35 @@
+# Making decisions
+
+When making decisions at Sourcegraph, it's important to:
+
+1. Gather input from people with relevant knowledge.
```
sqs, line 5:
```suggestion
1. Know which pending decisions need to be made soon.
1. Gather input for decisions from people with relevant knowledge.
```

@juliasourceress look good to address your point?

PR #742, opened 2020-04-02
Let's finally add newsletter subscription to footer.

aileenrose:
Screen Shot 2020-04-02 at 12 34 03 PM

Our previous attempt caused conflicts on pages with other embedded forms i.e. /contact/sales, /contact/request-demo.

### DIFF
--- a/website/src/components/Subscribe.tsx
+++ b/website/src/components/Subscribe.tsx
@@ -1,0 +1,28 @@
+import React from 'react'
+
+class HubspotSubscribe extends React.Component {
+    public componentDidMount(): void {
+        const script = document.createElement('script')
+        script.src = '//js.hsforms.net/forms/v2.js'
+        document.body.appendChild(script)
+
+        script.addEventListener('load', () => {
+            ;(window as any).hbspt.forms.create({
+                portalId: '2762526',
+                formId: '9e9cd4d5-7a88-47e0-97e2-93dd11348ae5',
+                cssClass: 'subscribe-form',
+                target: '#hubspotSubscribeForm',
+            })
+        })
+    }
+    public render(): JSX.Element {
+        return (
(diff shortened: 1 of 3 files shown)


website/src/css/components/_Footer.scss
```diff
@@ -39,4 +43,84 @@
...
     }
+
+    &__newsletter-form-heading {
+        opacity: 0.4;
+        text-transform: uppercase;
+        font-size: 0.9rem;
+        font-weight: bolder;
+    }
+}
+.subscribe-form {
```
ryan-blunden, line 54: Are these styles meant to be at the global level or should they be scoped to the footer?

website/src/components/Subscribe.tsx
```diff
@@ -0,0 +1,28 @@
...
+                portalId: '2762526',
+                formId: '9e9cd4d5-7a88-47e0-97e2-93dd11348ae5',
+                cssClass: 'subscribe-form',
+                target: '#hubspotSubscribeForm',
+            })
+        })
+    }
+    public render(): JSX.Element {
+        return (
+            <div id="newsletter-signup-footer" className="newsletter-form-footer row">
```
ryan-blunden, line 20: Does this id need to exist?

aileenrose: it would be good to keep for targeting in A/B tests.

website/src/css/components/_Footer.scss
aileenrose: I'm considering having it available for blog pages, but not yet sure what the style will be or look like overall.

ryan-blunden: If that's the case, then I'd move it into its own stylesheet.

PR #760, opened 2020-04-10
Remove solutions pages and redirect

aileenrose:
https://sourcegraph.slack.com/archives/CNC4YAL1E/p1586486359225100
this is to noindex /solutions pages that are still in draft 
also excluded 404 pages from sitemap.

Note:  gatsby-plugin-advanced-sitemap does not acknowledge Helmet configuration which is why I needed to also exclude from gatsby config.


aileenrose:
@sqs noted that these pages are no longer required.  Remove and added redirects.

PR #762, opened 2020-04-12
Update with sales interview process expectations.

juliasourceress:


handbook/sales/interviews/index.md
```diff
@@ -1,6 +1,15 @@
-# Sales interviews
+# Sales interview process
 
-- [Mock customer call](mock_customer_call.md) interview module
+- You apply.
```
sqs, line 3: Use a numbered list instead of a bullet list (i.e., make each line start with `1. ` instead of `- `. You can use `1.` for each line, you don't need `2. `, `3. `, etc. (it will count up automatically for you).

PR #763, opened 2020-04-12
Update mock call to be a mock pitch vs any other potential stage.

juliasourceress:


sqs:
When editing your own handbook section, no need to wait for reviews unless you want to. But looks good to me!

PR #765, opened 2020-04-13
use "Sentence case" in design section

sqs:
See https://about.sourcegraph.com/handbook/communication/style_guide#general.


sqs:
FYI @imtommyroberts

PR #766, opened 2020-04-13
use "Sentence case" in customer data policy

sqs:
See https://about.sourcegraph.com/handbook/communication/style_guide#general.


sqs:
FYI @ebrodymoore

PR #772, opened 2020-04-14
blog: add link to LSIF definition

creachadair:
I accidentally omitted this during the initial post.


creachadair:
Merging around the LSIF build failure, which is unrelated to this change.

PR #775, opened 2020-04-15
Adding engineering prototyping to product docs

christinaforney:
We want to encourage developers to take initiative on projects they are passionate about. This section shows the team how they can prioritize working on these ideas part-time, that we trust them to do this well, and how we'll be evaluating their work.

### DIFF
--- a/handbook/product/prioritizing.md
+++ b/handbook/product/prioritizing.md
@@ -6,7 +6,7 @@ We receive tons of feature requests and bug reports, more than we can handle. Th
 
 ## How do I get something prioritized by our product team?
 
-- [Create a GitHub issue](https://github.com/sourcegraph/sourcegraph/issues/new/choose).
+- [Create a GitHub issue](https://github.com/sourcegraph/sourcegraph/issues/new/choose) OR [write an RFC](../communication/rfcs/index.md).
   - Include as much detail as possible about the issue.
   - Provide context around urgency and priority.
   - Link to the customer (if applicable) in HubSpot (see [below](#customer-requests)).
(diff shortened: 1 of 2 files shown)


handbook/product/prioritizing.md
```diff
@@ -16,6 +16,10 @@ We receive tons of feature requests and bug reports, more than we can handle. Th
   - Send to the Product Manager to prioritize with the team.
   - Add the item in suggestion mode to [the roadmap](https://docs.google.com/document/d/1cBsE9801DcBF9chZyMnxRdolqM_1c2pPyGQz15QAvYI/edit#) in your best guess at relative priority. The team or PM will accept or update the priority of the item.
 
+### Engineering prototyping
+
+Sometimes there is a general product area that the CEO/CTO/Product Manager/CE/Sales seem interested in, and a developer is passionate/inspired/excited about it and wants to work on prototyping in that area. The developer has total freedom for a couple months to work on it part-time as long as they make progress in prototyping with the Sales and CE teams and are showing progress to customers regularly. The sole test will be, if we ask the Sales/CE teams "is $DEVELOPER's prototype/project drawing interest in customer calls and is $DEVELOPER iterating on it well in response to feedback?"
```
nicksnyder, line 21: This paragraph doesn't adequately describe how this interacts with our planning process or our planned work.

Do developers need to communicate that they are doing this work to anyone? Who needs to approve of this work?
Is this work reflected in the tracking issue?

Does "part time" mean "part of your normal working hours" or "as an extracurricular activity"?

If the former, it should be planned and accounted for during our normal planning process I think (and there doesn't appear to be a need to have this paragraph at all other than to say, "if you have an idea, then please speak up during planning!".)

If the latter, it also seems unnecessary to document this since engineers can do as they please in their free time.

PR #777, opened 2020-04-16
Fix rendering of table

asdine:
This fixes the rendering of a table in the "Commit message guidelines" page.

Before:
Screenshot 2020-04-16 at 12 11 03

After:
Screenshot 2020-04-16 at 12 11 16


nicksnyder:
This table should be nestled under 4, so I think the right fix the formatting is to indent more, not less?

asdine:
Makes sense, I fixed it 👍

PR #778, opened 2020-04-16
Add support for Node 13.x

asdine:
I tried installing the website dependencies using Node 13 but it didn't work. 
The problem came from the Sharp version used, which doesn't support Node 13.x.
I have fixed that by using the latest release of Sharp and updating the dependencies, as explained in this [issue](https://github.com/lovell/sharp/issues/2005).

This fix works on both Node v13.13.0 and v12.16.2.


felixfbecker:
Seems like there are more dependency upgrades needed:
```
10:37:19 AM: error @typescript-eslint/eslint-plugin@2.28.0: The engine "node" is incompatible with this module. Expected version "^8.10.0 || ^10.13.0 || >=11.10.1". Got "11.4.0"
```

Side note: I think it would be good to have Renovate upgrade dependencies in this repo automatically cc @ryan-blunden

asdine:
Fixed, I forgot to update the .nvmrc file, it should be fine now

website/.nvmrc
```diff
@@ -1 +1 @@
-v11.4.0
+v12.16.2
```
felixfbecker, line 1: Is there a reason we shouldn't use 13?

asdine: I figured it would be better to target the latest LTS version to make sure we never run into problems with our dependencies. However, I'm not that versed into the Node world so maybe it's an unnecessary precaution.

asdine: As discussed orally, it's safe to use 13.13.0 directly, I made the change 👍

attfarhan:
Thanks @asdine and @felixfbecker. Agree with getting renovate set up, definitely needed in this repo.

PR #780, opened 2020-04-16
style guide: document how to refer to features

sqs:


sqs:
@adamfrankl2015 There are some exceptions to what I wrote:

- Code Intelligence
- Universal Code Search

I think these things are not really "features" per se. They are something else, like "concepts". Still, why are they capitalized (when our general style is not to capitalize things)? Should they always be capitalized? Personally, I find it strange to capitalize them in most places, but it does feel "right" to capitalize them in sales decks.

adamfrankl2015:
> @adamfrankl2015 There are some exceptions to what I wrote:
> 
> * Code Intelligence
> * Universal Code Search
> 
> I think these things are not really "features" per se. They are something else, like "concepts". Still, why are they capitalized (when our general style is not to capitalize things)? Should they always be capitalized? Personally, I find it strange to capitalize them in most places, but it does feel "right" to capitalize them in sales decks.

They are in fact named ideas, which we should capitalize as they are proper nouns. Using Agile methods does not mean the same thing as merely being agile. It is a named idea. Other examples include Big Data and by extension Big Code.

PR #782, opened 2020-04-16
update owners of website, handbook, etc.

sqs:


.github/CODEOWNERS
```diff
@@ -5,6 +5,14 @@
 # Order is important; the last matching pattern takes the most
 # precedence
 
+/handbook @sqs
```
nicksnyder, line 8: we do this in sourcegraph/sourcegraph

```suggestion
# The "nobody" team is an alerting mechanism that indicates when there are files in a PR that are
# not owned by anyone. This is a signal to the PR author that certain modified files are unowned and
# they need to (1) track down the owners using git blame and (2) update CODEOWNERS.
* @sourcegraph/nobody

/handbook @sqs
```

.github/CODEOWNERS
```diff
@@ -5,6 +5,14 @@
...
+/company @sqs
+/workflow @christinaforney
+/direction @christinaforney
+/blogposts @aileenrose
+/website @aileenrose
+/docs @aileenrose
+
```
christinaforney, line 15:
```suggestion

/handbook/product @christinaforney
/handbook/design @imtommyroberts 
```

PR #784, opened 2020-04-16
handbook: update the code intel team index

creachadair:
- Add a "Direction" section to the code intel index, to give a higher-level view of the team's purpose.
- Add a contacts & planning section, to give pointers to team documentation.
- Clarify the code intel tech stack (chiefly: Removes mention of LSP).

Please pay particular attention to the wording of the "Direction" section. I believe this is consistent with our previous discussions, but since we haven't previously written these points in the Handbook, I'd like to be sure we are in agreement.


efritz:
I'm glad you added the direction section, and I have no issues linking to other documents here. Looks like a great update!

PR #788, opened 2020-04-20
RFC 137: Document testing philosophy and owners

nicksnyder:
[Action item #5 in RFC 136](https://docs.google.com/document/d/14f7lwfToeT6t_vxnGsCuXqf3QcB5GRZ2Zoy6kYqBAIQ/edit#heading=h.twdj57egor10)


handbook/engineering/testing.md
```diff
@@ -2,11 +2,87 @@
...
 This file documents how we test code at Sourcegraph.
 
+## Philosophy
 
-## Percy
+We rely on automated testing to ensure the quality of our product.
+
+Any addition or change to our codebase should be covered by an appropriate amount of automated tests to ensure that:
+
+1. Our product works the way we intend it to when we ship it to customers.
```
lguychard, line 11:
```suggestion
1. Our product works as intended when we ship it to customers.
```

lguychard (approved):

tsenart (approved):

efritz (approved):

eseliger (approved):

mrnugget (approved):

attfarhan (approved):

keegancsmith (approved):

beyang (approved):

handbook/engineering/testing.md
```diff
@@ -2,11 +2,87 @@
...
+
+Benefits:
+
+- They are usually very fast to execute because slow operations can be mocked.
+- They are the easiest tests to write, debug, and maintain because the code under test is small.
+- They only need to run on changes that touch code which could make the test fail, which makes CI faster and minimizes the impact of any flakiness.
+
+Tradeoffs:
+
+- They don't verify our systems are wired up correctly end-to-end.
```
felixfbecker, line 37: I think this trade-off should be called out - I frequently meet this in the webapp. This is better or worse depending on how well designed the implementation and test is, but unit tests are also generally more vulnerable to this because they test the smallest unit of test, and any change that removes the unit, splits the unit, merges two units, changes the return type of the unit, etc etc will require updates to unit tests (while integration tests and e2e tests are more likely to remain unchanged). This can result in _a lot_ of work per change and can become a change preventer.

```suggestion
- They don't verify our systems are wired up correctly end-to-end.
- They often require changes when refactoring code.
```

handbook/engineering/testing.md
```diff
@@ -2,11 +2,87 @@
...
+- They are not as easy to write as unit tests.
+
+Examples:
+
+- Tests that call our search API to test the behavior of our entire search system.
+- Tests that validate the behavior of complex UI flows in the browser while mocking out all network requests so no backend is required.
```
felixfbecker, line 56: _complex_ UI flows will probably fall into the category that are _hard_ to test with integration tests, because a complex UI flow usually means there is state change on the server involved, which is difficult to replicate on the client without reimplementing the logic of the server inside tests (a mini in-memory backend), which then becomes time-consuming to maintain with every backend update. I'd actually say "complex" UI flows mark the verge to where e2e tests are more suitable. We hopefully will find some solution for this in the webapp, but this may not be the strongest example for where integration tests shine.

Possible alternative:

```suggestion
- Tests that validate UI behavior in the browser while mocking out all network requests so no backend is required.
```

handbook/engineering/testing.md
```diff
@@ -2,11 +2,87 @@
...
+- Run our Sourcegraph Docker image and verify that users can sign in and perform a search.
+
+## Visual testing
 
 We use [Percy](https://percy.io/) to detect visual changes in Sourcegraph features. You may need permissions to update screenshots if your feature introduces visual changes. Post a message in #dev-chat that you need access to Percy, and someone will add you to our organization (you will also receive an invitation via e-mail). Make sure you are logged in with your GitHub credentials.
 
+## Ownership
+
+- [Distribution team](distribution/index.md) owns build and test infrastructure.
+- [Web team](web/index.md) owns any tests that are driven through the browser.
```
felixfbecker, line 84: 👍

handbook/engineering/web/index.md
```diff
@@ -6,6 +6,11 @@
 
 Bring the power of Sourcegraph (e.g. search, code intelligence, campaigns) to all developers everywhere that they interact with code (e.g. command line, editors, web browser, code hosts).
 
+## Contact
+
+- #web channel or @web-team in Slack.
+- [team/web](https://github.com/sourcegraph/sourcegraph/issues/new?labels=team/web) label in GitHub.
```
felixfbecker, line 12:
```suggestion
- [team/web](https://github.com/sourcegraph/sourcegraph/issues/new?labels=team/web) label and [@sourcegraph/web](https://github.com/orgs/sourcegraph/teams/web) team on GitHub.
```

felixfbecker (commented):
I like this a lot!

What do you think of adding an image of the testing pyramid somewhere, to communicate the ratios between amount of tests we're aiming for?

handbook/engineering/testing.md
```diff
@@ -2,11 +2,87 @@
 
 This file documents how we test code at Sourcegraph.
 
+## Philosophy
```
rvantonder, line 5: I'm not confident on where to fit this in, but it's worth pointing out why tests are important beyond quality of our product (which feels just a tad overly customer-centric), and what level of quality assurance it gives us. Feel free to incorporate as much or as little as the following bits that resonate with you:

> "Program testing can be used to show the presence of bugs, but never to show their absence!" - Dijkstra

- Ergo, testing allows us to find bugs, that's why they're so useful, and that's why we want them!
- A massive benefit of a well-tested code base is that it provides confidence for refactors and identifies gaps if a refactor breaks things. Refactoring and tests go hand-in-hand.
- Testing is a form of software development "self-care": we don't just write tests to feel good about shipping a feature _now_. We also write tests for our future ourselves and other team members because tests are a knowledge base for ourselves, a coarse and growing specification of what the code should do.

handbook/engineering/testing.md
rvantonder: > any change that removes the unit, splits the unit, merges two units, changes the return type of the unit, etc etc will require updates to unit tests... This can result in a lot of work per change and can become a change preventer.

My feeling is that if the refactoring changes described here impact tests, then changing the associated unit tests is also a benefit. I'm not saying that's not a lot of work (there is a tradeoff), but this behavior is also beneficial in that if the changes result in a lot of work, that only speaks to the significance of the change in structure in question that we should take care to keep testing.

handbook/engineering/testing.md
nicksnyder: > Ergo, testing allows us to find bugs, that's why they're so useful, and that's why we want them!

I think this is covered by the first documented point "Our product works as intended when we ship it to customers." Not working as intended is the definition of a bug :)

> Testing is a form of software development "self-care": we don't just write tests to feel good about shipping a feature now. We also write tests for our future selves and other team members because tests are a knowledge base for ourselves, a coarse and growing specification of what the code should do.

I think this is covered by the second documented point "Our product continues to work and doesn't accidentally break as we make changes to our product over time."

> A massive benefit of a well-tested code base is that it provides confidence for refactors and identifies gaps if a refactor breaks things. Refactoring and tests go hand-in-hand.

I think this is also covered by the second point, but I did add an extra sentence to this effect in d2a90d97bdb7faa26285d779612c014de61e83f4. I also changed "our product" to "our product and code" so it would sound less product centric (although our code only exists to serve the product).

unknwon (approved):
👍

handbook/engineering/testing.md
nicksnyder: I share Rijnard's perspective here: I don't think this is a tradeoff of unit tests (as compared to integration or e2e tests). You don't experience this with e2e tests because it simply isn't possible or feasible to test the entire state space of what you want/should test. If you did/could, then e2e tests would have the same problem.

To the extent that state space of the code changes, then yes, tests need to be updated to cover that new state space (because testing is valuable), and that is a good thing (not a bad thing).

Writing good tests is extra work, and the point of adding this to our handbook is to document that engineers should do this work. The extra work bit is not a tradeoff of unit tests in particular, it is a feature of us prioritizing testing at all and unit tests being the most effective way to test most of the state space.

nicksnyder:
> What do you think of adding an image of the testing pyramid somewhere, to communicate the ratios between amount of tests we're aiming for?

Added one in 89b85a122a6a3abaf87ffdb2eba968d9038f0bef

PR #791, opened 2020-04-20
Add "conciously invest in tooling" as Core Services principle

mrnugget:
This is a follow-up to our [Core Services retrospective last week](https://docs.google.com/document/d/1Bk0c6vf5g5yqjk-xoBIw3xVb_Rs44YX4aiZwOLkS5X0/edit?ts=5e99ea72).

Since we didn't have a place for this already, I've added the
"Principles" header.

The problem is that this is now, of course, a non-exhaustive list and
might make the wrong impression.

What do you think?


ryanslade:
I think since it is under "Principles" it's ok to be fairly general.

PR #796, opened 2020-04-21
Adding Extension/UI improvements to 3.15 blogpost

christinaforney:
Felix noticed I missed a section on UI improvements on 3.15, so we should add that in!

@felixfbecker - do you have an image or video we can add?


felixfbecker:
Maybe this?

PR #797, opened 2020-04-21
release publishing: cherry pick version substitution into release branch

uwedeportivo:
we have docs @ 3.15 or docs @3.14 etc

they also need the same version substitution, otherwise they have the versions from master when the branch was cut (so the previous version)

related to https://github.com/sourcegraph/sourcegraph/pull/10065


uwedeportivo:
i did a `docsite check` locally and it's clean. i did a `yarn build` in the website subdir and that went well too.

so with apologies to the checks, i'm just going to land this.

PR #826, opened 2020-04-27
Fix mobile view and remove old events

aileenrose:
Screen Shot 2020-04-27 at 3 19 53 PM

### DIFF
--- a/website/src/css/pages/_events.scss
+++ b/website/src/css/pages/_events.scss
@@ -1,5 +1,5 @@
 .event {
     &__item h6 {
-        color: rgb(105, 120, 130);
+        color: $light-2;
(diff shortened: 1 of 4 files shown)


website/src/css/pages/_events.scss
```diff
@@ -1,5 +1,5 @@
-.events {
+.event {
     &__item h6 {
-	color: #ccc;
+        color: rgb(105, 120, 130);
```
attfarhan, line 3:
```suggestion
        color: $light-2;
```

You can get a similar color using $light-2 from our color palette, so we have consistent colors throughout our site (see _colors.scss). Does this color work?

PR #851, opened 2020-05-04
blogpost: A guide to LSIF

efritz:
This closes the long-standing need as described in https://github.com/sourcegraph/sourcegraph/issues/9719.


nicksnyder:
Thanks for writing this up! Should this be evergreen documentation that we keep updated instead of a blog post?

efritz:
Possibly should be. Where do you think this should live if so?

nicksnyder:
The evergreen documentation would probably be product docs (https://docs.sourcegraph.com/).

Once it is there, I think it would still be valuable to publish a blog post to say something like "this is now documented and ready for people to try this for a few languages, we have it working on sourcegraph.com for repos X, Y, Z, ..."

PR #852, opened 2020-05-04
Move TypeScript guide into handbook

felixfbecker:
This applies to all repos, not just the main repo, so it belongs to the handbook.


felixfbecker:
Yep, good idea.

sqs:
@felixfbecker For the build failure, you need to add an index.md in the new subdirectory you created. The broken link is in the generated HTML in the breadcrumb.

felixfbecker:
Thanks! Maybe that error message could be improved 😅

PR #855, opened 2020-05-05
handbook: update physical worksites info in all-remote section 

mercadon:
As discussed with @sqs, this change elaborates that our remote-all policy is inclusive of making explicit provision for physical worksites as required by some employment arrangements.


rvantonder:
Thanks @mercadon!

PR #910, opened 2020-05-07
Create hiring-flow.md

mercadon:
Adding link to shared folder that will hold PDFs of each team's pipeline process.


sqs:
To make this page discoverable, you need to link to it from some other page. Otherwise it’ll only be accessible via search. Chat me on Slack if I can help.

PR #913, opened 2020-05-07
Add hints to engineering onboarding

daxmc99:
Some slight hints to make the engineering onboarding a bit more clear and provide some guidance on using Postgres.

PR #918, opened 2020-05-09
docs: removes extra backquote

unknwon:
Accidentally added in https://github.com/sourcegraph/about/pull/913.


sqs:
(For obvious/small fixes like this, you can just commit to master if you want.)

PR #930, opened 2020-05-13
Direct communication channels

christinaforney:
How do we communicate when Slack goes down? Slack is having an outage and I feel completely blind with how to get in touch with anyone on the team, start new conversations, or add any information that would be read asynchronously. I think it would be wise for us to have a documented plan to guide the team if/when these things happen. Do we have any email lists set up? Should we just start email threads to team@?

### DIFF
--- a/handbook/communication/index.md
+++ b/handbook/communication/index.md
@@ -2,6 +2,10 @@
 
 We're an [all-remote](../../company/remote/index.md) company, with teammates from all around the world and (soon) no primary office. To make this work, we need to be deliberate about how we communicate.
 
+## Direct communication
+
+We rely on [Slack](team_chat.md) for all our internal communications and conversations so discussion can happen in public channels. For the most part we only use email for external communications. In the rare occasion of a Slack outage, we communicate by....???
+
 ## Sources of truth
 
 These places are the source of truth for information at Sourcegraph. Information in these places is expected to be accurate and up-to-date:


slimsag:
I mean, it's not just our company. I also can't talk with customers right now.

handbook/communication/index.md
```diff
@@ -2,6 +2,10 @@
 
 We're an [all-remote](../../company/remote/index.md) company, with teammates from all around the world and (soon) no primary office. To make this work, we need to be deliberate about how we communicate.
 
+## Direct communication
+
+We rely on [Slack](team_chat.md) for all our internal communications and conversations so discussion can happen in public channels. For the most part we only use email for external communications. In the rare occasion of a Slack outage, we communicate by....???
```
slimsag, line 7:
```suggestion
We rely on [Slack](team_chat.md) for all our internal communications and conversations so discussion can happen in public channels. For the most part we only use email for external communications. In the rare occasion of a Slack outage, we communicate [via other means.](https://www.smithsonianmag.com/innovation/scientists-prove-that-telepathic-communication-is-within-reach-180952868/)
```

slimsag:
All: See your @sourcegraph.com for a discord link in the meantime. Not saying we should/shouldn't use this in the future.

nicksnyder:
I am not sure we need to merge a PR like this. Slack might be down for hours but it probably won't be down for days. We can and should be able to survive not having Slack down for hours without doing anything special.

I think a better set of PRs to merge would be PRs that add mailing lists for each team.

unknwon:
2cents: GitHub has team discussions and basically reuses the whole notification system of GitHub, and often resulted in receiving emails. Though not everyone would have configured to receive emails from GitHub, but if email isn't the reliable option, so the mailing list.

PR #934, opened 2020-05-13
engineering: describe how to merge deploy-sourcegraph into dot-com

uwedeportivo:
move docs from the dot-com and dogfood repos into handbook


nicksnyder:
We already have a page on deployments. Does this info belong there? https://about.sourcegraph.com/handbook/engineering/deployments

uwedeportivo:
ah, nice

i'll move it as a section in there

PR #965, opened 2020-05-20
Credit 3.16 external contributors

rvantonder:
I'm not sure where we pull the list from, but we had PRs from external contributors that didn't make it in here so just adding them :-)


christinaforney:
Good catch @rvantonder - thanks for adding. I run a script to get the list of contributors: https://sourcegraph.com/github.com/sourcegraph/about/-/blob/bin/generate_contributors.go not sure where things went wrong in this case.

ryan-blunden:
@christinaforney That script only brings in users that have created issues, not submitted pull requests. Created https://github.com/sourcegraph/about/issues/967 and I can fix by next release.

PR #976, opened 2020-05-22
Update structural search blog post

rvantonder:
- One of the search links doesn't work because a repo was moved to a different org
- Removing the 'future improvements +1 this on GH section' which was more meant as an immediate way to get feedback when the blog was posted and this won't age well.


rvantonder:
Will merge without approval, just putting up PR for visibility.

PR #980, opened 2020-05-27
handbook: add Go test naming convention

rvantonder:
See https://github.com/sourcegraph/sourcegraph/pull/10967#issuecomment-633835528

### DIFF
--- a/handbook/engineering/testing.md
+++ b/handbook/engineering/testing.md
@@ -92,7 +92,7 @@ We use [Percy](https://percy.io/) to detect visual changes in Sourcegraph featur
 ## Conventions
 
 - **Naming tests in Go code.** We strive to follow these conventions when naming Go test functions:
-  - General rule: avoid using `_` immediately after `Test`. Instead, if testing a function like `fooBar`, the test is called TestFooBar and not `Test_fooBar`. 
+  - General rule: avoid using `_` immediately after `Test`. Instead, if testing a function like `fooBar`, the test is called `TestFooBar` and not `Test_fooBar`. 
   - Using underscores to separate function names and types, [for example](https://sourcegraph.com/search?q=repo:%5Egithub%5C.com/golang/go%24%40go1.14.3+func+Test.*_&patternType=regexp&case=yes), is fine. 
 
 ## Other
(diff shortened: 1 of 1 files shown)


handbook/engineering/testing.md
```diff
@@ -89,6 +89,12 @@ We use [Percy](https://percy.io/) to detect visual changes in Sourcegraph featur
 - [Distribution team](distribution/index.md) owns build and test infrastructure.
 - [Web team](web/index.md) owns any tests that are driven through the browser.
 
+## Conventions
+
+- **Naming tests in Go code.** We strive to for the following conventions when naming Go test functions:
```
rvantonder, line 94:
```suggestion
- **Naming tests in Go code.** We strive to follow these conventions when naming Go test functions:
```

handbook/engineering/testing.md
```diff
@@ -89,6 +89,12 @@ We use [Percy](https://percy.io/) to detect visual changes in Sourcegraph featur
 - [Distribution team](distribution/index.md) owns build and test infrastructure.
 - [Web team](web/index.md) owns any tests that are driven through the browser.
 
+## Conventions
+
+- **Naming tests in Go code.** We strive to follow these conventions when naming Go test functions:
+  - General rule: avoid using `_` immediately after `Test`. Instead, if testing a function like `fooBar`, the test is called TestFooBar and not `Test_fooBar`.
```
bobheadxi, line 95:
```suggestion
  - General rule: avoid using `_` immediately after `Test`. Instead, if testing a function like `fooBar`, the test is called `TestFooBar` and not `Test_fooBar`. 
```

nicksnyder:
Is this documented convention different than what is documented in the standard library? If not, can we link to the standard documentation instead of (or in addition to) trying to explain it ourselves?

https://golang.org/pkg/testing/#hdr-Examples

rvantonder:
Oh that's interesting @nicksnyder, had no idea. The one comment I'd make is that the linked section refers to example functions rather than tests, but other than that it's seems to be exactly what we're going for. So something like "We use the same convention for naming test functions as example functions as described in the Go testing library [ink]"

PR #1036, opened 2020-06-05
Add web testing guide

felixfbecker:
This should serve as the single source of truth going forward for the testing principles we strive for on the web team that are currently not written down anywhere (and not possible to enforce with ESLint).

It was a fair bit of work to reflect on all this but I hope it's worth it as we grow our team ❤️

### DIFF
--- a/handbook/engineering/web/testing.md
+++ b/handbook/engineering/web/testing.md
@@ -8,12 +8,12 @@ This is the implementation of our [company testing philosophy](../testing.md) fo
 
 2. **Our product and code doesn't accidentally break as we make changes over time.**
 
-3. **Failures should be clear about _what_ is not working and give as many hints as possible to _why_ it's not working.**
+3. **Failures are clear about _what_ is not working and give as many hints as possible to _why_ it's not working.**
 
-4. **Tests should not commonly need manual updates when the implementation is changed.**
+4. **Tests do not commonly need manual updates when the implementation is changed.**
    If tests commonly need manual updates, it creates the potential to accidentally change the test so that it doesn't catch a regression introduced in the same change. It also puts more burden on the code reviewers. In addition, updating tests can feel daunting and reduces the motivation to make changes to that part of the codebase.
 
-5. **Tests should be easy to write and maintain.**
+5. **Tests are easy to write and maintain.**
 
 ## Testing UI Code
 
--- a/handbook/engineering/web/testing.md
+++ b/handbook/engineering/web/testing.md
@@ -64,7 +64,7 @@ Even if it results in more code, prefer passing in dependencies as parameters (o
 ### Strive for flat call trees
 
 When writing deeply nested call trees, a test for a unit in the middle of the tree inevitably also tests every layer beneath it.
-This means when a layer below is is changed or breaks, the tests for the layers above it will likely break too.
+This means when a layer below is changed or breaks, the tests for the layers above it will likely break too.
 It also means that the test doesn't just need to pass stubs for the immediate dependencies of the unit, but also for the dependencies of every unit beneath it in the call tree.
 This directly means that if the interface of any unit beneath it changes, the tests of all units above it need to change too, violating goal (4).
 
--- a/handbook/engineering/web/testing.md
+++ b/handbook/engineering/web/testing.md
@@ -61,7 +61,7 @@ Relying on global variables makes dependencies hard to stub for testing (a varia
 It also means figuring out all transitive dependencies that need stubbing for a test is difficult.
 Even if it results in more code, prefer passing in dependencies as parameters (or props for React components), which makes dependencies explicit.
 
-### Strive for flat call trees
+### Favor flat & wide call trees over deep & narrow call trees
 
 When writing deeply nested call trees, a test for a unit in the middle of the tree inevitably also tests every layer beneath it.
 This means when a layer below is changed or breaks, the tests for the layers above it will likely break too.
--- a/handbook/engineering/web/testing.md
+++ b/handbook/engineering/web/testing.md
@@ -151,8 +151,8 @@ describe('ExampleDetailsComponent', () => {
   describe('view mode', () => {
     it('shows the details of the example', ...)
     describe('user has edit permissions', () => {
-      it('shows an edit button if and user has permissions', ...)
-      it('shows a delete button if and user has permissions', ...)
+      it('shows an edit button if user has permissions', ...)
+      it('shows a delete button if user has permissions', ...)
     })
   })
   describe('edit mode', () => {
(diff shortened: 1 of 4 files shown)


handbook/engineering/web/testing.md
```diff
@@ -0,0 +1,177 @@
...
+1. **Our product and code works as intended when we ship it to customers.**
+
+2. **Our product and code doesn't accidentally break as we make changes over time.**
+
+3. **Failures should be clear about _what_ is not working and give as many hints as possible to _why_ it's not working.**
+
+4. **Tests should not commonly need manual updates when the implementation is changed.**
+   If tests commonly need manual updates, it creates the potential to accidentally change the test so that it doesn't catch a regression introduced in the same change. It also puts more burden on the code reviewers. In addition, updating tests can feel daunting and reduces the motivation to make changes to that part of the codebase.
+
+5. **Tests should be easy to write and maintain.**
```
marekweb, line 16: To be more consistent, we could pick between writing all of these goals with "should" or writing them as regular present tense like in 1 and 2

handbook/engineering/web/testing.md
```diff
@@ -0,0 +1,177 @@
...
+
+### Avoid globals
+
+Relying on global variables makes dependencies hard to stub for testing (a variable exported from a module counts as a global).
+It also means figuring out all transitive dependencies that need stubbing for a test is difficult.
+Even if it results in more code, prefer passing in dependencies as parameters (or props for React components), which makes dependencies explicit.
+
+### Strive for flat call trees
+
+When writing deeply nested call trees, a test for a unit in the middle of the tree inevitably also tests every layer beneath it.
```
nicksnyder, line 66: I totally agree with the premise of this section: strive for flat call trees. Given the fact that we won't be able to make all call trees completely flat, I am interested to know how you think about tests for those middle layers. I will share my thoughts here:

> When writing deeply nested call trees, a test for a unit in the middle of the tree inevitably also tests every layer beneath it.

I can't tell if you think this is a good thing or a bad thing, but I actually think this is good.

> This means when a layer below is is changed or breaks, the tests for the layers above it will likely break too.

Here are the scenarios that I see:

1. Layer below breaks so tests for higher layers break. I think this is OK, and even good. Those higher layers are broken so their tests should fail!
1. Layer below changes.
  - If the behavior of the higher layer doesn't change, then the tests shouldn't need to be updated -> good
  - If the behavior of the higher level changes as a result in the change of the lower level, then the tests of that higher layer will need to be changed. This also seems good (tests should change when the behavior at any given layer needs to change).

handbook/engineering/web/testing.md
```diff
@@ -0,0 +1,177 @@
...
+This means when a layer below is is changed or breaks, the tests for the layers above it will likely break too.
```
nicksnyder, line 67:
```suggestion
This means when a layer below is changed or breaks, the tests for the layers above it will likely break too.
```

handbook/engineering/web/testing.md
```diff
@@ -0,0 +1,177 @@
...
+
+Depending on the component, it can be important to test the _render output_ of the component.
+This encompasses the DOM structure and that the correct props are passed to other components.
+For example, when a component renders a link, we want to make sure right destination is passed.
+
+This is accomplished by Jest snapshot tests, using a test renderer for React like `enzyme`.
+While this will inevitably also test parts that would be considered internal implementation details of the component, it generally has a better cost-benefit ratio than the alternative of plucking out specific elements and writing manual assertions.
+
+Testing internal implementation details is generally bad because it is at odds with goal (4).
+Since snapshots are updated automatically though, this concern is not a problem in practice.
```
twop, line 41: I don't think this aligns with my experience

handbook/engineering/web/testing.md
```diff
@@ -0,0 +1,177 @@
...
+Storybook tests can be seen as a form of unit test for a component.
+Any story that is added to our codebase is automatically screenshotted on every CI build (but only in its initial state).
+
+### Render output
+
+Depending on the component, it can be important to test the _render output_ of the component.
+This encompasses the DOM structure and that the correct props are passed to other components.
+For example, when a component renders a link, we want to make sure right destination is passed.
+
+This is accomplished by Jest snapshot tests, using a test renderer for React like `enzyme`.
```
twop, line 37: I will add my thoughts a bit later on but I do consider snapshot tests for react an anti-pattern

felixfbecker: The idea of adding this documentation is to enable a process where you (or anyone else) can open a PR to the handbook to propose & discuss changing our practices 🙂

handbook/engineering/web/testing.md
felixfbecker: Just to be clear, I'm not proposing to make them completely flat - I'm aware that completely flat call trees are unfeasible. What I intended to express with "strive for" was when there is an alternative way to implement a set of code with a flatter & wider call tree compared to a deeper & narrower call tree, favor the flatter & wider call tree. I just changed the title of this section to hopefully make that more clear.

I agree that when only looking at a system with deep & narrow call trees, tests for upper layers generally should fail when a lower layer is broken too. But it does create more work especially during refactors because there are more tests to fix _in addition to fixing the application code_. I'm especially thinking of refactorings here where the **interfaces** change (not behavior) and cause failures or type errors in upper-level tests in ways that _require manual fixes to those upper-level tests_. This doesn't apply to all changes of course, in some cases upper levels may fail and then pass again as soon as you fix the lower level, but then those upper-level errors are still noise and often red herrings on the path to find the root cause.

So what I tried to express here is that when comparing that system to the same system implemented with flat & wide call trees, which doesn't impose that extra work (or at least less of it), this additional work is a negative property of the system with deep & narrow call trees. Yes, they are broken, and within that system they should be reported, but the existence of those breakages stems from a systemic anti pattern that could be avoided. And therefor we strive to keep call trees as flat & narrow as feasible.

nicksnyder: Feel free to merge this PR (this thread is not blocking), I am just interested in discussing :)

> Just to be clear, I'm not proposing to make them completely flat

This was clear to me, my original comment was more about "If I am a dev and I do have a nested call tree, then what do I do and how do I think about writing those tests?"

> But it does create more work especially during refactors because there are more tests to fix in addition to fixing the application code. I'm especially thinking of refactorings here where the interfaces change and cause failures or type errors in upper-level tests in ways that require manual fixes to those upper-level tests.

The only case I can think of where this would be true is where the interface of a lower level changes and we have to manually change the mocks of an upper layer. Are there other cases that you have in mind? If not, it seems like this pain would be non-existent if mocks were generated.

> In this simplified example call graph, to test unit B, we need to provide stubs for all the dependencies of C and D in addition to those of B itself.

Another option to avoid this pain is to not mock out as many things and let the real code execute (i.e., don't mock all the transitive dependencies of B, C, D).

For example, there are no mocks anywhere in https://github.com/nicksnyder/go-i18n/blob/master/v2/goi18n. There are multiple layers to the API, and I have tests at each layer, but higher layers don't mock out lower layers. There is definitely a tradeoff between "if I break the code how many tests fail and how hard is it for me to understand what failed" and "if I want to refactor the code, how painful is it to do so".

PR #1051, opened 2020-06-11
Update company values

nicksnyder:
It is important to have clear and inspiring values as we grow. The current documentation of our values is sparse and could be improved. This is my attempt to do so.

We value being open and transparent. I added this as a core value and removed the other separate page to reduce duplication.

We value people. I added qualities of a successful teammate at Sourcegraph, which is the result of the content and comments on [RFC 126](https://docs.google.com/document/d/1_L-XOT4oPTOuUkT4_JMykneo_wPn2WqC2BXR6_-U5dI/edit#). Please read that RFC for more context.

We value progress. I added more context on what that means at Sourcegraph.

I removed journey because the meaning is not clear. I think it had weird overlap between both people and progress that caused confusion.

We value being bottoms-up and delegating ownership. The best descriptor that I could think of for this was "servant leadership."

I would appreciate feedback from anyone and everyone on the team. We all own our values, I am merely a steward of the team to document them. Do these values and qualities resonate with you? Is there anything missing?


company/teammate_qualities.md
```diff
@@ -0,0 +1,52 @@
+# Teammate qualities
```
twop, line 1: I wonder if this list is intended to be actionable? If yes, for whom?

I'm curios about the qualities because some of them mention skills (e.g. empathy) but some are more of a statement "You believe in our mission and are motivated".

I think I would personally derive more value from a list of "Skills we consider essential" because it will hint which skills I should develop next.

+ I wonder if we should mention "Diversity" in general and making space for "Diversity of opinions" in particular.
To be honest the list below seems very prescriptive of qualities :)

company/values.md
```diff
@@ -1,15 +1,34 @@
 # Sourcegraph values
 
-Our values are:
+Our values reflect who we are and how we work.
 
-## People
+## People first
```
twop, line 5: This paragraph mentions both internal and external context of people. 
Does it make sense to move 
> Together we are advancing technology for the good of people all around the world

to "Customer driven progress"?
And orient this value more "internally"? 
If I understand correctly, values should be designed to be a guiding star in decision making process. If so, I feel like, the description here doesn't provide clear guidance how to prioritize trade offs. Does it make sense?

company/values.md
```diff
@@ -1,15 +1,34 @@
...
 
-Together we are advancing technology for the good of people all around the world. We will attract, hire and retain the best teammates in the world and treat everyone in a first-class manner.
+[Our amazing team](teammate_qualities.md) is what makes Sourcegraph special. Together we are advancing technology for the good of people all around the world and we treat everyone in a first-class manner.
 
-## Journey
+## Customer driven progress
```
twop, line 9: This paragraph reminded me of "Validated learning" concept from "Lean startup"  book. If that was the reference for it maybe make it more explicit and actionable?
Something like:
```
We have a vision and to achieve it we are committed to learn from our customers, validate our assumption quickly via small and intentional iterations
```

company/values.md
```diff
@@ -1,15 +1,34 @@
...
 
-The collection of moments, experiences, and memories that the team shares as we make progress: the light moments, the laughter, the team coming together to solve a problem.
+We have a [vision](strategy.md#vision) to build a product that is loved and valued by every company and developer. We learn what our customers want by understanding the challenges that they face, iterating quickly, and soliciting feedback. We feel good when we help other people, we are passionate about developer tools, and our business succeeds when customers are happy to pay for the value that we provide.
 
-## Progress
+Being customer driven does not mean we build bespoke features. Doing so would be a very inefficient and costly way to achieve our vision. Instead, we use customer feedback as data to help us identify work that would be valuable to a broad set of customers, and then directly engage with specific customers to validate that work.
 
-The continuous march toward achieving our mission as a team: the milestones, the successes, and the breakthroughs.
\ No newline at end of file
+## Open and transparent
```
marekweb, line 15: Is there still room for the phrase "open company" here somewhere? I like the phrase because it illustrates that openness is not just one of our many values, it's at the core of the company.

company/values.md
twop: Also maybe mentions something along the lines:
> we value and leverage our strengths, celebrate differences and be kind to our weaknesses.

Maybe a bit to poetic, but I would love to convey a meaning "it is ok to be a human". I think it nicely connects to  our mission aspect of "sourcegraph is for everybody", thus we need/should to acknowledge that everybody is different

company/values.md
```diff
@@ -1,15 +1,34 @@
 # Sourcegraph values
 
-Our values are:
+Our values reflect who we are and how we work.
 
-## People
+## People first
 
-Together we are advancing technology for the good of people all around the world. We will attract, hire and retain the best teammates in the world and treat everyone in a first-class manner.
+[Our amazing team](teammate_qualities.md) is what makes Sourcegraph special. Together we are advancing technology for the good of people all around the world and we treat everyone in a first-class manner.
```
tsenart, line 7: It's unclear to me what the difference is between company values and team mate qualities. Is there room to unify these?

company/values.md
```diff
@@ -1,15 +1,34 @@
...
+
+Transparency is hard. It requires extra effort and excellent communication. It means it is possible to observe things that are inconsistent or incomplete. We think this is a good thing because it gives everyone visibility into the process, not just the outcome, and an opportunity to contribute to that process at any point in time.
+
+## Servant leadership
+
+Leaders empower everyone at Sourcegraph to be the change that they wish to see.
+
+Decisions and ownership are delegated to the people closest to the problem.
+
+Sourcegraph is a vessel and a playground for our team to do it's best work.
```
mrnugget, line 34:
```suggestion
Sourcegraph is a vessel and a playground for our team to do its best work.
```

company/values.md
```diff
@@ -1,15 +1,34 @@
...
+- Our [code](https://github.com/sourcegraph/sourcegraph)
+- Our [handbook](../handbook/index.md)
+- Our [purpose, mission, vision, strategy, and product direction](strategy.md)
+- Our [OKRs](okrs/index.md)
+
+Being open and transparent builds trust with our customers and within our team. It makes it easy for us to partner and collaborate with our customers so we make Sourcegraph the best that it can be. It enables our teammates to work autonomously to solve tough problems while maintaining alignment.
+
+Transparency is hard. It requires extra effort and excellent communication. It means it is possible to observe things that are inconsistent or incomplete. We think this is a good thing because it gives everyone visibility into the process, not just the outcome, and an opportunity to contribute to that process at any point in time.
+
+## Servant leadership
```
mrnugget, line 28: I like how this works on multiple levels: people in leadership roles practice servant leadership in regards to their team and Sourcegraph itself, as a "vessel and playground", does it to everyone at Sourcegraph.

company/values.md
felixfbecker: What I like about this quote ^ is that it applies not just to the vision for the whole product but also for features and product areas, that it gives room for someone on the team to have an idea, a vision for feature, that may not be derived from customer feedback, but that the teammate makes a strong case for why they believe it would be valuable to customers. These are partially based on assumptions, _but we commit to learn from our customers and validate those assumptions quickly via small and intentional iterations_.

company/values.md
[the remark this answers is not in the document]
gbrik: oh, this is supposed to mean that the leaders are servants. having not seen the term before, it wasn't obvious to me from context. I think even with that context I would prefer something with a word like "flat" in it, but I wouldn't have a negative reaction if I understood the term.

company/teammate_qualities.md
mrnugget: I think I understand (and share) your concerns, @twop, but at the same time, I also think that a "prescriptive" list of traits isn't necessarily a bad thing. It can serve as a guide, as something we all aspire towards. We just need to make sure that it's inclusive enough and nobody feels excluded simply by reading this and thinking "oh, so I'll never be this ideal teammate".

company/values.md
mrnugget: If I had to guess, I'd say that when you combine them into one, you might lose clarity, since some individual traits might start so sound really wishy-washy when raised to the company level.

company/teammate_qualities.md
twop: To play devil's advocate and try to illustrate a possible different perception  (not necessarily mine):
I read the list and cannot help but think: "it is not me". Therefore, "me" right now is not good enough, so I must be someone else, which can apply immense pressure.

Also a lot of points can be flipped:
> Proactive: You do not wait for others to tell you what to do

To be proactive two things must be true:
- You believe that your ideas are good
- You believe that your actions will matter, e.g. somebody cares about what you think and do.

Both of this points are context dependent. Aka it is hard to be proactive in the environment when you are told what to do (aka you need to execute what **I** think is right)

So In my head, "Proactive" is a function of the culture not a "trait" that you can develop with no context. In other words, if you trust people to do their best work "Proactivity score" will go up.

> Responsible: and humbly admit when you could have done better

That is definitely disirable but might be challenging if there is an expectation to do everything "the right way", e.g. no room for mistakes and experiments. In my head, "Responsibility score" can be increased by a) making space for mistakes b) let people own their learnings, which again, context dependent.

I hope that illustration makes sense.

Hot take: my personal preference would be to either remove that part completely or rephrase it in this way:  "traits we strive to develop", and then bullet points should describe why we think they are important for the company.

mrnugget: Yes, agree. I think an additional paragraph that puts this into context would help.

I guess I did have some implicit context already, because I didn't read the list and thought "Oh, I don't do X, that's a problem." Instead, when I for example read this:

 > You deliver the right quality at the right time; you don’t over-engineer or under-engineer a solution.

I thought "haha, yeah, buddy, I wish" because _that_ is IMHO one of the hardest things in software engineering. But at the same time I also thought "that is a fantastic goal. I really want to be able to do that better and I'm actively working on getting better at it".

That just as an explanation on why I don't think "prescriptive" is per se a bad thing :)

nicksnyder:
Superseded by https://github.com/sourcegraph/about/pull/1051

PR #1053, opened 2020-06-12
distribution: add upgrading grafana docs

bobheadxi:
buddy PR to https://github.com/sourcegraph/sourcegraph/pull/11427

* nests grafana stuff under new grafana section
* starting up the Grafana image standalone no longer works without `DISABLE_SOURCEGRAPH_CONFIG`, adds note about this
* add TODOs when upgrading grafana


daxmc99:
This looks good. Does it seem like anything that automatically upgrades _major_ versions of Grafana will need to check that there are no breaking changes? 
I assume the grafana alerts format would not change on patch updates?

bobheadxi:
> Does it seem like anything that automatically upgrades major versions of Grafana will need to check that there are no breaking changes?

I think it should be up the the discretion of whoever is performing the update - while we can expect that non-major versions won't introduce breakages, it probably doesn't hurt to check because sometimes they can 🤷

PR #1059, opened 2020-06-15
Add page to configure charities through gusto (US employees only)

efritz:
A step-by-step guide on how to give some of that sweet, sweet tech money to those that definitely need it more than us.

It may be nice to have a no-judgement opinions page on what charities have good bang-for-your-buck as well (if someone has a passion for a cause, they can add a description of the charity).

I've been using this feature to donate to ACLU and some environmental funds since I found out about the feature. Current circumstances are making many more charities much more prominent and I've myself felt overwhelmed by the choice of charity itself in the past.


handbook/people-ops/index.md
```diff
@@ -19,6 +19,7 @@
 - [Expense reports](expenses.md)
 - [Vacation/FTO](https://github.com/sourcegraph/Graphbook/blob/master/Holidays,%20FTO,%20and%20leaves%20of%20absence/Vacation%20and%20FTO.md)
 - [Holidays](holidays.md)
+- [Charities](charities.md)
```
efritz, line 22: cc @mercadon not sure if there's a better place for this link.

mercadon: I think for now it's probably the best place, but I think if we named it "Donations" might make this a bit more clear for the index.

efritz: https://github.com/sourcegraph/about/pull/1078

PR #1060, opened 2020-06-15
Add page on Go testing tips/guidelines

efritz:
Add tips and patterns for writing Go unit tests.

### DIFF
--- a/handbook/engineering/languages/testing_go_code.md
+++ b/handbook/engineering/languages/testing_go_code.md
@@ -129,6 +129,8 @@ Invocations can be inspected by querying the `History` of a method, which stores
 
 ## Testing time
 
+Testing code that interfaces with wall time is [difficult](https://github.com/sourcegraph/sourcegraph/pull/11426#issuecomment-642428217).
+
 If you have code that requires use of the `"time"` package, your first action should be to refactor the code, if possible, so that the parts that deal with time and the parts that deal with the other testable logic are cleanly separable. If your function calls `time.Now`, see if it is possible to pass the current time as a parameter. If your function sleeps or requires a ticker, see if you can extract the time-irrelevant code into a function that can be tested separately.
 
 If all else fails, you can make use of [efritz/glock](https://github.com/efritz/glock) to mock the behavior of the time package. This requires that the code under test uses a `glock.Clock` value rather than the time package directly (see the [section above](#organizing-code-and-refactoring-for-testability) for tips on refactoring your code). This package provides a _real_ clock implementation, which is a light wrapper over the time package, and a _mock_ clock implementation that allows the clock (and underlying tickers) to be advanced arbitrarily during the test.
--- a/handbook/engineering/languages/testing_go_code.md
+++ b/handbook/engineering/languages/testing_go_code.md
@@ -91,7 +91,7 @@ func TestSprocket(t *testing.T) {
 
 #### Asserting expected complex values
 
-Expected values that are not simple scalars (really, anything not comparable with `==`) should use [go-cmp](github.com/google/go-cmp/cmp) to create a highly-readable diff string.
+Expected values that are not simple scalars (really, anything not comparable with `==`) should use [go-cmp](https://github.com/google/go-cmp/cmp) to create a highly-readable diff string.
 
 ```go
 import "github.com/google/go-cmp/cmp"
--- a/handbook/engineering/languages/testing_go_code.md
+++ b/handbook/engineering/languages/testing_go_code.md
@@ -1,0 +1,136 @@
+# Testing Go code
+
+This document contains tips for writing unit tests for Go code.
+
+## Organizing code and refactoring for testability
+
+If you find yourself having a difficult time testing a piece of Go code, it is likely due to one of the following factors:
+
+- The code does too much
+- The code relies on global state
+- The code relies on real external services
+
+In all of these cases, you will find it easier to test if you break the code into smaller units where each one can be tested independently and without global state or external behavior.
+
+**If your code relies on a global or external service**, then you can refactor that piece of code to instead take a _reference_ to that service. This section runs through an example of how to refactor such code without affecting the callers. The original code relies on a global database connection, but the same issue occurs in code that refers to external APIs (GitHub, Slack, Zoekt, etc).
+
+```go
+func UntestableFunction() (*SomeThing, error) {
+    value, err := GlobalDbConn.DoThing()
+    check(err)
+
+    st := // some things with value
+    return st, nil
+}
+```
+
+First, we break this into two parts: a public function with the same signature, and a new function that injects its dependencies explicitly.
+
+```go
+func UntestableFunction() (*SomeThing, error) {
+    return testableFunction(GlobalDbConn)
+}
+
+func testableFunction(conn *DbConn) (*SomeThing, error) {
+    value, err := conn.DoThing()
+    check(err)
+
+    st := // some things with value
+    return st, nil
+}
+```
+
+If the dependency of `testableFunction` is already something that can be mocked (if the value is an interface, or the struct value can be created easily and has very little internal behavior), then most of the benefit has already been seen. 
+
+Otherwise, we should apply an additional step to make the dependency mockable in the tests. This requires that we extract the _interface_ of the dependency from its implementation, and take any value that conforms to that interface.
+
+```go
+func UntestableFunction() (*SomeThing, error) {
+    return testableFunction(GlobalDbConn)
+}
+
+type IDBConn interface {
+    DoThing() (*Thing, error)
+}
+
+func testableFunction(conn IDBConn) (*SomeThing, error) {
+    value, err := conn.DoThing()
+    check(err)
+
+    st := // some things with value
+    return st, nil
+}
+```
+
+Now, the tests for this new function can create their own `IDBConn` with whatever behavior for `DoThing` that is required. See the section on [mocking](#mocks) below for more details.
+
+## Testing APIs
+
+External HTTP APIs should be tested with the `"httptest"` package.
+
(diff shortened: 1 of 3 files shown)


handbook/engineering/languages/testing_go_code.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Testing time
+
+If you have code that requires use of the `"time"` package, your first action should be to refactor the code, if possible, so that the parts that deal with time and the parts that deal with the other testable logic are cleanly separable. If your function calls `time.Now`, see if it is possible to pass the current time as a parameter. If your function sleeps or requires a ticker, see if you can extract the time-irrelevant code into a function that can be tested separately.
+
+If all else fails, you can make use of [efritz/glock](https://github.com/efritz/glock) to mock the behavior of the time package. This requires that the code under test uses a `glock.Clock` value rather than the time package directly (see the [section above](#organizing-code-and-refactoring-for-testability) for tips on refactoring your code). This package provides a _real_ clock implementation, which is a light wrapper over the time package, and a _mock_ clock implementation that allows the clock (and underlying tickers) to be advanced arbitrarily during the test.
```
rvantonder, line 134: Concrete example: https://github.com/sourcegraph/sourcegraph/pull/11426#issuecomment-642428217

efritz:
> Ultimately I think we should set aside some time to do a "test bash" that follows these guidelines.

That would be great, and I also think that incremental improvements are going to be super beneficial (make sure that you write all new tests following guidelines; if you modify code w/o tests or without outdated tests, update them!)

handbook/engineering/languages/testing_go_code.md
```diff
@@ -0,0 +1,136 @@
...
+```
+
+Now, the tests for this new function can create their own `IDBConn` with whatever behavior for `DoThing` that is required. See the section on [mocking](#mocks) below for more details.
+
+## Testing APIs
+
+External HTTP APIs should be tested with the `"httptest"` package.
+
+For an example usage, see the [bundle manager client tests](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@0cb60598806d68e4c4edace9ed2a801e3f8495bf/-/blob/enterprise/internal/codeintel/bundles/client/bundle_client_test.go#L13), that mock the internal bundle manager service with canned responses. Request values are asserted in the test HTTP handler itself, comparing the requested HTTP method, path, and query args against the expected values.
+
```
mrnugget, line 72:
```suggestion

If you need to test interactions with an external HTTP API, take a look at the `"httptestutil"` package. The `NewRecorder` function can be used to create an HTTP client that records and replays HTTP requests. See [`bitbucketcloud.NewTestClient`](https://github.com/sourcegraph/sourcegraph/blob/f2e55799acad8b6b28cb3b6fd47cc55993d36dc4/internal/extsvc/bitbucketcloud/testing.go#L22-L47) for an example. Or take a look at [other usages of `httptestutil.NewRecorder`](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@master/-/blob/enterprise/internal/campaigns/resolvers/main_test.go#L37:27&tab=references).

```

handbook/engineering/languages/testing_go_code.md
```diff
@@ -0,0 +1,136 @@
...
+	expectedPlanets := []string{
+		"Neptune",
+		"Uranus",    // ...
+		"Pluto",     // we still love you
+		"NOT Venus", // this thing is hot af
+	}
+	if diff := cmp.Diff(expectedPlanets, planets); diff != "" {
+		t.Errorf("unexpected planets (-want +got):\n%s", diff)
+	}
```
mrnugget, line 110:
```suggestion
	}
}
```

handbook/engineering/languages/testing_go_code.md
```diff
@@ -0,0 +1,136 @@
...
+```
+
+*Caveat*: go-cmp uses reflection therefore all comparable fields must be exported.
+
+## Mocks
+
+If your code depends on a value defined as an interface (which all dependencies **should** be), you can use [efritz/go-mockgen](https://github.com/efritz/go-mockgen) to create programmable stubs that conform to the target interface. This replaces an old pattern in the code base that declared mocks defined as [globally settable functions](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@b5b5fc8d885710eb559ff3d6122c9360b31fec78/-/blob/internal/vcs/git/mocks.go#L15).
```
mrnugget, line 117: Nice affiliate link there buddy 😎

handbook/engineering/languages/testing_go_code.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Testing time
+
+If you have code that requires use of the `"time"` package, your first action should be to refactor the code, if possible, so that the parts that deal with time and the parts that deal with the other testable logic are cleanly separable. If your function calls `time.Now`, see if it is possible to pass the current time as a parameter. If your function sleeps or requires a ticker, see if you can extract the time-irrelevant code into a function that can be tested separately.
+
+If all else fails, you can make use of [efritz/glock](https://github.com/efritz/glock) to mock the behavior of the time package. This requires that the code under test uses a `glock.Clock` value rather than the time package directly (see the [section above](#organizing-code-and-refactoring-for-testability) for tips on refactoring your code). This package provides a _real_ clock implementation, which is a light wrapper over the time package, and a _mock_ clock implementation that allows the clock (and underlying tickers) to be advanced arbitrarily during the test.
+
```
mrnugget, line 135: FWIW in `campaigns` and in `repos` we also often accept a `clock func() time.Time`, which is enough for 99% of the cases. For some tests I created a 10-line `testClock` that implements a 2-method interface: https://sourcegraph.com/github.com/sourcegraph/sourcegraph@master/-/blob/enterprise/internal/campaigns/store_test.go#L24-34

handbook/engineering/languages/testing_go_code.md
efritz: I get kickbacks

handbook/engineering/languages/testing_go_code.md
efritz: Looks like you need to update to fit these guidelines then 😛

handbook/engineering/languages/testing_go_code.md
efritz: Added link!

keegancsmith:
This looks like a fun PR for me to upset people who like TDD culture. If the only reason you are using an interface is for testing, then more often than not you are making your code more obscure for cargo cult reasons. Interfaces kill jump to definition and find references, so I think they should only be used when an interface improves code/higher level decoupling, not tests.

handbook/engineering/languages/testing_go_code.md
```diff
@@ -0,0 +1,141 @@
...
+Testing code that interfaces with wall time is [difficult](https://github.com/sourcegraph/sourcegraph/pull/11426#issuecomment-642428217).
+
+If you have code that requires use of the `"time"` package, your first action should be to refactor the code, if possible, so that the parts that deal with time and the parts that deal with the other testable logic are cleanly separable. If your function calls `time.Now`, see if it is possible to pass the current time as a parameter. If your function sleeps or requires a ticker, see if you can extract the time-irrelevant code into a function that can be tested separately.
+
+If all else fails, you can make use of [efritz/glock](https://github.com/efritz/glock) to mock the behavior of the time package. This requires that the code under test uses a `glock.Clock` value rather than the time package directly (see the [section above](#organizing-code-and-refactoring-for-testability) for tips on refactoring your code). This package provides a _real_ clock implementation, which is a light wrapper over the time package, and a _mock_ clock implementation that allows the clock (and underlying tickers) to be advanced arbitrarily during the test.
```
daxmc99, line 139: This looks great! I feel like I may have already committed some sins here with some of the code written that depends on `time.Now()` https://github.com/sourcegraph/sourcegraph/blob/b65f041fc2b2c363df40802f595418031a2ad2db/internal/version/version.go#L40-L60

efritz: Super easy to make that testable by converting it to something like `func HowLongOutOfDate(now time.Time) (int, error) {`

efritz:
> cargo cult reasons

I don't think making an interface in order to improve the separation of concerns between units cargo cult.

> Interfaces kill jump to definition and find references

Then use go to implementation instead. That works, no?

> so I think they should only be used when an interface improves code/higher level decoupling, not tests

That's reasonable. Can you show an example in sg/sg where a struct is better than an interface (but still being testable without having to mock the dependency)?

I'm assuming that anything with a non-trivial amount of behavior/logic is going to either be (1) difficult to test, or (2) difficult to test in isolation (you will end up testing your dependency's internal behavior in the caller and the unit tests for the dependency).

keegancsmith:
> > cargo cult reasons
> 
> I don't think making an interface in order to improve the separation of concerns between units cargo cult.

I don't think it is. I was referring to doing it for the sole purpose of making code testable. However, I also don't think one should do it by default. If you only ever have one caller, what is the point? Whenever you update the implementation API you end up having to update the interface as well.

> > Interfaces kill jump to definition and find references
> 
> Then use go to implementation instead. That works, no?

Find references though? Depends if you declare an interface that everyone imports vs declare smaller interfaces at consumers (the latter is probably cleaner). Then you really want a find references on Union(structs and interfaces of this type).

> > so I think they should only be used when an interface improves code/higher level decoupling, not tests
> 
> That's reasonable. Can you show an example in sg/sg where a struct is better than an interface (but still being testable without having to mock the dependency)?

You may not like it, but its the places we have a global func call mockFoo, then just use that. Those are likely better done without global state (ie a struct which contains dependencies). But its pretty effective and only looks ugly if you believe in mantras like all globals are bad.

> I'm assuming that anything with a non-trivial amount of behavior/logic is going to either be (1) difficult to test, or (2) difficult to test in isolation (you will end up testing your dependency's internal behavior in the caller and the unit tests for the dependency).

As long as you not asserting internal dependency stuff in your test, is this test not better? IE you don't end up testing what your mock does, you end up testing what the real thing does? You test reality vs test developer assumptions.

If a dependency you use changes its behaviour, your tests breaking are a good thing since it exposes an assumption the dependency developer wasn't aware of / etc.

If running E2E test was fast and reliable, would those not be the best tests vs unit tests? The inbetween where you still keep fast and reliable sounds good to me.

#devilsadvocate

handbook/engineering/languages/testing_go_code.md
daxmc99: Refactored to add a test https://github.com/sourcegraph/sourcegraph/pull/11697

PR #1061, opened 2020-06-15
Exchange feedback with candidates

nicksnyder:
This PR is mostly to start a discussion (I didn't spend much time thinking about the words in the diff).

I think we should figure out how to exchange feedback with candidates.

1. We value people and transparency, and I think sharing feedback that we collect in our interview process would be something that makes our candidate experience exceptional. I have spoken with other engineering managers who have done this and made it part of their interview process (i.e., send feedback and see how candidates respond).
2. We should be more explicitly soliciting feedback from candidates on our interview process (we used to do surveys but haven't recently AFAIK), but this is hard to justify if we don't provide feedback ourselves.
3. Communicating reasons gives the candidate an understanding if/when they should apply again in the future, and also if they know of anyone who could be a better fit. I have had multiple candidates refer someone else after I rejected them and gave them specific reasons! Examples from the last month:

    > I wanted to say thank you again for your transparency, and for the nicest rejection letter I've ever received! ... I have been telling a few friends about ya'll
    
    > Understood and thanks for the follow up. I enjoyed chatting with you all. I may have a suggestion for someone that might be better for the role.

In practice, I think this means that hiring managers should:
1. Own sending rejection notes (not recruiting coordinator)
2. Have their own template in their own voice. That template should include a place to plug in custom context for that candidate.

### DIFF
--- a/handbook/people-ops/hiring.md
+++ b/handbook/people-ops/hiring.md
@@ -59,7 +59,7 @@ See [interview process](interview_process.md).
 
 We value transparency and like to exchange feedback with candidates.
 
-If you are a hiring manager who has decided to not move forward with a candidate at any stage in our hiring process, you are encourage to be as transparent as possible with the candidate when communicating the rejection. Use your judgement and ask for help crafting a message if you are unsure.
+If you are a hiring manager who has decided to not move forward with a candidate at any stage in our hiring process, you are encouraged to be as transparent as possible with the candidate when communicating the rejection. Use your judgement and ask for help crafting a message if you are unsure.
 
 If you are a candidate, we would love to know how you think we can make our hiring process better.
 
--- a/handbook/people-ops/hiring.md
+++ b/handbook/people-ops/hiring.md
@@ -55,18 +55,13 @@ For any members of a hiring team, please see [this folder](https://drive.google.
 
 See [interview process](interview_process.md).
 
-## Sharing interview feedback
+## Exchanging feedback
 
-We don't share specific reasons or feedback if we [decide](#decision-making-process) not to move forward in our hiring process with a candidate.
+We value transparency and like to exchange feedback with candidates.
 
-It would be nice if we could give feedback like this, but we don't do this today. Why?
+If you are a hiring manager who has decided to not move forward with a candidate at any stage in our hiring process, you are encourage to be as transparent as possible with the candidate when communicating the rejection. Use your judgement and ask for help crafting a message if you are unsure.
 
-- The written feedback that we collect during our interview process is intended for an internal audience. We don't ask our interviewers to spend time and effort packaging their feedback in a way that can be delivered directly to candidates.
-- Adequately summarizing all interview feedback for a candidate is challenging and time consuming. We choose to spend our time focusing on delivering developmental feedback to our existing teammates.
-
-If any interviewer is asked by a candidate for interview feedback, they may respond with the following message:
-
-> I understand the desire to receive feedback from our interview process, but unfortunately we are not in a position to give it. Here is some public documentation to that effect: https://about.sourcegraph.com/handbook/people-ops/hiring#sharing-interview-feedback.
+If you are a candidate, we would love to know how you think we can make our hiring process better.
 
 ## Respecting time
 


christinaforney:
I support this and would love to give candidates more information. At the same time, I receive requests for feedback from a lot of candidates that were rejected in the application stage. I do not feel like I would be able to sustain providing feedback to all these applicants who request it. What are your thoughts on this? I view application feedback quite differently from feedback for candidates who I have spoken to.

As written it sounds like every rejection should include more transparency and why? Or just when a candidate reaches out for feedback on the process?

nicksnyder:
I agree that feedback looks different depending on which stage the candidate is in the process.

For applications, I mostly have a template that is honest (e.g., "...we received a lot of applicants and I can't move forward with everyone...") and in my own voice. It is from me and it uses first person (e.g. "I reviewed your application" not "we reviewed your application"). If I liked something about their application, I will start with the template and add some extra color to make my appreciation specific to their application even if we aren't moving forward. If the candidate was close (e.g., a promising junior candidate that I can't move forward with right now), then I will add a sentence about that to encourage them to apply in the future. Some applications are clearly not worth it and I don't spend any time customizing the template.

For candidates who make it to screens/interviews, I think the goal is to provide enough information that they don't feel the need to reply with "why?". 99% of my replies now are "thanks for letting me know". While I still do start with a template for these candidates, much more of the body is custom.


> As written it sounds like every rejection should include more transparency and why? Or just when a candidate reaches out for feedback on the process?

Just to re-iterate, I do think we should proactively be more transparent at all phases, but transparency looks different at each phase.

christinaforney:
That makes sense to me. I'm still a bit worried about the time investment for me in application review and feedback, but I agree 100% with everything else you've said. Perhaps this is just me needing to spend some time to find the right template for my application rejections.

handbook/people-ops/hiring.md
```diff
@@ -55,18 +55,13 @@ For any members of a hiring team, please see [this folder](https://drive.google.
...
 See [interview process](interview_process.md).
 
-## Sharing interview feedback
+## Exchanging feedback
 
-We don't share specific reasons or feedback if we [decide](#decision-making-process) not to move forward in our hiring process with a candidate.
+We value transparency and like to exchange feedback with candidates.
 
-It would be nice if we could give feedback like this, but we don't do this today. Why?
+If you are a hiring manager who has decided to not move forward with a candidate at any stage in our hiring process, you are encourage to be as transparent as possible with the candidate when communicating the rejection. Use your judgement and ask for help crafting a message if you are unsure.
```
nicksnyder, line 62:
```suggestion
If you are a hiring manager who has decided to not move forward with a candidate at any stage in our hiring process, you are encouraged to be as transparent as possible with the candidate when communicating the rejection. Use your judgement and ask for help crafting a message if you are unsure.
```

nicksnyder:
> I'm still a bit worried about the time investment for me in application review and feedback

The existing docs say "we don't do this" which is too strong I think. The docs in the proposed diff (which I didn't think too much about) says that giving feedback is "encouraged" (but not required). In practice, it is not common for me to modify beyond my personal template for application stage, but if the opportunity presents itself then I do because I think it is good.

sqs:
@nicksnyder You should try this yourself for a month and see how it goes. The idea is nice, but it is probably best to do on a case-by-case basis. It can also have the unintended effect of making it so people are less likely to "take a chance" and interview a candidate who has a lower perceived chance of offer but who brings something new to the team.

nicksnyder:
> It can also have the unintended effect of making it so people are less likely to "take a chance" and interview a candidate who has a lower perceived chance of offer but who brings something new to the team.

I don't understand this concern because it has had the exact opposite effect on me. If this a side effect for other people then I think it points to the ineffectiveness of our docs communicating what we are trying to achieve here. I am more, not less, willing to take bets on candidates if I know that we might be building a longer term relationship even if they aren't a fit for the roles we have today. I have found these personalized responses are most helpful in _exactly_ the situation you described. I take bets on candidates, have an intro call, realize they are great people but not a fit for the role we have today, and I send an honest rejection email to them that they very much appreciate. This encourages them to apply later on if/when they are a fit (and they understand what that means), and/or refer other people who they know would be a fit, which optimizes for long term hiring. 

> You should try this yourself for a month and see how it goes. The idea is nice, but it is probably best to do on a case-by-case basis.

Tomás and I have been and will continue to do this. It is definitely case-by-case, and I don't feel obligated to. I just don't want people to feel obligated not to.

To be clear, the diff on this PR is aspirational and I don't think it is ready to be merged today. There may be some other diff we could merge today that makes it clear feedback isn't prohibited.

sqs:
@nicksnyder How is it going trying this?

nicksnyder:
@tsenart and I have collaborated on crafting messages for certain candidates. @tsenart how were these messages received?

tsenart:
> @tsenart how were these messages received?

The majority of time there's no significant response from the candidate to this feedback.

PR #1064, opened 2020-06-16
Delete unused PricingTable.tsx

nicksnyder:
This appears to be an old/unused pricing table. I found it when I searched for instances of "whitelist"

Part of https://github.com/sourcegraph/sourcegraph/issues/11389


nicksnyder:
@sqs @dadlerj you are on the blame for this file, can you confirm this is no longer needed?

dadlerj:
Confirmed — feel free to delete, thanks!

PR #1068, opened 2020-06-16
Add Slack messaging during off-hours policy

dadlerj:


ebrodymoore:
I'll leave it up to you on whether this is the place for it, but some related advice I've gotten is to be really intentional with written communication (especially with different time zones) because otherwise an async conversation can take days and possible slow your work down

dadlerj:
@ebrodymoore I think that's great advice in general for working remote! I'm not convinced that it should be part of this section though (which is specifically about Slack messages). What do you think about adding it to https://about.sourcegraph.com/handbook/communication?

ebrodymoore:
@dadlerj already here, actually! https://about.sourcegraph.com/company/remote/tips

PR #1088, opened 2020-06-20
Fixes #967 - update contributors script with PRs and pull team members

### LINKED ISSUE #967: Revise generate contributors script
The [generate_contributors.go script](https://sourcegraph.com/github.com/sourcegraph/about/-/blob/bin/generate_contributors.go) needs to be revised to:

- Pull in users who have submitted a pull request
- Automatically exclude members of the Sourcegraph team without having to maintain a manual exclude list

christinaforney:
I updated the script to:
* Automatically pull the list of Sourcegraph team members to exclude from the list
* Add PR contributions (previously was only GitHub issue authors)

This is the first Go code I've written and I followed patterns already in the script, so I'd love some feedback from someone more experienced than me :)


bin/generate_contributors.go
```diff
@@ -22,6 +22,15 @@ type IssueAuthorDetails struct {
 	url    string
 }
 
+func MapUsers(vs []*github.User, f func(*github.User) string) map[string]string {
```
efritz, line 25: If this is to simulate a set of strings, you can do `map[string]struct{}` instead, as the values would be zero-width values.

bin/generate_contributors.go
```diff
@@ -120,6 +90,37 @@ func main() {
 			}
 			page++
 		}
+
+		// Get all created PR contributors
+		prRangeOk = true
+		page = 0
+
+		for page < 15 && prRangeOk{
```
efritz, line 98: Why 15? Just an upper bound?

bin/generate_contributors.go
```diff
@@ -120,6 +90,37 @@ func main() {
...
+			}
+
+			if len(prs) == 0 {
+				break
+			}
+
+			for _, pr := range prs {
+				
+				if pr.CreatedAt.Before(issueDate) {
+					prRangeOk = false
```
efritz, line 113: You can label loops and break out of multiple, e.g.

```go
outer: for page < 15 {
   for {
       break outer
    }
}
```

christinaforney: 🤯  so cool!

bin/generate_contributors.go
christinaforney: Yeah, I kept this from the original script - is there a better way to manage this?

bin/generate_contributors.go
christinaforney: This whole thing is just so I can apply a function to each element in my list, and create a new list of the result. What I would want to do in Ruby down below is:
```
sourcegraphTeam.map{ |m| m.Login }
```

bin/generate_contributors.go
efritz: Is this just to stop a runaway loop? Why not just let the API dictate when the results are done (you already break when the issue/pr is too old)?

bin/generate_contributors.go
efritz: If you don't care about uniqueness,

```
transformed := make([]string, 0, len(vs))
for _, v := range vs {
    transformed = append(transformed, f(v))
}

return transformed
```

bin/generate_contributors.go
christinaforney: Yeah, I think that's why it was there before. Good point, will fix!

bin/generate_contributors.go
christinaforney: 🎉  sweet! Thanks, that's what I needed.

bin/generate_contributors.go
```diff
@@ -36,87 +36,78 @@ func main() {
...
+	for _, m := range sourcegraphTeamMembers {
+		sourcegraphUsernames[*m.Login] = struct{}{}
+	}
+
+	contributors := make(map[string]ContributorDetails)
 	page := 0
 
 	// First print via repo
 	for _, repo := range repos {
+		page = 0
```
efritz, line 61: This value is already zero due to assignment above

christinaforney:
```suggestion
```

bin/generate_contributors.go
```diff
@@ -36,87 +36,78 @@ func main() {
...
```
christinaforney, line 62:
```suggestion
```

PR #1096, opened 2020-06-23
add Just Workplace entry to values

gbrik:
I've found this to be generally true, and think we are working towards making it more true, so I took a stab at writing it down!

I'm less sure about including the examples. I think they're really helpful to make concrete what we mean by the more abstract values, and also help prove that these aren't just things that we say.

This is intended to be a starting point for discussion of what our goals are here and how we want to communicate them, though I would be OK with including it in the handbook as is.

### DIFF
--- a/company/values.md
+++ b/company/values.md
@@ -14,12 +14,15 @@ The collection of moments, experiences, and memories that the team shares as we 
 
 The continuous march toward achieving our mission as a team: the milestones, the successes, and the breakthroughs.
 
-## Just Workplace
+## Ethics & Justice First
 
 - We strive to work against and avoid reproducing systemic injustices. 
 - We're grateful and not defensive when people point out our failings here.
-- All teammates are empowered to create an accessible and inclusive work environment, and encouraged to spend their work hours doing so whenever they see an opportunity.
+- This is not an attempt to define our exact ethical framework, but documents how thinking about ethics is part of our process and reinforces a culture of talking about ethics rather than avoiding them.
+- When any teammate notices any unethical or unjust practice, they should spend work hours bringing attention to the issue and advocating to fix it.
+- Ethical concerns are blockers; we do not release and resolve these issues later.
 - Some examples of what this has looked like in practice:
   - Juneteenth is a paid holiday to take anti-racist actions.
   - Updating the team page and e-mail signatures to include pronouns.
   - Enabling a slack bot to help teammates learn about uninclusive or discriminatory language.
+  - Intentional outbound hiring with the goal of diversifying our hiring pipeline.
--- a/company/values.md
+++ b/company/values.md
@@ -13,3 +13,13 @@ The collection of moments, experiences, and memories that the team shares as we 
 ## Progress
 
 The continuous march toward achieving our mission as a team: the milestones, the successes, and the breakthroughs.
+
+## Just Workplace
+
+- We strive to work against and avoid reproducing systemic injustices. 
+- We're grateful and not defensive when people point out our failings here.
+- All teammates are empowered to create an accessible and inclusive work environment, and encouraged to spend their work hours doing so whenever they see an opportunity.
+- Some examples of what this has looked like in practice:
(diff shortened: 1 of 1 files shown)


nicksnyder:
I like the direction here!

What do you think about placing this content under the proposed "People first" value in [this PR](https://github.com/sourcegraph/about/pull/1051)? I think this content is relevant to that value and it would clarify that value more (currently it is the least well specified in that PR).

On naming, I would worry that "Just" could come off wrong because of its double meaning (i.e., "simply/barely/only/no more than a workplace"). This is another reason why I think including it under "people first" could be nice. Other words that could work: fair, equitable, inclusive.

I like providing proof points for our values, but something about inlining examples here makes me hesitate. Will the list be maintained? I think we **should** have a source of truth for all the things we do to create an inclusive workplace, and https://about.sourcegraph.com/handbook/people-ops/teammate_philosophy is a start, so maybe we should build that list elsewhere and link to that page from our values.

gbrik:
> What do you think about placing this content under the proposed "People first" value in [this PR](https://github.com/sourcegraph/about/pull/1051)? I think this content is relevant to that value and it would clarify that value more (currently it is the least well specified in that PR).

It might make sense to move it into that PR. I was a bit hesitant as there hadn't been any movement on that PR in a while, and there's a lot of different conversational threads so adding another seems like it could add to the confusion. I wonder if PRs are the wrong format for this sort of conversation? I feel like ideating/rapid iteration works better in google docs.

Separately, I think there is more here than fits under 'people first', I will do a second pass tmrw to make this more clear.  There needs to be somewhere we capture the unjust things a company can do without treating its employees poorly, but I agree that this ties in closely with the latter.

> On naming, I would worry that "Just" could come off wrong because of its double meaning as an adverb (i.e., "simply/barely/only/no more than a workplace"). I assume you were aiming for the adjective definition (i.e. "based on or behaving according to what is morally right and fair."). This is another reason why I think including it under "people first" could be nice. Other words that could work: fair, equitable, inclusive.

Maybe 'ethical'? I would like this bullet to cover not just how we treat employees but our ethics in general. I don't want to describe our ethics, I'm sure none of us share the same exact ethical framework I want to document that we are encouraged to think and talk about the ethics of our decisions (and our colleague's decisions) as part of our job, which seems obvious when you write it out but is not the industry standard in my experience.

> I like providing proof points for our values, but something about inlining examples here makes me hesitate. Will the list be maintained? I think we **should** have a source of truth for all the things we do to create an inclusive workplace, and https://about.sourcegraph.com/handbook/people-ops/teammate_philosophy is a start, so maybe we should build that list elsewhere and link to that page from our values.

Is there somewhere else we can put them where they will be more likely to be maintained? I agree that if we want to keep a complete + growing list, it will get too big to reasonably fit in this doc unless hidden by default.

This did make me think that it would be cool to have an automated process which reminds us to maintain the proof points on our values, which will make it very obvious when we're not getting better as an organization at fulfilling those values. This also frames the proof points as a way for us to keep ourselves accountable. I feel like this could help a lot with our values page not being "Don't be evil" 2.0.

gbrik:
@nicksnyder updated!

company/values.md
```diff
@@ -13,3 +13,16 @@ The collection of moments, experiences, and memories that the team shares as we
 ## Progress
 
 The continuous march toward achieving our mission as a team: the milestones, the successes, and the breakthroughs.
+
+## Ethics & Justice First
+
+- We strive to work against and avoid reproducing systemic injustices. 
+- We're grateful and not defensive when people point out our failings here.
+- This is not an attempt to define our exact ethical framework, but documents how thinking about ethics is part of our process and reinforces a culture of talking about ethics rather than avoiding them.
```
gbrik, line 21: I still think we can and should enumerate points of ethics that we all agree on. General points like anti-racism/homophobia/transphobia/ableism. Are there business or pricing strategies which we can commit to never using? etc

gbrik:
Closing and moving discussion. Didn't feel like we were getting as much input as we want because of the PR being public.

PR #1101, opened 2020-06-24
New design

sqs:
Making a PR just so we get Netlify deploy previews instead of using my ngrok tunnel. Feel free to not use this branch and use another. (This branch does have a lot of the copy and code changes already in it that could save time, though!)

Deploy preview: https://deploy-preview-1101--sourcegraph.netlify.app/

New copy: https://docs.google.com/document/d/1LFyCo5PtLu94ZUCxvYKn_aldeKrYTel6g2mFjNZnkGo/edit

---

- [x] Navbar should have an inner container so that it is no wider than the main body content 
image
- [x] Navbar on customers page should not have its own background; it should have the blue background of the customers hero IMO
- [x] When navbar is fixed (ie when you scroll down) and it stays there, it should have a 1px gray bottom border IMO or else the boundary between the fixed navbar and the content looks unintended


website/src/css/styles.scss
```diff
@@ -82,20 +72,17 @@ $spacers: (
     ),
 );
 
-$body-bg: $black;
-$body-color: $white;
+$body-bg: #f9f9f9;
+$body-color: #222233;
```
felixfbecker, line 76: Could you delete the text color override of the cookiebot styles (or remove it altogether) to make sure it uses the same text color as the body? (The override was only there because previously the default body color was white, resulting in white-on-white text)

aileenrose: Yes.  This branch will be synced with origin (master).

felixfbecker: I see no changes in this PR, the color overrides still seem to exist in master:

https://github.com/sourcegraph/about/blob/14ce1719ce81e64cb7572d2097bbbc7fe29e451f/website/src/css/components/_Cookiebot.scss#L5

PR #1116, opened 2020-06-27
Sharing company financials with candidates

sqs:
Some candidates ask these questions. I want candidates to know it is OK and encouraged for them to ask these kinds of questions so they can understand the offer and the opportunity here.

### DIFF
--- a/handbook/people-ops/hiring.md
+++ b/handbook/people-ops/hiring.md
@@ -74,7 +74,7 @@ Interviewing requires time and effort, both for us and for candidates. The hirin
 
 ## Sharing company financials
 
-Choosing to join a company is a significant commitment. We want people to have an accurate view of Sourcegraph's financial and business position before they join. (Every Sourcegraph team member after they join has regular access to this information by viewing the board slides or asking the CEO directly.)
+Choosing to join a company is a significant commitment. We want people to have an accurate view of Sourcegraph's financial and business position before they join. (Every Sourcegraph team member after they join has regular access to this information on our Google Drive, and they can ask the CEO directly if they have any other questions.)
 
 Here are some related questions candidates have asked in the [CEO interview](../ceo/index.md#interviews-with-me) or after receiving an offer. We don't make all this information public in the handbook, but candidates are encouraged to ask the CEO. (Note from the CEO: I love when candidates ask these questions!)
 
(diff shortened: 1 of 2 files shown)


handbook/people-ops/hiring.md
```diff
@@ -71,3 +71,16 @@ If any interviewer is asked by a candidate for interview feedback, they may resp
 ## Respecting time
 
 Interviewing requires time and effort, both for us and for candidates. The hiring manager is responsible for monitoring feedback during the interview phase, and if it becomes obvious that we won't be moving forward with a candidate, then the hiring manager should cancel any remaining interviews to be respectful of both the candidate's and interviewers' time.
+
+## Sharing company financials
+
+Choosing to join a company is a significant commitment. We want people to have an accurate view of Sourcegraph's financial and business position before they join. (Every Sourcegraph team member after they join has regular access to this information by viewing the board slides or asking the CEO directly.)
```
dadlerj, line 77:
```suggestion
Choosing to join a company is a significant commitment. We want people to have an accurate view of Sourcegraph's financial and business position before they join. (Every Sourcegraph team member after they join has regular access to this information by viewing the board slides or asking the CEO directly.)
```

Can you clarify that "asking the CEO directly" is not actually needed here? It's an odd addition, given all of our top-line forecasts/planning and results are already publicly available in GDrive. And anyone can get access to our full F/S by requesting them from the CEO _or_ anyone on the Ops team. My point is that this sentence makes it sound more secretive than it actually is!

handbook/people-ops/hiring.md
```diff
@@ -71,3 +71,16 @@ If any interviewer is asked by a candidate for interview feedback, they may resp
...
+
+Here are some related questions candidates have asked in the [CEO interview](../ceo/index.md#interviews-with-me) or after receiving an offer. We don't make all this information public in the handbook, but candidates are encouraged to ask the CEO. (Note from the CEO: I love when candidates ask these questions!)
+
+1. What percentage of the company does my option grant represent?
```
dadlerj, line 81: We actually already share this during the initial offer calls CC @mercadon (it's also essentially written out in the offer letters — we state the latest private round valuation, so it can be calculated). Did you not expect that to be the case? I think it's right to be transparent about this at that stage.

sqs: I expected it to be shared and agree it's right. There was a candidate who did not know this figure. I'll follow up on Slack.

handbook/people-ops/hiring.md
sqs: Fixed. Also added you as someone people can ask for the precise answers (not just me).

PR #1123, opened 2020-06-29
rm docsite.json (docsite config is set in private config repo

beyang:
It appears the docsite config is set in a private config repo, and this docsite.json is overridden. So its existence is misleading.


sqs:
This file is still useful. The comment in it says:

> This configures the documentation server used in local development only.

And you can run `docsite serve` in the root of this repo to preview the handbook locally.

sqs:
https://github.com/sourcegraph/about/pull/1131 clarifies this.

PR #1124, opened 2020-06-29
Add security team responsibilities

nicksnyder:


tsenart:
@chayim: While we currently don't have a security team, it'll likely eventually exist on its own. This future organization is what we didn't have time to cover in our 1:1 today. Let's chat about it on Thursday.

PR #1133, opened 2020-06-30
document how to set up Zoom recordings to go to Slack automatically

slimsag:
It's important for remote team members to have access to recordings when they cannot attend a meeting (e.g. a team sync or all-hands company sync). Allowing others to watch these meetings and contribute post-mortem thanks to our strong written culture is incredibly useful.

However, this relies on someone recording the meeting and posting the link properly - and unfortunately Zoom makes this 1000x harder than it should be. I have found the following setup to configure a Slack email relay + gmail forwarding filter extremely useful as it automatically records our Distribution team sync and posts the recording in our Slack channel:

image

I suspect that other teams may find this useful, and it may be useful for the company meeting in particular.

Additionally, because Zoom meetings can be reused at any time, you can cheat by using the same Zoom link as an easy way to start a recorded call and have the result posted to Slack. For example in the Distribution team we have this in Slack:

image

If there is a topic where having a recording makes sense for posterity or other reasons, you can just use that link and it'll post to Slack automatically with the recording (of course this is not a substitute for proper note taking).

### DIFF
--- a/handbook/engineering/index.md
+++ b/handbook/engineering/index.md
@@ -16,6 +16,7 @@
 - [Continuous releasability](continuous_releasability.md)
 - [Commit message guidelines](commit_messages.md)
 - [Ignoring editor config files in Git](ignoring_editor_config_files.md)
+- [Configuring Zoom to send recordings to Slack automatically](configuring_zoom_recordings_to_slack_automatically.md)
 - [Incidents](incidents.md)
 - [Releases](releases/index.md)
   - [Release issue template](releases/release_issue_template.md)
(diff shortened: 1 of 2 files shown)


handbook/engineering/index.md
```diff
@@ -16,6 +16,7 @@
 - [Continuous releasability](continuous_releasability.md)
 - [Commit message guidelines](commit_messages.md)
 - [Ignoring editor config files in Git](ignoring_editor_config_files.md)
+- [Configuring Zoom to send recordings to Slack automatically](configuring_zoom_recordings_to_slack_automatically.md)
```
bobheadxi, line 19: I remember some discussion in slack a while back about disabling zoom auto-recording - maybe zoom tips in general warrants its own page? ie `zoom.md` -> section `zoom recordings ...`

slimsag: That would be great to have indeed!

PR #1134, opened 2020-06-30
Clarify team structure

christinaforney:
The homepage of the handbook is essentially a list of the teams we have at Sourcegraph, and this information does not exist anywhere else. This minor hierarchy change helps to clarify this and fill the gap.

I brought the engineering team list up a level to help with this, but understand it's now duplicated  (which is sub-optimal). Open to ideas!


sqs:
1. Duplicating the eng team list is fine. Just add Markdown comments in both places mentioning it needs to be updated in both places in sync.
1. I don't like making "teams" the top-level navigation concept. It optimizes for the writer (and specifically the manager of the team), not the reader. It also makes it feel like you can't/shouldn't edit other teams' pages. You could add a note to this page saying that these headers are roughly the teams at Sourcegraph; that would convey the info without having these negative consequences.

christinaforney:
@sqs - does the update I made resolve your concerns?
@pecigonzalo - does this change help clarify your request for better understanding the teams at Sourcegraph?

PR #1151, opened 2020-07-03
Add proper cookiebot styles

felixfbecker:
The Cookiebot template is now defined with Bootstrap classes that work on both our website and the webapp, in both light and dark theme.

The style here is currently needed because cards are defined as having a white background, but the body text color is also defined as white text. @sqs I'll let you handle this as I assume with the new homepage design that has a white background the default body color will be changed to be dark and this whole file can be deleted then.


felixfbecker:
Merging since this is already deployed and currently looks very bad (white text on white background)

PR #1161, opened 2020-07-06
Schedule interviews according to candidate preferences

nicksnyder:
There is no one-size-fits-all solution. We should ask candidates what they want and then try to accommodate the best we can.

### DIFF
--- a/handbook/people-ops/interview_process.md
+++ b/handbook/people-ops/interview_process.md
@@ -15,11 +15,11 @@ We want to mimimize the bias in our hiring process, so we will try to avoid sche
 
 ## Scheduling
 
-We are flexible when it comes to scheduling interviews because we are all-remote and we don't need to schedule all interviews back-to-back (unlike typical onsite interviews at other companies).
+We are flexible when it comes to scheduling interviews because we are all-remote and we don't need to schedule all interviews back-to-back (unlike typical onsite interviews at other companies). Some candiates like spreading interviews out over multiple days and others prefer to batch them as much as possible to get it over with. *Ask the candidate what their preferences are and then try to accomidate those preferences as much as possible (given interviewer availability)*.
 
-- We try to schedule interviews as soon as possible given the availability of the candidate and our interviewers.
-- We try to avoid more than 2 hours of consecutive interviews for a candidate.
-- If interviews are spread over multiple days and it is clear from the initial feedback that we won't be moving forward, we will cancel the remaining scheduled interviews to save ourselves and the candidate time.
+Even if candidates prefer to batch all their interviews together, it might not be possible due to the timezones of available interviewers.
+
+If interviews are spread over multiple days and it is clear from the initial feedback that we won't be moving forward, we will cancel the remaining scheduled interviews to save ourselves and the candidate time.
 
 ## Writing feedback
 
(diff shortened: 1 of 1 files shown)


handbook/people-ops/interview_process.md
```diff
@@ -15,11 +15,11 @@ We want to mimimize the bias in our hiring process, so we will try to avoid sche
 
 ## Scheduling
 
-We are flexible when it comes to scheduling interviews because we are all-remote and we don't need to schedule all interviews back-to-back (unlike typical onsite interviews at other companies).
+We are flexible when it comes to scheduling interviews because we are all-remote and we don't need to schedule all interviews back-to-back (unlike typical onsite interviews at other companies). Some candiates like spreading interviews out over multiple days and others prefer to batch them as much as possible to get it over with. *Ask the candidate what their preferences are and then try to accomidate those preferences as much as possible (given interviewer availability)*.
```
eseliger, line 18:
```suggestion
We are flexible when it comes to scheduling interviews because we are all-remote and we don't need to schedule all interviews back-to-back (unlike typical onsite interviews at other companies). Some candiates like spreading interviews out over multiple days and others prefer to batch them as much as possible to get it over with. *Ask the candidate what their preferences are and then try to accommodate those preferences as much as possible (given interviewer availability)*.
```

handbook/people-ops/interview_process.md
```diff
@@ -15,11 +15,11 @@ We want to mimimize the bias in our hiring process, so we will try to avoid sche
 
 ## Scheduling
 
-We are flexible when it comes to scheduling interviews because we are all-remote and we don't need to schedule all interviews back-to-back (unlike typical onsite interviews at other companies).
+We are flexible when it comes to scheduling interviews because we are all-remote and we don't need to schedule all interviews back-to-back (unlike typical onsite interviews at other companies). Some candiates like spreading interviews out over multiple days and others prefer to batch them as much as possible to get it over with. *Ask the candidate what their preferences are and then try to accommodate those preferences as much as possible (given interviewer availability)*.
```
slimsag, line 18:
```suggestion
We are flexible when it comes to scheduling interviews because we are all-remote and we don't need to schedule all interviews back-to-back (unlike typical onsite interviews at other companies). Some candiates like spreading interviews out over multiple days and others prefer to batch them as much as possible to get it over with. *Ask the candidate what their preferences are / if they prefer having a 30m break between interviews, and then try to accommodate those preferences as much as possible (given interviewer availability)*.
```

I think it's important to be explicit that is a preference they can express not just "all-on-one-day or spread-across-multiple-days

nicksnyder: Yeah not a binary choice. I tweaked your wording a bit.

```suggestion
We are flexible when it comes to scheduling interviews because we are all-remote and we don't need to schedule all interviews back-to-back (unlike typical onsite interviews at other companies). Some candidates like spreading interviews out (e.g., across multiple days and/or having breaks between interviews on the same day) and others prefer to batch them as much as possible to get it over with. *Ask the candidate what their ideal interview schedule looks like and then try to accommodate those preferences as much as possible (given interviewer availability)*.
```

slimsag: great!

PR #1164, opened 2020-07-07
Creating roadmap handbook page

christinaforney:
image

### DIFF
--- a/_resources/assets/scripts.js
+++ b/_resources/assets/scripts.js
@@ -1,0 +1,37 @@
+// Load and render Mermaid diagrams.
+window.addEventListener("load", () => {
+    if (document.querySelector('pre.mermaid')) {
+        const element = document.createElement('script')
+        element.src = 'https://unpkg.com/mermaid@8.0.0-alpha.6/dist/mermaid.min.js'
+        document.head.appendChild(element)
+        element.addEventListener("load", function() {
+            mermaid.mermaidAPI.initialize({
+                startOnLoad: false,
+                flowchart: {
+                    useMaxWidth: false,
+                    htmlLabels: true,
+                }
+            })
+            const elements = document.querySelectorAll('pre.mermaid')
+            for (const [i, element] of elements.entries()) {
+                const id = 'mermaid-' + i
+                mermaid.mermaidAPI.render(id, element.innerText, function(svg) {
+                    const overrideWidth = element.dataset.renderedWidth
+                    const scrollRight = element.dataset.scrollRight
+                    
+                    const container = document.createElement('div')
+                    container.innerHTML = svg
+                    element.replaceWith(container);
+                    container.firstChild.style.height = "auto"
+                    if (overrideWidth) {
+                        container.style.overflow = "auto"
+                        container.firstChild.style.width = overrideWidth
+                    }
+                    if (scrollRight) {
+                        container.scrollTo((container.scrollWidth - container.clientWidth) * parseInt(scrollRight)/100, 0)
+                    }
+                })
+            }
+        })
+    }
+})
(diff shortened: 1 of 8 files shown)


sqs:
@christinaforney I fixed that problem and pushed to this branch. Check it out! Screenshot mostly the same as what you posted, but it works upon the initial page load.

This is awesome!

image

sqs:
@christinaforney Let me know when you're happy about the structure of this page, and I will go and add/update campaigns priorities, then work with the rest of the campaigns team to get the roadmap down (since we're saying the roadmap is becoming an eng team responsibility). That will be how I can help beta-test this new way of doing things.

handbook/product/roadmap.md
```diff
@@ -0,0 +1,146 @@
...
+- Support for non-Git VCS :2020-07-08, 0d
+- Streaming search support :2020-06-22, 45d
+- RFC 162 - Devolve the site admin role :2020-09-01, 0d
+- RFC 52 - Other security concerns :2020-08-01, 0d
+- Spike - investigate Gitaly as a replacement for gitserver :2020-08-01, 2d
+- RFC 161 - Authentication management :2020-08-01, 0d
+- Search personal public repos, pinned repos, starred :2020-08-10, 0d
+- API rate limiting :2020-07-15, 21d
+- Rebalancing support for gitserver :2020-08-10, 0d
+- Repo metadata indexing :2020-07-27, 14d
```
tsenart, line 80: This has been removed.

```suggestion
```

tsenart:
Apart from that, so cool to have this page and visualization!

_resources/assets/scripts.js
```diff
@@ -0,0 +1,37 @@
+// Load and render Mermaid diagrams.
+window.addEventListener("load", () => {
+    if (document.querySelector('pre.mermaid')) {
+        const element = document.createElement('script')
+        element.src = 'https://unpkg.com/mermaid@8.0.0-alpha.6/dist/mermaid.min.js'
+        document.head.appendChild(element)
+        element.addEventListener("load", function() {
+            mermaid.mermaidAPI.initialize({
+                startOnLoad: false,
```
felixfbecker, line 9: I love the idea of defining a Gantt chart with mermaid, but the usage docs example looks a lot simpler than this: https://mermaid-js.github.io/mermaid/#/usage?id=include-mermaid-on-your-web-page
Could you explain what all that extra code is for?

christinaforney: @sqs to confirm, but the solution in the docs was resulting in the height and width of the injected SVG being set to 0, and we were needing to do some manual tweaks to make it show.

sqs: @felixfbecker:

1. That window.mermaid.init API is deprecated. The docs say: "Warning This type of integration is deprecated. Instead the preferred way of handling more complex integration is to use the mermaidAPI instead."
1. We only want to load the JavaScript if Mermaid is used on the page.
1. We want to let people customize the size with `data-rendered-width` (so it scrolls to the right if desired, instead of being tiny to fit the width) and `data-scroll-right` (so it's not scrolled all the way to the beginning of time).

felixfbecker: Could you add that as code comments? 🙂

christinaforney: Yes, I'll add in these comments!

christinaforney: @felixfbecker - I updated the file with comments as you requested. Let me know if I need to add any other clarifications!

christinaforney:
I would like to merge this PR as is (with TODO items in place for each team). This will allow the PMs and teams to own and iterate on their individual sections and moves this PR forward (and removes references to our old roadmap).

macraig (approved):

christinaforney:
I am going to go ahead and merge this so that we can start to iterate on each section. Please let me know if you have any questions/concerns/feedback!

Joelkw (commented):
Awesome, thanks for setting this up!

PR #1165, opened 2020-07-07
Add #customer-updates to channels with non-standard retention

dadlerj:
Add #customer-updates to the list of channels with non-standard message retention


dadlerj:
See https://sourcegraph.slack.com/archives/CMB6K7SMN/p1594152217045600?thread_ts=1594152032.045500&cid=CMB6K7SMN for context

nicksnyder:
We don't want to try to store this information outside of Slack moving forward?

PR #1175, opened 2020-07-09
distribution: Weekly team updates

pecigonzalo:
While the document describes updates as a bi-weekly process, we are effectively doing weekly updates. Bi-weekly would be too infrequent given we would only get 2 updates on each iteration. The reasons for async and weekly are still valid from the bi-weekly ones, so I kept them.

> - Weekly updates make it easy to forget what you did at the start of the week, forget to ask for help, and make it hard for others to offer help in a timely fashion.

~I have removed this point as it was not clear to me "why" weekly updates would cause this.~

### DIFF
--- a/handbook/engineering/distribution/recurring_processes.md
+++ b/handbook/engineering/distribution/recurring_processes.md
@@ -166,8 +166,6 @@ On the Mon. following the [20th (release day)](https://about.sourcegraph.com/han
 
 These meetings are recorded (posted automatically to the #distributioneers Slack channel) so that anyone whose timezone does not permit can participate after the fact.
 
-### Bi-weekly async updates
-
 #### Wednesday update
 
 Wed, before EOD (local time): distribution members are expected to post an update in Slack communicating:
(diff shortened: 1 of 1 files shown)


bobheadxi:
> I have removed this point as it was not clear to me "why" weekly updates would cause this.

I believe this means "by friday I might not remember what I did on monday" (ie on wednesday you are more likely to reflect on what you did on monday and ask for help early)

bobheadxi:
I personally quite like having to reflect a bit mid-week and read some updates from everyone 🤔 I would advocate to keep our current bi-weekly format (light one on wednesday, proper tracking issue writeup on friday)

pecigonzalo:
@bobheadxi we are still having 2 updates per week. Maybe I'm confused here, but as I understand it, bi-weekly means it happens every 2 weeks, not that we have 2 updates per week.

pecigonzalo:
I just saw your update, I was not aware that `bi-weekly` could also be used as "twice a week".  I would prefer to rephrase it to make it more clear then.

pecigonzalo:
I have updated the docs adding the section back, but rewording. Bi-weekly seems to be a more confusing term than "twice a week", let me know if its more clear now.

Thanks @bobheadxi

handbook/engineering/distribution/recurring_processes.md
```diff
@@ -188,7 +188,7 @@ Wed, before EOD (local time): distribution members are expected to post an updat
 
 The goal of this update is to ensure we're discussing things as a team, asking for help when appropriate, reflecting on our progress, and giving others the opportunity to provide help and guidance.
 
-#### Friday update
+#### Friday Github update
```
bobheadxi, line 191:
```suggestion
#### Friday GitHub update
```

PR #1177, opened 2020-07-10
Goals (not OKRs): continuous, fewer in number, less strict format

sqs:
See the `company/goals/index.md` "Why not just use OKRs?" section (in the **Files changed** tab) for the reasons for this change.


company/okrs/index.md
```diff
@@ -1,38 +1,56 @@
...
-## What are OKRs?
+We believe:
 
-[Objectives and Key Results (OKRs)](https://en.wikipedia.org/wiki/OKR) is a specific way of setting goals that we use at Sourcegraph.
+- "Plans are useless, but planning is essential."
+- "No plan survives first contact with the customer."
+- "A plan is only useful for determining the immediate next thing to do. Beyond that, you must be ready to abandon the plan in the face of new evidence."
+- Discarding a goal that turns out to be wrong is better than following through on the wrong goal. People know this intellectually but disregard it in practice.
+- Each problem we're solving has a different (and often unknown/unknowable) timeline and rate of progress.
+- Only people with skin in the game can set useful goals.
```
mrnugget, line 21: I love, love, love this! (And the whole "Philosophy" section)

nicksnyder:
> One addition that may address these objections is having some sort of regular reflection process or template that people write up to reflect on and report on progress. It could be a document they write every quarter that uses their goals commit log as a starting point and explains to themselves, their manager, and the rest of the team the evolution of their goals and accomplishments. I could provide a draft if you like this concept.

I like this idea and I think it would be really cool to be able to go to a handbook page for a team and see "what did this team do  last quarter/year"

AlicjaSuska:
Thank you for sharing those changes @sqs!

I've seen that the 'Structure and format' section is still in 'to do' so I hope my feedback will be helpful :)

As we are planning to have all short-, medium- and long-term goals, I think it would be useful to define sub-goals or steps to take for each goal. As for medium- and long-term goals we may not know all the steps from the beginning, I've come up with the following proposition:

Criteria for defining goals:

- Short-term goal - all steps should be defined
- Medium-term goal - define 3-4 steps that need to be taken first
- Long-term goal - define 1-2 initial steps

For mid- and long-term goals, plans can change a lot. By adding the first steps at the beginning, we define the most immediate actions that should be taken. By leaving the possibility to add more steps as time goes by, we make continuous planning easier.

Please, let me know what do you think about this approach.

sqs:
@AlicjaSuska That sounds like a great idea. As you write down your goals, please try it, and share what you're doing with other people to see if that works for them as well. If so, then let's suggest more people do it that way (you can open up another PR to this handbook page to document it).

sqs:
Ready for review. I've tagged a few people for review who have expressed interest in seeing this. I would love feedback from anyone on the team.

Assuming the general idea is good, I think we should merge this before we have every last detail figured out (such as how each team will write up and share their goals). It will help to have this merged and available in the handbook for everyone to consult as each team does that (and figures out what's effective and what's not).

As for the actual goals, they are in the gdoc at https://docs.google.com/document/d/1Z6avFUmnGW-ZC6zrktqqQd_g5mhBE96C_mTwXwFc1H4/edit#, and we'll keep them in the gdoc until we have a good first pass, at which point I'll move them to the handbook (again, in a later PR).

handbook/product/planning.md
```diff
@@ -6,9 +6,9 @@ Planning is a continuous process of negotiation between product and engineering.
 
 Planning requires several artifacts for communicating. This section clarifies how each artifact is used in the planning process.
 
-### OKRs
+### OKGoals
```
pecigonzalo, line 9:
```suggestion
### Goals
```

sqs: Haha, good find. Reminds me of https://en.wikipedia.org/wiki/OK_Go.

company/goals/index.md
```diff
@@ -0,0 +1,72 @@
...
+   - No single update/review cadence (such as quarterly) makes sense for all goals.
+1. Change your goal immediately if/when you discover it's the wrong goal.
+   - Reflect on why you needed to change the goal (so you can learn), but don't keep working toward the wrong goal.
+
+## Choosing goals
+
+A goal's format is:
+
+> **Goal title** \
+> Description of the goal and how to evaluate whether we met the goal (with a link to an automatically updated metrics dashboard, if appropriate).
```
christinaforney, line 35: In support of the first item below, should the person responsible be listed by the goal? 

```suggestion
> **Goal title** \
> Owner: the name of the single person who is responsible. 
> Description of the goal and how to evaluate whether we met the goal (with a link to an automatically updated metrics dashboard, if appropriate).
```

company/goals/index.md
```diff
@@ -0,0 +1,72 @@
...
+
+1. Each goal has a single person (not multiple people) who's ultimately responsible for it.
+   - Many people can be working toward a goal, but there must be one person who's ultimately responsible.
+1. No person can have more than 7 goals at once. (This number is arbitrary, but it feels right.) Fewer is better.
```
christinaforney, line 39: I think 7 is a lot, I would feel more comfortable with 5 as the max. This also aligns with our desire to be more focused as an organization because it makes us choose what is most important.

```suggestion
1. No person can have more than 5 goals at once. (This number is arbitrary, but it feels right.) Fewer is better.
```

company/goals/index.md
```diff
@@ -0,0 +1,72 @@
...
+1. Pick goals where failure would be painful.
+1. Pick goals where you can influence the outcome. Avoid using a lagging indicator as a goal.
+1. For a new project, it can be hard to define the right goal and measurement. Do your best. Then refine the goal as you learn more.
+1. Set goals to be ambitious but achievable, so that they convey your aims and what you could achieve if things go well.
+   - In (rough) numeric terms, if you achieve less than 70% of your goal, it was probably not achievable. If you regularly achieve 100%+ of your goals, they are probably not ambitious enough.
```
christinaforney, line 44: Since these goals are continuous, I would expect that we're working towards 100% on each goal (presuming this goal is still important and has not been invalidated). 

Another concern/question - it feels likely that we will start to reach 70% on a goal, declare it good enough and move on to a new goal. Perhaps that is ok, but would indicate that goals are too hard, when in fact we just chose "good enough" and the true goal was actually appropriate.

company/goals/index.md
sqs: I'm not sure. In most cases, all of the goals in a given list of goals will have the same owner (I think). So it might be clearer to state that in the section header. Let's see how it goes when we start to compile the goals.

company/goals/index.md
sqs: Yes, good point. I removed this because the 70% is no longer relevant.

PR #1188, opened 2020-07-12
Add Plugins and  Project Tools

AlicjaSuska:
Let's add Red Lines and Contrast Checker plugins to the onboarding process. They are very useful. 

In addition, new designers should get familiar with the Project Tools document to be able to document their Figma files properly.


rrhyne:
> Let's add Red Lines and Contrast Checker plugins to the onboarding process. They are very useful.

I wasn't super happy with redlines on the CTA project. I used a very simple component instead. Let's use it a bit more and make sure it fits on the list. 

Contrast checker seems to be working out well though, thanks for adding.

PR #1190, opened 2020-07-13
TODO remove "not a platform"

sqs:


sqs:
already done in another commit

PR #1208, opened 2020-07-14
edit Series B-1 blog post

sqs:


sqs:
@adaaaam @adamfrankl2015 Here is an updated blog post. Can one of you take it from here? I wanted to be more specific about what we're doing with the money, and mention the pandemic's impact in a really conversational way.

PR #1218, opened 2020-07-15
Refresh CE team index and add CE to org chart

dadlerj:


company/team/org_chart.md
```diff
@@ -18,6 +18,8 @@ The org chart is generated automatically. [Need to edit it?](#how-to-edit)
 ### [Engineering: Security team](../../handbook/engineering/security/index.md#members)
 ### [Engineering: Web team](../../handbook/engineering/web/index.md#members)
 
+## [Customer Engineering](../../handbook/ce/index.md#members)
```
sqs, line 21: Oh, this is a limitation I just realized and need to fix. Only 3rd-level headings will work here. I wasn’t thinking of teams with no sub-teams. So just make the 2nd-level one just text and add a 3rd-level heading right underneath it that also says customer engineering and links to this section.

sqs: I just fixed it. So you can use what you had originally.

sqs: (and I pushed a change to your branch to go back to what you had)

dadlerj: Thanks!

PR #1226, opened 2020-07-17
cloud: Add Goals section

tsenart:
As per discussion on https://threads.com/34382402992 and a live meeting
to come up with the longer term goals for the team.


keegancsmith:
I like this goal. 1 line that is easy to remember. Also clear way to guide any work you do. 👍

nicksnyder:
After reading https://github.com/sourcegraph/about/pull/1177 I think this goal is missing two things:
1. A named owner
2. A clear definition of success (i.e., it should be defined in a way that there should be no ambiguity about whether it is "done" or not).

tsenart:
> This goal is missing two things according to the documentation in #1177:

Since that PR has since been merged, I'd like to discuss those two points in the product eng sync meeting tomorrow. Merging for now.

PR #1244, opened 2020-07-22
Revert "Lh/3 18 release blogs"

LaureenH:
Reverts sourcegraph/about#1242

### DIFF
--- a/blogposts/2020/3.18/C++-precise-code.md
+++ b/blogposts/2020/3.18/C++-precise-code.md
@@ -1,14 +1,13 @@
+
 ---
 title: 'New C++ precise code intelligence solution'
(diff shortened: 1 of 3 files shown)


christinaforney:
@LaureenH - Is the reason you reverted to unpublish the blogs? Instead you could have just changed the metadata to "published: false" and it would not show up on the site.

LaureenH:
> @LaureenH - Is the reason you reverted to unpublish the blogs? Instead you could have just changed the metadata to "published: false" and it would not show up on the site.

I reverted because for some reason, every blog I published grabbed the C++ blog content, not its actual content, despite looking fine in the files.

PR #1250, opened 2020-07-22
Added security philosophy to security team

chayim:
We received excellent feedback from a candidate asking what our approach and overall security philosophy is. This is an early attempt to outline that.

### DIFF
--- a/handbook/engineering/security/index.md
+++ b/handbook/engineering/security/index.md
@@ -10,6 +10,10 @@ Make Sourcegraph secure by design.
 - #security channel Slack.
 - [@sourcegraph/security](https://github.com/orgs/sourcegraph/teams/security) on GitHub.
 
+## Philosophy
+
+At Sourcegraph we think that security is an enabler for the business. Sourcegraph is committed to proactive security, addressing vulnerabilities in a timely manner. We approach security with a can-do philosophy, and look to achieve product goals while maintaining a positive posture, and increasing our security stance over time.
+
 ## Responsibilities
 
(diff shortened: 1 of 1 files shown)


ElizabethStirling:
Looks good to me, seems to encompass what we've talked about. Might consider saying something about how we are dedicated to keeping our customers safe, but I'd defer to you @chayim on that decision.

chayim:
Sounds reasonable to me. Want to try an edit.

handbook/engineering/security/index.md
```diff
@@ -10,6 +10,10 @@ Make Sourcegraph secure by design.
 - #security channel Slack.
 - [@sourcegraph/security](https://github.com/orgs/sourcegraph/teams/security) on GitHub.
 
+## Philosophy
+
+At Sourcegraph we think that security is an enabler for the business. Sourcegraph is committed to proactive security, addressing vulnerabilities in a timely manner. We approach security with a can-do philosophy, and look to achieve product goals while maintaining a positive posture, and increasing our security stance over time.
```
nicksnyder, line 15:
```suggestion
At Sourcegraph we think that security is an enabler for the business. Sourcegraph is committed to proactive security, and addressing vulnerabilities in a timely manner. We approach security with a can-do philosophy, and look to achieve product goals while maintaining a positive posture, and increasing our security stance over time.
```

nicksnyder: This is great and I think it should be inlined immediately after the page title (not in a separate `## Philosophy` section) and we can remove `## Vision` too.

PR #1260, opened 2020-07-23
Introduce web group issue triages

felixfbecker:
We did these multiple times a while ago but haven't done one in a while. As we define our own goals for the team, long-term and for each iteration, I think it's even more important than before we all have a shared awareness of our issue backlog (which includes everything from bugs, feature requests, debt, perf problems and tooling problems) so we can make informed decisions on what goals to set and what to work on. I am therefor proposing we make this part of our process, through a PR to the handbook 🙂 

Especially interested to hear from @marekweb and @twop if this would be helpful!


marekweb:
Sounds great for maintaining awareness of older issues and to surface things that should be re-surfaced. For example revisiting certain older low-urgency issues that have a time-frame attached e.g. 3 months as was mentioned in our recent discussion.

felixfbecker:
Cool! I'm hesitant to hard-code Thursday or Friday evening prematurily in defense of CET people, but I'm happy for us to document the timeframe later when we found one that works well 🙂

PR #1294, opened 2020-07-29
Add distribution team goals

pecigonzalo:
Adds our initial team goals

### DIFF
--- a/handbook/engineering/distribution/goals.md
+++ b/handbook/engineering/distribution/goals.md
@@ -27,7 +27,8 @@ We will start by looking at our Kubernetes deployment and working on an easier u
   - Upgrading a deployment configuration requires less than 2 hours of work
 
 ### Improve the debugging and troubleshooting process
-As we deploy Sourcegraph to multiple different environments, we need to provide a consistent and straightforward process. We will initially focus on reducing the time it takes to collect troubleshooting information.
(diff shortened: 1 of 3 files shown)


handbook/engineering/distribution/goals.md
```diff
@@ -0,0 +1,41 @@
...
+We will start by looking at our Kubernetes deployment and working on an easier update process.
+
+- **Owner**: Distribution Team
+- **Status**: In Progress
+- **Outcomes**:
+  - Upgrades to deployments do not require resolving diff conflicts from upstream
+  - Upgrading a deployment configuration requires less than 2 hours of work
+
+### Improve the debugging and troubleshooting process
+As we deploy Sourcegraph to multiple dissimilar environments, we need to provide a consistent and straight forward process. We will initially focus on reducing the time it takes to collect troubleshooting information.
```
nicksnyder, line 30: A problem statement is missing here. What is wrong with the current debugging a troubleshooting process?

straight forward -> straightforward

uwedeportivo, line 30:
```suggestion
As we deploy Sourcegraph to multiple different environments, we need to provide a consistent and straightforward process. We will initially focus on reducing the time it takes to collect troubleshooting information.
```

pecigonzalo: > As we deploy Sourcegraph to multiple dissimilar environments, we need to provide a consistent and straight forward process to debug issues. We are currently lacking tools to collect debugging information (configuration, type, size, diff from upstream, etc) consistently and a process to capture the output of debugging sessions to feed back into our priorities and documentation.
We will initially focus on reducing the time it takes to collect troubleshooting information.

I think this might reflect better.

PR #1295, opened 2020-07-29
cloud: Team page updates

tsenart:
This commit updates the Cloud team page with a number of changes:

- Latest planning process
- Daily updates
- Changing weekly updates to be goal updates rather than personal ones
- Retrospectives section
- Team syncs
- Full stack engineer in Hiring section (job req TBD and discussed with
the team in the next sync)

### DIFF
--- a/handbook/engineering/cloud/index.md
+++ b/handbook/engineering/cloud/index.md
@@ -17,7 +17,59 @@ The cloud team owns all work that is necessary to build, secure, scale, and oper
 
 ## Processes
 
-We do [weekly check-ins](../tracking_issues.md#using-a-tracking-issue-for-progress-check-ins) and [monthly planning](../tracking_issues.md#planning-a-milestone-with-a-tracking-issue).
+### Planning
+
+We meet every month for a planning session as the current iteration approaches its ends. In this meeting we collaborate on what our few and focused goals for the iteration should be, what their scope is and which teammates work on what.
+These goals are then captured in an iteration [tracking issue](../tracking_issues.md).
+
+It's fine for an iteration to start with only clear goals and for the specific work to make progress on those to be discovered afterwards.
+
+### Updates
+
+We do regular updates to communicate our progress to members of the team, and to external stakeholders.
+
+#### Daily personal Slack updates
+
+Collaborating across timezones requires regular communication to keep each other updated on progress, and coordinate work handoff if needed. We use daily Slack updates to achieve this.
+
+Every day, Slackbot will post a reminder in the #cloud channel for you to write your daily update.
+
+**At the end of each working day**, you should post your update as a threaded response to the Slackbot message.
+
+You should include in your update:
+- What you worked on during your day.
+- Whether you're blocked on anything to make progress (a code review, input in an RFC or in a GitHub issue...).
+- What you plan on tackling next.
+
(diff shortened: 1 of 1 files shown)


handbook/engineering/cloud/index.md
```diff
@@ -17,7 +17,59 @@ The cloud team owns all work that is necessary to build, secure, scale, and oper
...
+
+**At the beginning of each working day**, you should read the updates thread for the previous working day, to learn what your teammates have been working on, and check if they need your help.
+
+#### Weekly goal updates in the tracking issue
+
+We use weekly [progress updates in the tracking issue](../tracking_issues.md#progress_updates) to inform external stakeholders of the progress of the team on the iteration goals.
+
+Every Friday, Slackbot will post a reminder in #cloud for us to write weekly progress updates for each goal in the iteration.
+
+The teammates working on a goal are responsible for this update which is essentially the written version of what should be communicated in the Cloud Sync meetings the next Monday.
```
nicksnyder, line 52: This makes it sound like we will talk about status in the cloud sync, but I don't think that is true. This also makes it sound like only one update per goal is necessary. Is that true? I would just delete this bit and defer to the general documentation on progress updates.

```suggestion
```

handbook/engineering/cloud/index.md
```diff
@@ -37,4 +89,4 @@ _Updated 2020-07-03_
 
 We are hiring for these roles:
 
-- +1 [Software Engineer - Backend](https://github.com/sourcegraph/careers/blob/master/job-descriptions/software-engineer-backend.md)
+- +1 Software Engineer - Full Stack (Job req TBD)
```
nicksnyder, line 92: Are we not hiring more backend engineers for this team? I think it might be valuable to change this to a more prose "Growth plan" like other team pages: https://about.sourcegraph.com/handbook/engineering/search#growth-plan, https://about.sourcegraph.com/handbook/engineering/web#growth-plan

Update date on L88.

handbook/engineering/cloud/index.md
tsenart: > This makes it sound like we will talk about status in the cloud sync, but I don't think that is true.

I think it makes sense to *briefly* give goal updates to everyone in the cloud sync, rather than personal updates.

handbook/engineering/cloud/index.md
tsenart: > Are we not hiring more backend engineers for this team? I think it might be valuable to change this to a more prose "Growth plan" like other team pages.

It's worth discussing with the team what growth needs we can identify currently beyond the full stack engineer role. I'll come back and convert this to a growth plan like other teams after that discussion (next cloud sync).

tsenart, line 92:
```suggestion
- +1 [Software Engineer - Backend](https://github.com/sourcegraph/careers/blob/master/job-descriptions/software-engineer-backend.md)
```

handbook/engineering/cloud/index.md
nicksnyder: > I think it makes sense to briefly give goal updates to everyone in the cloud sync, rather than personal updates.

That makes sense to me, but I think that would be better documented in the team sync section rather than here. As it is, it feels like this definition is incorrectly inverted (it makes it sound like the written update should be based on what you want to say in the weekly meeting rather than the other way around)

PR #1297, opened 2020-07-29
Document Terraform state styleguide

pecigonzalo:
This will add information about our standard for terraform state configuration.

### DIFF
--- a/handbook/engineering/languages/terraform.md
+++ b/handbook/engineering/languages/terraform.md
@@ -2,6 +2,36 @@
 
 - General Terraform [styleguide](https://www.terraform.io/docs/configuration/style.html)
 
+## State
+
+State must be stored using a [GCS Terraform state backend](https://www.terraform.io/docs/backends/types/gcs.html).
+
+Example configuration
+```
+terraform {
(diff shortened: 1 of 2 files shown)


handbook/engineering/languages/terraform.md
```diff
@@ -2,6 +2,36 @@
...
+  backend "gcs" {
+    bucket = "sourcegraph-tfstate"
+    prefix = "infrastructure/dns"
+  }
+}
+```
+
+### State for state buckets
+
+Because we need to create state buckets as code, we also need to store the state of the code that creates the state bucket. Given this code rarely changes and that moving it to be stored in a remote location creates a chicken and egg situation, we will store state bucket creation's state in Git.
```
uwedeportivo, line 23: turtles all the way down :-)

nicksnyder:
Can you update CODEOWNERS so distribution team owns relevant docs like this?

handbook/engineering/languages/terraform.md
pecigonzalo: I have that on a shirt :D

PR #1299, opened 2020-07-29
Add training sessions to CE handbook

dadlerj:


dadlerj:
Thanks for the quick review @christinelovett !

PR #1306, opened 2020-07-30
Add docs for class->function component conversion

lguychard:


handbook/engineering/languages/typescript.md
```diff
@@ -143,5 +143,196 @@ These are usually where mutation is unavoidable.
...
+
+```typescript
+useEffect(() => {
+    eventLogger.logViewEvent('ApiConsole')
+}, [])
+```
+
+### Fetching data in componentDidMount
+
+A lot of components will fetch initial data using observables, calling `setState()` in the `susbcribe()` callback. See [this example from `SiteUsageExploreSection`](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@9997438a8ba2fdc54920f1f6ad22dd08d4a37215/-/blob/web/src/usageStatistics/explore/SiteUsageExploreSection.tsx?subtree=true#L32-38):
```
eseliger, line 163:
```suggestion
A lot of components will fetch initial data using observables, calling `setState()` in the `subscribe()` callback. See [this example from `SiteUsageExploreSection`](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@9997438a8ba2fdc54920f1f6ad22dd08d4a37215/-/blob/web/src/usageStatistics/explore/SiteUsageExploreSection.tsx?subtree=true#L32-38):
```

PR #1308, opened 2020-07-30
not-used-useless value feedback

sqs:
This value is not actually lived by the team. It is aspirational. We will propose adding it back once we feel we are living it more or are ready to do so.

### DIFF
--- a/company/values.md
+++ b/company/values.md
@@ -27,8 +27,3 @@ We prefer to [build teams, not workgroups](https://dzone.com/articles/workgroups
 
 As a team's size and scope grows, it (unintentionally) tends to become a workgroup. Resist this by aggressively focusing, or by splitting it into smaller teams.
 
-## If it's not used, it's useless
-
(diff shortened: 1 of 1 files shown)


adaaaam:
"If it's not used, it's useless" seems more relevant to product and engineering teams. With marketing, we often ideate and try new things. Sometimes those things don't work and we need to iterate based on feedback and other learnings. But it is important for us to keep that "useless" plan to then improve the next plan. The term "useless" is quite strong diction and invalidates the learning process from running experiments. The notion of how "useless" something is could also be argued in terms of how many people (including both users and colleagues) use it. 

For me, language like "Prioritize and pivot" applies to more teams' work, and aligns with our new [Goals](https://about.sourcegraph.com/company/goals) process. We discuss and agree on goals with our colleagues and leadership. We are encouraged to recognize if that goal is the wrong goal, so we can abandon it and start working on a new goal. Measuring the usefulness of a goal in terms of delivering value (both to our users and colleagues) should be part of that continuous review process.

sqs:
This value is not actually lived by the team. It is aspirational. We will propose adding it back once we feel we are living it more or are ready to do so.

mrnugget:
I understand why you removed it, but just for the record: I really liked it. I think it's a nice reminder to have a "sense of urgency" (which is something I've heard a lot) and a good counterbalance to an engineering-driven roadmap, because we all know that engineers are easily tempted to build things that are fun to build and cool to have, but not necessarily something a customer asked for or will use.

sqs:
@mrnugget Yeah, thanks for sharing this! I do think it is an important thing to be mindful of, and I hope we can feel like we're all living this value enough to put it in the handbook.

PR #1314, opened 2020-07-31
cloud: document manual migrations we're performing

slimsag:
Historically Sourcegraph.com has not been a priority for us, and as such we do not have Sourcegraph.com-only migrations today. See https://github.com/sourcegraph/sourcegraph/issues/12590

This means that today there have been a number of migrations, indexes, etc. that have been applied by hand against the Sourcegraph.com production DB. Some of these have been captured in the [Ops incidents log](https://docs.google.com/document/d/1dtrOHs5STJYKvyjigL1kMm6u-W0mlyRSyVxPfKIOfEw/edit) but I fear the vast majority have not.

Recently, we saw another one performed and [have a need to perform one that gets us away from this prod-DB-tech-debt issue further](https://github.com/sourcegraph/sourcegraph/pull/12561).

Until we have a [better process in place](https://github.com/sourcegraph/sourcegraph/issues/12590), this document will serve as a record of which manual migrations were ran against prod, when, why, and how.

### DIFF
--- a/handbook/engineering/cloud/manual_migrations.md
+++ b/handbook/engineering/cloud/manual_migrations.md
@@ -28,5 +28,5 @@ https://github.com/sourcegraph/sourcegraph/pull/12591
 Exact SQL:
 
 ```sql
-TODO(beyang,rijnard)
+CREATE INDEX CONCURRENTLY repo_name_idx ON public.repo USING btree (lower(name::text) COLLATE pg_catalog."C");
 ```
--- a/handbook/engineering/cloud/manual_migrations.md
+++ b/handbook/engineering/cloud/manual_migrations.md
@@ -1,0 +1,32 @@
+# Cloud manual migrations
+
+Sourcegraph.com occassionally requires migrations be ran against it that are not ran against other deployments of Sourcegraph.
+
+Examples of this include:
+
+- Modifying DB users, which may not always be possible in customer environments https://github.com/sourcegraph/sourcegraph/pull/12561
+- One-off index modification tests like https://twitter.com/beyang/status/1287826896281989120 that _should_ eventually become migrations in our product.
+
+We soon want to eliminate the need for such manual migrations against prod ([tracking issue](https://github.com/sourcegraph/sourcegraph/issues/12590)), but in the meantime this document serves as a list of manual migrations that have been ran against production.
+
+## Aug 3rd, 2020 - read only user addition
+
+https://github.com/sourcegraph/sourcegraph/pull/12561
+
+Exact SQL:
+
+```sql
+TODO(efritz,chayim)
+```
+
+## July 27th, 2020 - index addition
+
+https://twitter.com/beyang/status/1287826896281989120
+
+https://github.com/sourcegraph/sourcegraph/pull/12591
+
+Exact SQL:
+
+```sql
+TODO(beyang,rijnard)
+```
--- a/handbook/engineering/cloud/manual_migrations.md
+++ b/handbook/engineering/cloud/manual_migrations.md
@@ -19,7 +19,11 @@ Exact SQL:
 TODO(efritz,chayim)
 ```
 
+**Impact if this migration is undone:** Sourcegraph personnel would need to use the write-only `sg` user, does not impact product or performance behavior.
+
 ## July 27th, 2020 - index addition
+
+https://gist.github.com/rvantonder/7745c2360cfced6f12c922588bfbef79
 
 https://twitter.com/beyang/status/1287826896281989120
 
@@ -30,3 +34,5 @@ Exact SQL:
 ```sql
 CREATE INDEX CONCURRENTLY repo_name_idx ON public.repo USING btree (lower(name::text) COLLATE pg_catalog."C");
 ```
+
+**Impact if this migration is undone:** Degraded performance when searching a lot of repos with a common prefix, no impact otherwise.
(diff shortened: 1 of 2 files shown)


handbook/engineering/cloud/manual_migrations.md
```diff
@@ -0,0 +1,32 @@
...
+## July 27th, 2020 - index addition
+
+https://twitter.com/beyang/status/1287826896281989120
+
+https://github.com/sourcegraph/sourcegraph/pull/12591
+
+Exact SQL:
+
+```sql
+TODO(beyang,rijnard)
```
rvantonder, line 31:
```suggestion
CREATE INDEX CONCURRENTLY repo_name_idx ON public.repo USING btree (lower(name::text) COLLATE pg_catalog."C");
```

If we're happy going with that as in https://github.com/sourcegraph/sourcegraph/pull/12591.

handbook/engineering/cloud/manual_migrations.md
```diff
@@ -0,0 +1,32 @@
...
+
+Exact SQL:
+
+```sql
+TODO(efritz,chayim)
+```
+
+## July 27th, 2020 - index addition
+
+https://twitter.com/beyang/status/1287826896281989120
```
rvantonder, line 24: Problem/Diagnosis in https://gist.github.com/rvantonder/7745c2360cfced6f12c922588bfbef79. We could copy somewhere or just link to it.

handbook/engineering/cloud/manual_migrations.md
```diff
@@ -0,0 +1,32 @@
...
+
+https://github.com/sourcegraph/sourcegraph/pull/12561
+
+Exact SQL:
+
+```sql
+TODO(efritz,chayim)
+```
+
+## July 27th, 2020 - index addition
```
rvantonder, line 22: One thing that would be nice to capture in this document, maybe, is the motivation/ to what extent a command/migration is reversible or removable (say, for debugging). For example, one way to convey this for the index change is something like "if we removed this index, you'll get degraded performance when searching a lot of repos with a common prefix but other than that things will still work"

For the above read-only change, something like "we don't want to leave ourselves open to accidental changes. FOO service relies on `sgreader` so if you remove it then BAR will break" (or whatever succinct description we can give it).

slimsag: Great, makes sense!

tsenart:
There should be no cloud only migrations. Is there a good reason to not add these migrations as normal migrations?

pecigonzalo:
@tsenart I dont think there should be for indexes, unless we are testing something and even then, they 
could be done in code and feature flagged or reverted.

There are another cases that are being referenced here, which are non-application related tasks, as the RO user creation. Those types of changes or settings should not be part of the application migration, as they are not part of the application, they are more on the side of provisioning and deployment and are not relevant to all environments and deployments.


Lets say we would like to require one or multiple read-only users, we should not impose how those are created as different environments will deploy and configure their databases differently, some might not even grant the Sourcegraph migration script the permissions to create users.
As an administrator, I should be able to choose to use my own database and administer it following my internal requirements and guidelines as long as it meets the requirements by the Sourcegraph service.

chayim:
@tsenart It's much as @pecigonzalo mentioned.  In conversation it came out that we have 3-4 customers we support who don't have access to their underlying Database Management System (DBMS).  This means that if we do this as a pure migration (my preference) we'd break environments, with migrations that have to manually recorded.  

However, I also agree that in general a read-only connection to the database would be great. I'd generally like to encourage Principle of Least Principle.

slimsag:
I am merging this as I have not heard any strong opposition or further feedback.

This PR is merely documenting the current state of affairs (which is, yes - I agree, bad.) The fact is we do have cloud-only migrations today but we're only recording them in Slack or tweets - which is what this PR fixes.

I would love for us to have no manual migrations like this and have this document be removed, for that we'll need to solve https://github.com/sourcegraph/sourcegraph/issues/12590 or discuss further elsewhere.

tsenart:
Thank you all for clarifying.

> Lets say we would like to require one or multiple read-only users, we should not impose how those are created as different environments will deploy and configure their databases differently, some might not even grant the Sourcegraph migration script the permissions to create users.

This depends on what requirements we define for supporting [external databases](https://docs.sourcegraph.com/admin/external_database#using-your-own-postgresql-server). It would certainly simplify things for us if we didn't need a second out of band mechanism for administrative migrations. What would prevent us from requiring customers to connect Sourcegraph to an external Postgres database that is entirely owned by Sourcegraph, and thus, would not encourage admins to fiddle with it manually? In that case, could we not use a single migration mechanism for all kinds of migrations?

pecigonzalo:
I think there are several reasons to avoid supporting that in the service. Security wise, if the service needs to create the user, it means the service has admin permissions to create users and other functionality, as it requires them for migration, which is in most cases something undesired.

Aside from that, while you can perform any action in a migration, as it can execute any SQL command, I dont think this is a migration, as the scope of change its outside of its own database. Migrations traditionally manage the schema of a database, but not the database itself or any other part of the system. I would say this is provisioning and configurations which, in my opinion, is not in the scope of the service. 
Allowing admins to provision and manage their system using their own tools favors composability and reduces the amount permutations we have to account for when validating we can provision a database because we dont allow customers to do it on their own. We already have customers with restricted environments which will most likely hit this issue.

Just to clarify, we already require admins to provision a database for Sourcegraph, performing this migrations will require them to have a dedicated database server, not just a database for Sourcegraph.

tsenart:
> I would say this is provisioning and configurations which, in my opinion, is not in the scope of the service.

Fair enough, I agree with this after noodling on your answer. For the record, I was not suggesting the service itself would need to have these administrative privileges, but that instead of continuing to do migrations like we do today (done by the service itself), we could switch to another mechanism that serves all needs. Thanks for elaborating 👍 🙇

PR #1328, opened 2020-08-03
Fix line break and remove extra colon

unknwon:


unknwon:
Merging as-is, happy to address any post-merge comments!

mercadon:
Thanks @unknwon :)

PR #1350, opened 2020-08-04
Update index.md

greggstone:
Updating example for FY definition to accurately reflect FY21 start and end dates.

### DIFF
--- a/handbook/communication/index.md
+++ b/handbook/communication/index.md
@@ -111,8 +111,8 @@ Most meetings at Sourcegraph are video calls. We prefer [Zoom](https://zoom.us) 
 
 1. Always use [ISO dates](https://en.wikipedia.org/wiki/ISO_8601#Calendar_dates) in all writing and legal documents because other formats [lead to online confusion](http://xkcd.com/1179/). Use `yyyy-mm-dd`, for example 2020-04-13, and never 04-13-2020, 13-04-2020, 2020/04/13, nor April 13, 2020. Even if you use an unambiguous alternative format, it is still harder to search for a date, sort on a date, and for other team members to know we use the ISO standard. For months use `yyyy-mm`, so 2020-01 for January 2020.
 1. Prefix the year with `FY` (e.g., `FY20` for fiscal year 2020) when referring to fiscal year, which starts on February 1 and ends January 31. {#fiscal-year} We use this instead of calendar years (`CY`) because Q4 ending during the December holidays makes the quarters less even and predictable.
-  - FY__ is fiscal year 20__ (e.g., `FY20` is fiscal year 2020, which means 2020-02-01 through 2021-01-31)
-  - FY__-Q1 (e.g., `FY20-Q1`) is February 1 through April 30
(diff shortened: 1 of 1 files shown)


handbook/communication/index.md
```diff
@@ -111,8 +111,8 @@ Most meetings at Sourcegraph are video calls. We prefer [Zoom](https://zoom.us)
 
 1. Always use [ISO dates](https://en.wikipedia.org/wiki/ISO_8601#Calendar_dates) in all writing and legal documents because other formats [lead to online confusion](http://xkcd.com/1179/). Use `yyyy-mm-dd`, for example 2020-04-13, and never 04-13-2020, 13-04-2020, 2020/04/13, nor April 13, 2020. Even if you use an unambiguous alternative format, it is still harder to search for a date, sort on a date, and for other team members to know we use the ISO standard. For months use `yyyy-mm`, so 2020-01 for January 2020.
 1. Prefix the year with `FY` (e.g., `FY20` for fiscal year 2020) when referring to fiscal year, which starts on February 1 and ends January 31. {#fiscal-year} We use this instead of calendar years (`CY`) because Q4 ending during the December holidays makes the quarters less even and predictable.
-  - FY__ is fiscal year 20__ (e.g., `FY20` is fiscal year 2020, which means 2020-02-01 through 2021-01-31)
-  - FY__-Q1 (e.g., `FY20-Q1`) is February 1 through April 30
+  - FY__ is fiscal year 21__ (e.g., `FY21` is fiscal year 2021, which means 2020-02-01 through 2021-01-31)
```
limitedmage, line 114: Should this still be '20__'?

sqs: Oh, I think there is some weirdness here. When I did the fiscal year switch, I didn't realize that the convention was that FY*NN* means the fiscal year ending in the calendar year 20*NN*. So our fiscal year 2020 is defined to end in 2021. That is why there is confusion.

Is that convention indeed overwhelmingly common? If so, I definitely think we should adopt it to avoid confusion. We'll need to revise a bunch of spreadsheets and docs, but that won't be too hard.

sqs: @dadlerj What do you think? ^

dadlerj: Yes, this is correct—if the end of the fiscal year is in calendar year XX, then it's FYXX. So for, since our current fiscal year ends in January 2021, the current year should be referred to as fiscal year 2021 or FY21. 

Find more at https://www.investopedia.com/terms/f/fiscalyear.asp:

>Fiscal years are referenced by their end date or end year. For example, to reference a nonprofit organization's fiscal year-end, you may say, "FY 2020" or "fiscal year ending Jun 30, 2020." Similarly, if you referred to government spending that occurred on Nov 15, 2019, you would label that as an expenditure for the fiscal year 2020.

(FYI good timing on raising this, I went through the board deck last night and updated everything to use this convention)

dadlerj: Just googled examples of companies that have FYs ending January, and looked up an example. Find Target's 10-K using this wording convention as an example: https://investors.target.com/node/28571/html.

dadlerj:
Replaced by https://github.com/sourcegraph/about/pull/1981/files

PR #1362, opened 2020-08-05
front page edits

chrispine:
removed duplicate listing from Chrome; removed "$EDITOR" from JetBrains listing (wasn't sure about that last one, so would like a review)


chrispine:
If the $EDITOR was meant to be a cute way of saying "all JetBrains editors", I am happy to revert that commit and just remove the duplicate Chrome listing. As it was, I wasn't sure if it was intentional or not.

sqs:
@chrispine Thank you for pushing this!

> If the $EDITOR was meant to be a cute way of saying "all JetBrains editors", I am happy to revert that commit and just remove the duplicate Chrome listing. As it was, I wasn't sure if it was intentional or not.

Yes, it was intentional. But it clearly didn't work on you, so that's a good data point. Unless anyone chimes in and says they got it and liked it, let's remove it.

PR #1363, opened 2020-08-05
Add handbook section for exposing-service

daxmc99:
Add a section on how to expose services

### DIFF
--- a/handbook/engineering/languages/exposing-services.md
+++ b/handbook/engineering/languages/exposing-services.md
@@ -13,7 +13,6 @@ Binding to `localhost:port` or `127.0.0.1:port` is binding to a local-only inter
 
 This may be preferred in a sidecar pattern where you do not want a container accessible outside a Pod or when you don't want expose your laptop to the cofeeshop wifi but generally, code should not be merged that binds to `localhost` or `127.0.0.1`.
 
-
 ### How can I port-forward a local only service?
 
 Due to the way that kube-proxy works when you port-forward a pod or a service kube-proxy opens a tunnel to that pod (or a pod that backs that service). This can make debugging why a service is accessible in the pod but not outside of the pod hard to [understand](https://github.com/sourcegraph/zoekt/pull/46/files).
--- a/handbook/engineering/languages/exposing-services.md
+++ b/handbook/engineering/languages/exposing-services.md
@@ -1,6 +1,6 @@
 # Exposing services
 
-In Go, that looks like this
+In Go, that looks like this:
 ```
 http.ListenAndServe(":80", nil)
 ```
--- a/handbook/engineering/languages/exposing-services.md
+++ b/handbook/engineering/languages/exposing-services.md
@@ -1,4 +1,4 @@
-### Exposing services
+# Exposing services
 
 In Go, that looks like this
 ```
--- a/handbook/engineering/languages/exposing-services.md
+++ b/handbook/engineering/languages/exposing-services.md
@@ -19,6 +19,10 @@ Due to the way that kube-proxy works when you port-forward a pod or a service ku
 
 You can also use `kubectl port-forward --address 0.0.0.0` if you need to expose a port-forwarded service outside of your local machine (it defaults to `127.0.0.1`). [link](https://github.com/kubernetes/kubernetes/issues/40053) :exploding_head:
 
+###  CI
+
+We also have CI test for this [here](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@main/-/blob/dev/check/no-localhost-guard.sh)
+
 ### References
 
 - https://stackoverflow.com/questions/20778771/what-is-the-difference-between-0-0-0-0-127-0-0-1-and-localhost
--- a/handbook/engineering/languages/exposing-services.md
+++ b/handbook/engineering/languages/exposing-services.md
@@ -1,0 +1,30 @@
+
+### Exposing services
+
+In Go, that looks like this
+```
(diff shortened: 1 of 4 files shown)


handbook/engineering/languages/exposing-services.md
```diff
@@ -0,0 +1,30 @@
...
+http.ListenAndServe(":80", nil)
+```
+The above code will bind to all TCP interfaces. Since Kubernetes does support dual-stack IPv4 & IPv6 our services should bind to all interfaces (we do not currently have IPv6 services).
+
+If you must specify an ip address to expose your service on, choosing `0.0.0.0:port` is typically a good choice. What this does can be OS and platform-specific.
+
+Binding to `localhost:port` or `127.0.0.1:port` is binding to a local-only interface that constrained to the same "host". In Kubernetes, other containers within the Pod may still communicate with this service AND you may port-forward this container and associated service
+[(Why?)](#How-can-I-port-forward-a-local-only-service?).
+
+This may be preferred in a sidecar pattern where you do not want a container accessible outside a Pod or when you don't want expose your laptop to the cofeeshop wifi but generally, code should not be merged that binds to `localhost` or `127.0.0.1`.
```
bobheadxi, line 15: Might want to point out that we have a CI check for this: https://sourcegraph.com/github.com/sourcegraph/sourcegraph@main/-/blob/dev/check/no-localhost-guard.sh?subtree=true

handbook/engineering/languages/exposing-services.md
```diff
@@ -0,0 +1,30 @@
+
+### Exposing services
```
unknwon, line 2: qq: why not 
```suggestion
# Exposing services
```

handbook/engineering/languages/exposing-services.md
```diff
@@ -0,0 +1,30 @@
+
```
unknwon, line 1:
```suggestion
```

handbook/engineering/languages/go.md
```diff
@@ -6,6 +6,10 @@ For all things not covered in this document, defer to
 [Go Code Review Comments](https://code.google.com/p/go-wiki/wiki/CodeReviewComments)
 and [Effective Go](http://golang.org/doc/effective_go.html).
 
+We also have subsections here
```
unknwon, line 9:
```suggestion
We also have subsections here:
```

handbook/engineering/languages/exposing-services.md
```diff
@@ -0,0 +1,30 @@
...
+### Exposing services
+
+In Go, that looks like this
```
unknwon, line 4:
```suggestion
In Go, that looks like this:
```

handbook/engineering/languages/exposing-services.md
```diff
@@ -0,0 +1,30 @@
...
+The above code will bind to all TCP interfaces. Since Kubernetes does support dual-stack IPv4 & IPv6 our services should bind to all interfaces (we do not currently have IPv6 services).
+
+If you must specify an ip address to expose your service on, choosing `0.0.0.0:port` is typically a good choice. What this does can be OS and platform-specific.
+
+Binding to `localhost:port` or `127.0.0.1:port` is binding to a local-only interface that constrained to the same "host". In Kubernetes, other containers within the Pod may still communicate with this service AND you may port-forward this container and associated service
+[(Why?)](#How-can-I-port-forward-a-local-only-service?).
+
+This may be preferred in a sidecar pattern where you do not want a container accessible outside a Pod or when you don't want expose your laptop to the cofeeshop wifi but generally, code should not be merged that binds to `localhost` or `127.0.0.1`.
+
+
```
unknwon, line 17:
```suggestion
```

handbook/engineering/languages/exposing-services.md
```diff
@@ -0,0 +1,30 @@
...
+You can also use `kubectl port-forward --address 0.0.0.0` if you need to expose a port-forwarded service outside of your local machine (it defaults to `127.0.0.1`). [link](https://github.com/kubernetes/kubernetes/issues/40053) :exploding_head:
+
+### References
+
+https://stackoverflow.com/questions/20778771/what-is-the-difference-between-0-0-0-0-127-0-0-1-and-localhost
+
+https://serverfault.com/questions/21657/semantics-of-and-0-0-0-0-in-dual-stack-oses/39561#39561
+
+https://stackoverflow.com/questions/49067160/what-is-the-difference-in-listening-on-0-0-0-080-and-80
```
unknwon, line 30:
```suggestion
- https://stackoverflow.com/questions/20778771/what-is-the-difference-between-0-0-0-0-127-0-0-1-and-localhost
- https://serverfault.com/questions/21657/semantics-of-and-0-0-0-0-in-dual-stack-oses/39561#39561
- https://stackoverflow.com/questions/49067160/what-is-the-difference-in-listening-on-0-0-0-080-and-80
```

handbook/engineering/languages/exposing-services.md
daxmc99: Hmm... why did this not fail this PR then?...
https://github.com/sourcegraph/sourcegraph/pull/12758/files#r466173773
.................. 🙈

daxmc99:
😬 Thanks @unknwon, I am so used to go fmt.

handbook/engineering/languages/exposing-services.md
bobheadxi: Yeah not too sure about the context of that check in particular, but I've ran into it failing my build before and it's a related topic so 🤷

PR #1368, opened 2020-08-06
Update with new link to Dan's video with caption track

quinnkeast:


dadlerj:
Just to check, where is this file located in Drive? It's in the shared drive, right, not your personal folder? If so, can you move it to the top-level "Customer engineering" folder?

quinnkeast:
> Just to check, where is this file located in Drive? It's in the shared drive, right, not your personal folder? If so, can you move it to the top-level "Customer engineering" folder?

Thanks for checking! I've relocated the file to the new destination. Link remains the same.

PR #1371, opened 2020-08-06
Update customer issues process

pecigonzalo:
I would like to ensure all customer issues are tracked and created in GitHub. Right now, some issues are created, some are in other tracking systems, which makes it difficult to analyze the types of issues we have, or how many there are, etc.

As most of our workflow is already in GitHub, I think it will be a good idea to continue to do it here, as we can benefit from adding issues to labels, projects and milestones, as well as cross-link them to PRs or other issues.
We could, for example, use labels to categorize issues.

The downside of using GitHub is that there is no out of the box tool to analyze the data for any MTTR or other types of report. Given those are not required at the moment, I dont think it will be a problem.

Relates to: https://github.com/sourcegraph/sourcegraph/issues/11904

### DIFF
--- a/handbook/ce/customer_issues.md
+++ b/handbook/ce/customer_issues.md
@@ -8,7 +8,13 @@ Customer support tickets should be translated to GitHub issues. In such cases, [
 new issue for the request](https://github.com/sourcegraph/customer/issues/new).
 
 Provide the appropriate context and add a label with the affected customer as `customer/$name`. Once its created, sharing it with the required [team](routing_questions.md).
-If necessary, link to the appropriate JIRA Service Desk ticket or HubSpot notes.
+If necessary, link to the appropriate JIRA Service Desk ticket or [HubSpot](#find-the-unique-company-url) notes.
+
+### General issues
+
+General issues are those that affect more users than those of a particular deployment. In such cases, create a [new issue for the request](https://github.com/sourcegraph/sourcegraph/issues/new/choose) describing it. If there was a previous [customer issue](##create-a-customer-issue), please link the issue in its description.
+
+Remove any potentially private information (e.g. individual people's names, company names, self-hosted Sourcegraph URLs, repo names, screenshots, etc.)
 
 ## Find the unique company URL
 
--- a/handbook/ce/customer_issues.md
+++ b/handbook/ce/customer_issues.md
@@ -4,7 +4,7 @@ Read [the support overview](index.md) before filing an issue.
 
 ## Create a customer issue
 
-Customer support tickets should be translated to GitHub issues. In such cases, [create a
+Customer support tickets should be translated to GitHub issues when they require an engineering team to be involved. In such cases, [create a
 new issue for the request](https://github.com/sourcegraph/customer/issues/new).
 
 Provide the appropriate context and add a label with the affected customer as `customer/$name`. Once its created, sharing it with the required [team](routing_questions.md).
--- a/handbook/ce/customer_issues.md
+++ b/handbook/ce/customer_issues.md
@@ -14,7 +14,7 @@ If necessary, link to the appropriate JIRA Service Desk ticket or [HubSpot](#fin
 
 General issues are those that affect more users than those of a particular deployment. In such cases, create a [new issue for the request](https://github.com/sourcegraph/sourcegraph/issues/new/choose) describing it. If there was a previous [customer issue](##create-a-customer-issue), please link the issue in its description.
 
-Remove any potentially private information (e.g. individual people's names, company names, self-hosted Sourcegraph URLs, repo names, screenshots, etc.)
+Remove any potentially private information from this new issue (e.g. individual people's names, company names, self-hosted Sourcegraph URLs, repo names, screenshots, etc.)
 
 ## Find the unique company URL
 
--- a/handbook/ce/filing_customer_issues.md
+++ b/handbook/ce/filing_customer_issues.md
@@ -4,21 +4,11 @@ Read [the support overview](index.md) before filing an issue.
 
 ## 1. Begin to create an issue
 
-Customer support tickets should be translated to issues when they require a non-SWAT engineering team to own resolution. In such cases, create a new tracking issue for the request, and link to the company's anonymous HubSpot URL at the top:
-
->Requested by $COMPANY_URL
-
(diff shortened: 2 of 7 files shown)


handbook/ce/customer_issues.md
```diff
@@ -0,0 +1,32 @@
...
+new issue for the request](https://github.com/sourcegraph/customer/issues/new).
+
+Provide the appropriate context and add a label with the affected customer as `customer/$name`. Once its created, sharing it with the required [team](routing_questions.md).
+If necessary, link to the appropriate JIRA Service Desk ticket or [HubSpot](#find-the-unique-company-url) notes.
+
+### General issues
+
+General issues are those that affect more users than those of a particular deployment. In such cases, create a [new issue for the request](https://github.com/sourcegraph/sourcegraph/issues/new/choose) describing it. If there was a previous [customer issue](##create-a-customer-issue), please link the issue in its description.
+
+Remove any potentially private information (e.g. individual people's names, company names, self-hosted Sourcegraph URLs, repo names, screenshots, etc.)
```
uwedeportivo, line 17: private information should be in the linked customer issue. don't remove it, place it there

tsenart:
> The downside of using GitHub is that there is no out of the box tool to analyze the data for any MTTR or other types of report.

Hold our collective beers 🍻

handbook/ce/customer_issues.md
pecigonzalo: Ill rephrase it, I meant it to ensure no private information is on the general issue filed on `sourcegrpah/sourcegraph`

handbook/ce/customer_issues.md
```diff
@@ -0,0 +1,32 @@
+# Filing customer issues
+
+Read [the support overview](index.md) before filing an issue.
+
+## Create a customer issue
+
+Customer support tickets should be translated to GitHub issues. In such cases, [create a
+new issue for the request](https://github.com/sourcegraph/customer/issues/new).
```
slimsag, line 8: I don't think it makes sense to do this for *all* customer issues. For example, a vast majority of the issues that come in are solved by the Customer Engineering team at the point of contact (Slack, Jira, etc.) and creating a GitHub issue for all of these would be a lot of work and harm to _their_ workflow.

I would rephrase this to indicate that when a customer issue requires pulling in another engineering team (distribution or otherwise), then create a GitHub issue.

pecigonzalo: Ill rephrase it to indicate when it requires other engineers. While its not the intention for all the issues they manage to get a GitHub issue associated, I think it would be a useful metric for CE to be able to analyze what should we document more or what are the frequent type of questions we get.

handbook/ce/customer_issues.md
pecigonzalo: @uwedeportivo updated, please review

handbook/ce/customer_issues.md
pecigonzalo: @slimsag updated, please review

pecigonzalo:
@slimsag I have merged this with the updated text, please let me know this does not fix your concern and ill address it on a separate PR.

handbook/ce/customer_issues.md
dadlerj: Definitely agree with @slimsag . Thanks for updating the language!

If metrics were a priority, we definitely would not be using GitHub. We'd force all tickets to be entered into JIRA (or some replacement tool like ZenDesk) and use that for issue categorization and analysis. GitHub issues may be the last place I'd ever want to analyze issue/question frequency (or calculate any metrics, honestly) 😂 . 

Beyond the changes you made, though, I strongly disagree with one more part of this approach—I think we should be _aggressively_ minimizing the use of the sourcegraph/customer repo, and instead by default posting everything to sourcegraph/sourcegraph. Use of sourcegraph/customer means everyone on the team has to be aware of (and use) multiple trackers, there's no way for new users who encounter an issue to +1 it, devs may forget to close both issues (if mirrored), leading to issue debt, and more.

We open sourced our issue tracker for a reason—I'd prefer to have an internal goal of shutting down sourcegraph/customer, or of saying that sourcegraph/customer is NEVER used for discussion of an issue, and should instead only be a repository for customer data that we link out to from the actual issue filed in sourcegraph/sourcegraph. 

How would you respond to those concerns?

handbook/ce/customer_issues.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Create a customer issue
+
+Customer support tickets should be translated to GitHub issues when they require an engineering team to be involved. In such cases, [create a
+new issue for the request](https://github.com/sourcegraph/customer/issues/new).
+
+Provide the appropriate context and add a label with the affected customer as `customer/$name`. Once its created, sharing it with the required [team](routing_questions.md).
+If necessary, link to the appropriate JIRA Service Desk ticket or [HubSpot](#find-the-unique-company-url) notes.
+
+### General issues
```
dadlerj, line 13: This section describes to me a manual process for overcoming the underlying problem, which is having multiple sources of truth for issue tracking. See my proposal above.

pecigonzalo:
I think we can work out the details here and then create a new draft PR.

handbook/ce/customer_issues.md
pecigonzalo: I believe incidents are for the most part customer specific, a customer having a problem related in many cases to its particular environment or deployment. There might be a set of general issues to be created that are related and must be linked to the incident, and those should be done in `/sourcegraph` where they can be :+1: and publicly discussed.
This works in a similar way to incident response workflows, were the immediate issue is worked from the generated alert (and in some cases automated issue) and later on we create any follow-up issues for the permanent or general fixes.

The main concerns is that right now there does not seem to be a clear funnel and a lot of this flows through a Slack thread or conversation, or linked from Jira in Slack, which makes it impossible to get an overview of the current issues or past issues we worked on over an iteration. Additionally, issues in `/sourcegraph` because we cant label or add any customer information, its quite hard to group by customer and get an overview of a customer incidents.

I think the tradeoff of having 2 issues trackers in GitHub (one for most of our issues and one for customer specific ones) for the issues that engineering has to work on is fairly minimal, and creates a simple funnel for all incidents, which at present are distributed between many places (Slack, Jira, etc).
Ideally, we would have a single issue tracker, but unfortunately that is not possible when mixing customer and public incidents, and I think additionally having 2 trackers, makes it harder to leak information by incorrectly setting a message or setting an issue to private (if we had a tracker that allowed that).

PR #1374, opened 2020-08-06
Add onboarding item: browser search engines

chrispine:
Added an item to the onboarding checklist for engineers: Add Sourcegraph as browser search engine, including one for our internal instance. (This will also ensure that new hires can access the internal instance.)


dadlerj:
We provide this functionality through all browser extension users, actually (using `src` as the search engine starter term). It actually pulls in autocomplete suggestions as well, so it's even better than using a manual browser search engine offering.

image

chrispine:
Thanks, @dadlerj! My intention was that new hires should do exactly that (that's what I linked to), and that engineers should also create the manual entry for our internal instance, so they can do either one. (I have `src` via the Chrome extension, and manually added `sg` for out internal instance.) This also raised the issue that I didn't have access to the internal instance.

It is my (possibly incorrect) understanding that most engineers use the internal instance rather than the public one.

PR #1378, opened 2020-08-09
Churn docs

ebrodymoore:
@dadlerj @juliasourceress anything else we need to add for churn?

I added the process to the sales doc while it's being decided which team will own renewals.

### DIFF
--- a/handbook/sales/index.md
+++ b/handbook/sales/index.md
@@ -47,6 +47,10 @@ Expansion [IARR](#iarr) is IARR from *existing customers* (i.e., organizations t
 
 If within a single period a new customer signs a contract which then grows in ARR before the end of the period, the total ending ARR is all considered [new IARR](#new-iarr), not expansion IARR. For example, if Acme Corp signs a $100k contract on February 3 and then the contract expands to $200k on March 5, all $200k would be considered new IARR for Q1.
 
+### Churned IARR
+
+The amount of [IARR](#iarr) that is lost when a customer discontinues using Sourcegraph.
+
 ### Booking
 
 A booking is when a customer commits to pay us money. This includes when:
--- a/handbook/sales/index.md
+++ b/handbook/sales/index.md
@@ -188,6 +188,10 @@ If a deal comes through a referral or introduction, tell [BizOps](../bizops/inde
 1. Update the ‘Closed Lost Dropdown’ property to reflect the reason. If the reason doesn’t exist in the dropdown, you can talk to [BizOps](../bizops/index.md) about adding one
 1. Expand upon the reason in the longform ‘Closed Lost Reason’ field. This supports [2019-Q4 OKR 1 v](../../company/goals/2019_q4.md) to identify the top 3 reasons potential customers don't sign.
 
+### When a customer churns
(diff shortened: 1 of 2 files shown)


handbook/sales/index.md
```diff
@@ -184,6 +188,10 @@ If a deal comes through a referral or introduction, tell [BizOps](../bizops/inde
 1. Update the ‘Closed Lost Dropdown’ property to reflect the reason. If the reason doesn’t exist in the dropdown, you can talk to [BizOps](../bizops/index.md) about adding one
 1. Expand upon the reason in the longform ‘Closed Lost Reason’ field. This supports [2019-Q4 OKR 1 v](../../company/goals/2019_q4.md) to identify the top 3 reasons potential customers don't sign.
 
+### When a customer churns
+1. Notify the [marketing team](../marketing/index.md#members) in #marketing if they customer had logo and case study rights that will need to be removed.
```
dadlerj, line 192: If I'm not mistaken, our default TOS doesn't require us to take down logos/case studies—the story as written was true when it was published, and that won't change—but it's possible that the sales team would want to or need to take them down for some other reason. Fine to leave this bullet here since the conversation should happen, but don't imply that they will always "need" to be removed.

handbook/sales/index.md
```diff
@@ -47,6 +47,10 @@ Expansion [IARR](#iarr) is IARR from *existing customers* (i.e., organizations t
 
 If within a single period a new customer signs a contract which then grows in ARR before the end of the period, the total ending ARR is all considered [new IARR](#new-iarr), not expansion IARR. For example, if Acme Corp signs a $100k contract on February 3 and then the contract expands to $200k on March 5, all $200k would be considered new IARR for Q1.
 
+### Churned IARR
+
+The amount of [IARR](#iarr) that is lost when a customer discontinues using Sourcegraph.
```
dadlerj, line 52: I'm not sure if "churned" is the right word here. In some cases, customers will simply use less of Sourcegraph in a subsequent period and pay less as a result (e.g. a smaller renewal). That is "negative expansion", but not really churn (as defined here). Would that be included? Do we need a more expansive definition? Or even a separate term at all here?

handbook/sales/index.md
```diff
@@ -184,6 +188,10 @@ If a deal comes through a referral or introduction, tell [BizOps](../bizops/inde
 1. Update the ‘Closed Lost Dropdown’ property to reflect the reason. If the reason doesn’t exist in the dropdown, you can talk to [BizOps](../bizops/index.md) about adding one
 1. Expand upon the reason in the longform ‘Closed Lost Reason’ field. This supports [2019-Q4 OKR 1 v](../../company/goals/2019_q4.md) to identify the top 3 reasons potential customers don't sign.
 
+### When a customer churns
```
dadlerj, line 191: Does this section apply if a contract size falls, or only if it goes to zero?

ebrodymoore: Goes to zero for now

ebrodymoore:
@dadlerj thanks for the comments. I'm proposing adding contraction and cancelled IARR terms to represent decreases in IARR from one period to the next. In practice these will be infrequently used now but we've had a couple cases where we need a term for this. cc @sqs 

Based on [GItLab's framework](https://about.gitlab.com/handbook/sales/#arr-analysis-framework)

PR #1380, opened 2020-08-10
Remove possible focuses

christinaforney:
We haven't been doing these focuses for several weeks, so we should remove it. These were listed as potential focuses, but some of the bulleted items had suggested monthly timing that we are not following. 

Let me know if there are suggestions to improve wording here vs. removing entirely!

### DIFF
--- a/handbook/communication/company_meeting.md
+++ b/handbook/communication/company_meeting.md
@@ -44,11 +44,6 @@ Company meeting is an effective way to:
 1. Inspiration of the week (message @dan to volunteer to present an inspiration)
 1. [#thanks](team_chat.md#thanks) summary
 1. Progress on our [goals](../../company/goals/index.md)
-1. **Possible focuses** (note: these are guidelines only, and @sqs will decide what each company meeting is in advance and will tag people with at least 2 working days' notice)
-   1. CEO: company pitch, vision, "are we on track?", open Q&A
-   1. Marketing, sales, customer engineering: what's working, what's not, biggest customer pains/wins
-   1. Product release (usually the closest meeting to the [20th](../engineering/releases/index.md#releases-are-monthly)): what will ship in the release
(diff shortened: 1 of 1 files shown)


handbook/communication/company_meeting.md
```diff
@@ -44,11 +44,6 @@ Company meeting is an effective way to:
 1. Inspiration of the week (message @dan to volunteer to present an inspiration)
 1. [#thanks](team_chat.md#thanks) summary
 1. Progress on our [goals](../../company/goals/index.md)
-1. **Possible focuses** (note: these are guidelines only, and @sqs will decide what each company meeting is in advance and will tag people with at least 2 working days' notice)
-   1. CEO: company pitch, vision, "are we on track?", open Q&A
-   1. Marketing, sales, customer engineering: what's working, what's not, biggest customer pains/wins
-   1. Product release (usually the closest meeting to the [20th](../engineering/releases/index.md#releases-are-monthly)): what will ship in the release
-   1. Product preview (usually the meeting after the [20th](../engineering/releases/index.md#releases-are-monthly)): what we're planning to build and ship in the next release
```
felixfbecker, line 51: **thought**: I would actually really like to see us do these again. The company meeting currently feels entirely sales and hiring focused, which for sure is useful to some extend for everyone to hear about regularly, but maybe not so useful that e.g. even the engineering team in Europe needs to attend a 7:30-8pm meeting every single Monday for that and nothing else. The bullet points in this deleted section on the other hand seem much more valuable to me for everyone to come together for.

sqs: I'm already working on this for next company meeting. We will highlight some specific companies and how they are using Sourcegraph. That will occur every week; i.e., the sales pipeline update will go into much more interesting/useful depth about product usage. Do you think that would improve it and make it more useful?

Also, what do you think about either of the following?

- moved it to 8am PDT (5pm CEST); or
- alternated it between the current time and Mon 8am CEST (Sun 11pm PDT)

PR #1373, opened 2020-08-06
New code of conduct

kghopson:
Added the new, condensed version of the code after resolving comments in the draft

PR #1388, opened 2020-08-12
Added dropbox and public site-admin steps

tistru:
Added steps to obtain dropbox shared folder access and Sourcegraph.com site-admin account for managing customer accounts


sqs:
The build is failing because https://github.com/sourcegraph/about/pull/1373 was merged with a failing check, and so the master branch is failing. I'll fix the master branch, then you can rebase.

PR #1399, opened 2020-08-14
cloud: Milestones

tsenart:
This commit adds an initial draft of the Cloud milestones the team has
discussed and come up with. We expect to iterate on these as we gather
external feedback.

@sourcegraph/cloud: Please suggest more changes before we mark this PR as ready for review.

### DIFF
--- a/handbook/engineering/cloud/index.md
+++ b/handbook/engineering/cloud/index.md
@@ -6,6 +6,26 @@ The cloud team owns all work that is necessary to build, secure, scale, and oper
 
 1. Any user or organization can use Sourcegraph Cloud with both the private and public code that they care about in a secure way.
 
+## Milestones
+
+1. Any user can add public code
+2. Organizations can use Sourcegraph with up to 50 repos or 100k LoC of private repositories for free
+    - We need to build a scalable and reliable service to count lines of code and integrate that with the rest of our systems (to enforce limits). This would align with how we’re thinking about billing by LoC, not number of repositories.
+    - No plaintext tokens or secrets anywhere (including gitserver .git/config remotes).
+    - All private repository contents need to be encrypted at rest.
+        - Gitserver.
+        - Searcher and code intel caches.
+        - Any other service that needs access to repository contents needs to be able to decrypt what it needs on the fly.
+3. Sign-up and on-boarding of new users is fast, obvious and painless
+    - Adding your repos with a nice UI flow that leverages the OAuth tokens from signing up with the code host, where you can select repositories easily.
+    - Permissions and repo syncing progress is clear to the user and doesn’t take too long.
+    - Progress on repo indexing is clear to the user and doesn’t take too long.
+4. Any user or organisation can use Sourcegraph with more than 100k LoC and pay for it.
+    - Abuse protection: API rate limiting, DDoS mitigation, limiting user accounts.
+    - Scalable syncing of permissions, repos, changesets.
+    - High availability, SLOs, etc.
+    - Billing and subscriptions.
+
 ## Contact
 
 - #cloud channel or @cloud-team in Slack
--- a/handbook/engineering/cloud/index.md
+++ b/handbook/engineering/cloud/index.md
@@ -9,10 +9,11 @@ The cloud team owns all work that is necessary to build, secure, scale, and oper
 ## Milestones
 
 1. Any user can add public code
-2. Organizations can use Sourcegraph with up to 50 repos or 100k LoC of private repositories for free
-    - We need to build a scalable and reliable service to count lines of code and integrate that with the rest of our systems (to enforce limits). This would align with how we’re thinking about billing by LoC, not number of repositories.
+    - Code is indexed.
+    - It's easy to search over only their code (vs all public code).
+2. Any user or organization can use Sourcegraph Cloud for free before it's GA
     - No plaintext tokens or secrets anywhere (including gitserver .git/config remotes).
-    - All private repository contents need to be encrypted at rest.
+    - All private repository content are only decryptable by Sourcegraph services.
         - Gitserver.
         - Searcher and code intel caches.
         - Any other service that needs access to repository contents needs to be able to decrypt what it needs on the fly.
@@ -20,11 +21,12 @@ The cloud team owns all work that is necessary to build, secure, scale, and oper
     - Adding your repos with a nice UI flow that leverages the OAuth tokens from signing up with the code host, where you can select repositories easily.
     - Permissions and repo syncing progress is clear to the user and doesn’t take too long.
     - Progress on repo indexing is clear to the user and doesn’t take too long.
-4. Any user or organisation can use Sourcegraph with more than 100k LoC and pay for it.
+4. Sourcegraph Cloud is Generally Available (GA)
     - Abuse protection: API rate limiting, DDoS mitigation, limiting user accounts.
     - Scalable syncing of permissions, repos, changesets.
     - High availability, SLOs, etc.
     - Billing and subscriptions.
+        - Based on size (i.e. GB) of all repositories and selected feature set.
 
 ## Contact
 
--- a/handbook/engineering/cloud/index.md
+++ b/handbook/engineering/cloud/index.md
@@ -8,7 +8,7 @@ The cloud team owns all work that is necessary to build, secure, scale, and oper
 
 ## Milestones
 
-1. Any user can add public code
+1. Any user can add public code from github.com, gitlab.com and bitbucket.org
     - Code is indexed.
     - It's easy to search over only their code (vs all public code).
 2. Any user or organization can use Sourcegraph Cloud for free before it's GA


handbook/engineering/cloud/index.md
```diff
@@ -6,6 +6,26 @@ The cloud team owns all work that is necessary to build, secure, scale, and oper
 
 1. Any user or organization can use Sourcegraph Cloud with both the private and public code that they care about in a secure way.
 
+## Milestones
+
+1. Any user can add public code
+2. Organizations can use Sourcegraph with up to 50 repos or 100k LoC of private repositories for free
+    - We need to build a scalable and reliable service to count lines of code and integrate that with the rest of our systems (to enforce limits). This would align with how we’re thinking about billing by LoC, not number of repositories.
+    - No plaintext tokens or secrets anywhere (including gitserver .git/config remotes).
+    - All private repository contents need to be encrypted at rest.
```
daxmc99, line 15: All this data is stored in Persistent Volumes which are packed by GCP Persistent Disks which are default encrypted at rest by GCP.  I think the term `encrypted-at-rest` might not be specific enough.  What about:
```suggestion
All private repository content are only decryptable by the Sourcegraph instance
```

handbook/engineering/cloud/index.md
```diff
@@ -6,6 +6,26 @@ The cloud team owns all work that is necessary to build, secure, scale, and oper
 
 1. Any user or organization can use Sourcegraph Cloud with both the private and public code that they care about in a secure way.
 
+## Milestones
+
+1. Any user can add public code
+2. Organizations can use Sourcegraph with up to 50 repos or 100k LoC of private repositories for free
```
unknwon, line 12: The "Organizations" here seems pretty vague, how do we define the scope of "an organization"? 

As we allow both users and organizations to own external services, will we have different billing plans for private repos of users and organizations (e.g. GitHub Pro vs GitHub Teams)?

We could probably say the following for now (similar to milestone 4):
```suggestion
2. Any user or organization can use Sourcegraph with up to 50 repos or 100k LoC of private repositories for free
```

handbook/engineering/cloud/index.md
```diff
@@ -6,6 +6,26 @@ The cloud team owns all work that is necessary to build, secure, scale, and oper
...
+    - We need to build a scalable and reliable service to count lines of code and integrate that with the rest of our systems (to enforce limits). This would align with how we’re thinking about billing by LoC, not number of repositories.
+    - No plaintext tokens or secrets anywhere (including gitserver .git/config remotes).
+    - All private repository contents need to be encrypted at rest.
+        - Gitserver.
+        - Searcher and code intel caches.
+        - Any other service that needs access to repository contents needs to be able to decrypt what it needs on the fly.
```
keegancsmith, line 18: Zoekt? Encrypted indexes sounds untenable. I also worry in general about how much effort this is implying and balancing that against cost of simplicity and performance. Why do we have such strict encryption goals? Is it not enough that the backing block store / FS is encrypted? What are we actually trying to achieve?

tsenart: @chayim's RFC 214 should capture the rationale for needing to encrypt repos in general, but we need to discuss what threat model we're protecting against, why, and make that explicit. Block storage encryption doesn't protect against many attack vectors (e.g. lateral movement). In your view, @keegancsmith, what are the threat models that we absolutely need to protect against, and how, and what can we get by without for now?

keegancsmith: Reading the proposal it sounds like we want to do this at the operating system level via encrypted filesystems. This sounds like something we could support in a somewhat generic way (although every service would need modification). However, it isn't clear to me the performance impact and which specific cases this protects us vs relying on GCP disk encryption + better hygiene around service permissions.

To me this sounds like it is protecting against a bad actor getting hold of our GCP account in such a way where they can't get into our cluster, but they can maybe decrypt disk backups/mount disks? If they can get into the cluster, one can only assume that if the service can read it, the attacker can read it. How much of a threat is that really? What does GCP provide to protect against that? (I assume locking down IAM will help a lot here).

Just as an aside we have lots of service which rely on mmap a blob of disk which contains repository file contents. This implies that the encryption layer will need to be at the OS level, otherwise introducing it will be far too much effort. The alternative that jumps into my head is properly authenticating service actors, ensuring services which persist data also respect service actors, and for sensitive data sharding data in a privacy sensitive manor.

keegancsmith: @chayim FYI I hadn't seen RFC 214 before. I think you need to add a comment with someones email for them to be notified. The RFC looks really good, my only point of discussion is number 6 about encrypting repo contents. Don't want to derail the rest of whats a really great plan.

handbook/engineering/cloud/index.md
keegancsmith, line 12: I think this milestone should be split up between allowing private code on Sourcegraph.com and how we decide if its free. They seem quite independent and bundling them together will make it harder to measure success if we don't deliver both.

tsenart: Agreed, as per our live discussion today. I'll change.

handbook/engineering/cloud/index.md
```diff
@@ -6,6 +6,26 @@ The cloud team owns all work that is necessary to build, secure, scale, and oper
 
 1. Any user or organization can use Sourcegraph Cloud with both the private and public code that they care about in a secure way.
 
+## Milestones
+
+1. Any user can add public code
+2. Organizations can use Sourcegraph with up to 50 repos or 100k LoC of private repositories for free
+    - We need to build a scalable and reliable service to count lines of code and integrate that with the rest of our systems (to enforce limits). This would align with how we’re thinking about billing by LoC, not number of repositories.
```
keegancsmith, line 13: This is super tricky. I know we have had an RFC about this before, and I can't remember exactly all the different things we discussed. Something that occurred to me is when measuring LoC, which branch are we measuring? What if its a repo with lots of very different branches? What if someone is naughty and has a mostly empty HEAD, but does dev on another branch. How do we capture the value of our tool which works across history and branches?

I proposed in a call now rather than measuring LoC, we can measure the size of the `.git` repos. This gets around a lot of these issues, but does have concerns around fluctating sizes due to repacking (but order of magnitudes shouldn't change) and it will measure checked in binaries/deps.

Another idea is to measure the size of our indexed search shards. That will exclude binaries and more directly is related to our costs of serving a search index. However, that is very specific to indexed search and is hard to keep consistent as we explore different technologies/products/etc.

tsenart:
@sourcegraph/cloud: Addressed feedback so far. One question: If you were to project timelines for each of these milestones, what would those be?

handbook/engineering/cloud/index.md
tsenart: Added a sub-item under "Billing and subscriptions" to measure value by size of repositories.

handbook/engineering/cloud/index.md
```diff
@@ -6,6 +6,28 @@ The cloud team owns all work that is necessary to build, secure, scale, and oper
 
 1. Any user or organization can use Sourcegraph Cloud with both the private and public code that they care about in a secure way.
 
+## Milestones
+
+1. Any user can add public code
+    - Code is indexed.
+    - It's easy to search over only their code (vs all public code).
```
tsenart, line 13: @keegancsmith, @sourcegraph/search: I think this could be a dependency on some substantial search changes. I see two ways in which we can make a user's code easier to find / prioritised:

1. Automatic `repogroups` per user / organization and making that the default search scope.
2. Ranking support in search, which would rank a user's or organization's repos higher (as well as their dependencies).

I think we probably need 1. first, then 2. What do you think? Would you would take these on?

keegancsmith: We have the concept of version contexts which can likely be extended here. Setting the default scope of a query may be more straight forward now from a code perspective. But if its the right thing to do is another question which is good for search to decide on. I think ideally we have better ranking (might be investigated after streaming).

However, from a code perspective limiting down which repos are searched based on the query is a relatively straightforward change. I think as you are implementing this milestone experimenting with that and seeing how it feels would make a lot of sense for the people who are owning this milestone. This is likely better than potentially getting blocked by search who are working on other projects (potentially).

unknwon:
> @sourcegraph/cloud: Addressed feedback so far. One question: If you were to project timelines for each of these milestones, what would those be?

This is a question that I feel can only be answered by hitting my head against the wall. :joy: Then If I have to:
1. 1-2 iterations
2. 2-3 iterations
3. 2 iterations
4. 3 iterations

Disclaimer: I am not responsible for what I've estimated in this commit.

handbook/engineering/cloud/index.md
chayim: > Zoekt? Encrypted indexes sounds untenable. I also worry in general about how much effort this is implying and balancing that against cost of simplicity and performance. Why do we have such strict encryption goals? Is it not enough that the backing block store / FS is encrypted? What are we actually trying to achieve?

Encrypting indexes is very hard - unless the indexes are in memory. Scanning encrypted in memory data is at best a pain, and starts to become that tradeoff of nice to have security versus a need. I'd argue that specific customers might **need** it, but generally it's a nice to have - similar to your stance. For my own education and thinking, @keegancsmith does Zoekt store indices in memory, or flush them back to disk?

> @chayim FYI I hadn't seen RFC 214 before. I think you need to add a comment with someones email for them to be notified. The RFC looks really good, my only point of discussion is number 6 about encrypting repo contents. Don't want to derail the rest of whats a really great plan.

Thanks for the vote of confidence - especially when it's incomplete!  We were just discussing email notification of key rollover with regards to RFC196, and this might be an opportunity to kill two birds with one service.  I know FUSE is something that appeals to you and I wanted to drag you in @keegancsmith .

> @chayim's [RFC 214](https://docs.google.com/document/d/1ajVQ11Vyi0Wz67DN15W3girJ5kyB5Zb7vk0LK6r7gbQ/edit) should capture the rationale for needing to encrypt repos in general, but we need to discuss what threat model we're protecting against, why, and make that explicit. Block storage encryption doesn't protect against many attack vectors (e.g. lateral movement). In your view, @keegancsmith, what are the threat models that we absolutely need to protect against, and how, and what can we get by without for now?

Absolutely agreed. This is the beginning of the RFC for that reason - and thank you for affirming that need.  I'll make it a point to get that into the background information area, and explicitly delegate a threat model.

handbook/engineering/cloud/index.md
chayim: > This is super tricky. I know we have had an RFC about this before, and I can't remember exactly all the different things we discussed. Something that occurred to me is when measuring LoC, which branch are we measuring? What if its a repo with lots of very different branches? What if someone is naughty and has a mostly empty HEAD, but does dev on another branch. How do we capture the value of our tool which works across history and branches?
> 
> I proposed in a call now rather than measuring LoC, we can measure the size of the `.git` repos. This gets around a lot of these issues, but does have concerns around fluctating sizes due to repacking (but order of magnitudes shouldn't change) and it will measure checked in binaries/deps.
> 
> Another idea is to measure the size of our indexed search shards. That will exclude binaries and more directly is related to our costs of serving a search index. However, that is very specific to indexed search and is hard to keep consistent as we explore different technologies/products/etc.

I think LOC is a bad measurement / tough sell, but I really like per user. In theory users should relate to lines of code over time, and it nicely avoids the multi-repo/monorepo problem whereby we have to guess the number of repositories to bill appropriately.  Similarly I'd like suggest not measuring the size of .git as it hard limits us to a specific revision control system; one can easily find Team Foundation Server (TFS), and Perforce still going reasonably strong internally.

dadlerj: This is a great discussion to have in a forum where the sales team can be involved FYI! I have some thoughts/responses to all of the questions above, but I'll wait to do it in a different venue.

keegancsmith:
> For my own education and thinking, @keegancsmith does Zoekt store
> indices in memory, or flush them back to disk?

The posting lists are mmap backed. The posting lists are actually offset
lists, so the posting lists of all trigrams can be used to reconstruct
documents. So in theory you could encrypt the posting lists, I just
worry about performance impact.

handbook/engineering/cloud/index.md
```diff
@@ -6,6 +6,28 @@ The cloud team owns all work that is necessary to build, secure, scale, and oper
 
 1. Any user or organization can use Sourcegraph Cloud with both the private and public code that they care about in a secure way.
 
+## Milestones
+
+1. Any user can add public code
```
ryanslade, line 11: I think it's worth being specific here:

```suggestion
1. Any user can add public code on Github.com, Gitlab.com and Bitbucket Cloud
```

asdine (approved):

tsenart:
There's no rush in merging this in. I'm going on holiday tomorrow, and now that internally in the Cloud team we have agreed on these initial proposed milestones, they're now ready for feedback from external stakeholders (and later iteration on that feedback).

handbook/engineering/cloud/index.md
nicksnyder: This was initially discussed in [RFC 146](https://docs.google.com/document/d/1H1yGMVAgAUwSYHJL-94Glx8JFelx8mnoEhZdS0WOhgc/edit#heading=h.trqab8y0kufp)

tsenart:
Merging! We'll update these next week with other non-cloud things that this team owns.

PR #1411, opened 2020-08-18
Add Noah to team page

Strum355:
Also re-added @efritz 's name-coach.com that was removed with someone elses adding to team page. Not sure if that was intentional or not


Strum355:
That automatically requested from more than i was expecting :fearful: sorry Quinn!

efritz (approved):

efritz:
> That automatically requested from more than i was expecting :fearful: sorry Quinn!

This happens a lot 💩

PR #1419, opened 2020-08-20
Add near-term milestones to Campaigns team Goals section

chrispine:
This adds the Campaigns team's near-term milestones to the Goals section on their team page. Our longer-term goals still need some work, but we expect those to come into focus over the coming 4-6 weeks as we proceed through these milestones.

### DIFF
--- a/handbook/engineering/campaigns/index.md
+++ b/handbook/engineering/campaigns/index.md
@@ -41,7 +41,21 @@ Users provide the code to make the change, we provide the plumbing to turn it in
 
 ## Goals
 
-[WIP Goals (internal only)](https://docs.google.com/document/d/1ADmpfJuVhprHl8Eimuj4fMiVr1kCRVLZpt98HLr9618/edit)
+The short-term goal and primary focus of the team is to *deliver campaigns to all users*. Our milestones toward that goal:
+
+1. Someone on the Campaigns team makes one meaningful change to at least two repositories.
+1. A Sourcegraph engineer (not on this team) makes one meaningful change to at least two repositories.
+1. Reach out to one customer, summarizing the changes in the new Campaigns workflow, to give them a heads-up.
+  - Hypothesis: they will reply within 3 days
+  - Hypothesis: they will engage within 5 days
+1. Reach out to two other customers, as in previous step.
+1. After responding to feedback, we declare the state of Campaigns to be Good and Stable.
+1. Reach out to a small set of customers using Campaigns.
+  - Hypothesis: each one will reply within 3 days
+  - Hypothesis: majority will create at least one new campaign with the new workflow
+  - Hypothesis: majority will send us feedback about Campaigns
+
+[WIP longer term goals (internal only)](https://docs.google.com/document/d/1ADmpfJuVhprHl8Eimuj4fMiVr1kCRVLZpt98HLr9618/edit)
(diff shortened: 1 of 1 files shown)


handbook/engineering/campaigns/index.md
```diff
@@ -41,7 +41,21 @@ Users provide the code to make the change, we provide the plumbing to turn it in
...
 
-[WIP Goals (internal only)](https://docs.google.com/document/d/1ADmpfJuVhprHl8Eimuj4fMiVr1kCRVLZpt98HLr9618/edit)
+The short-term goal and primary focus of the team is to *deliver campaigns to all users*. Our milestones toward that goal:
+
+1. Someone on the Campaigns team makes one meaningful change to at least two repositories.
+1. A Sourcegraph engineer (not on this team) makes one meaningful change to at least two repositories.
+1. Reach out to one customer, summarizing the changes in the new Campaigns workflow, to give them a heads-up.
+   - Hypothesis: they will reply within 3 days
+   - Hypothesis: they will engage within 5 days
+1. Reach out to two other customers, as in previous step.
```
nicksnyder, line 51: I assume we wait until we get one customer to reply and engage before moving on to this step (would be nice for this to be explicit)

handbook/engineering/campaigns/index.md
```diff
@@ -41,7 +41,21 @@ Users provide the code to make the change, we provide the plumbing to turn it in
...
+1. After responding to feedback, we declare the state of Campaigns to be Good and Stable.
```
nicksnyder, line 52: ?

```suggestion
1. After responding to feedback from at least 3 customers, we declare the state of Campaigns to be Good and Stable.
```

handbook/engineering/campaigns/index.md
chrispine: Will do.

handbook/engineering/campaigns/index.md
```diff
@@ -41,7 +41,21 @@ Users provide the code to make the change, we provide the plumbing to turn it in
 
 ## Goals
 
-[WIP Goals (internal only)](https://docs.google.com/document/d/1ADmpfJuVhprHl8Eimuj4fMiVr1kCRVLZpt98HLr9618/edit)
+The short-term goal and primary focus of the team is to *deliver campaigns to all users*. Our milestones toward that goal:
+
+1. Someone on the Campaigns team makes one meaningful change to at least two repositories.
+1. A Sourcegraph engineer (not on this team) makes one meaningful change to at least two repositories.
+1. Reach out to one customer, summarizing the changes in the new Campaigns workflow, to give them a heads-up.
```
mrnugget, line 48: I think we should do a variation of this as soon as we can: we need to let our customers that already use campaigns know about the new workflow and offer to walk them through it. The release is out and the previous workflow doesn't exist anymore.

Take a look at this issue https://github.com/sourcegraph/customer/issues/12 to get an idea of what the email could look like.

I also think it might be good to create such a customer issue, so I went ahead and created a rough draft with email drafts in it: https://github.com/sourcegraph/customer/issues/92

I assigned it to you, but if you don't feel comfortable sending those emails, I can also do it.

We should also probably ask @sqs and #analytics whether there are more customers we should consider reaching out to.

PR #1441, opened 2020-08-25
distribution: Add on-call rotation

pecigonzalo:
Add team on call rotation https://github.com/sourcegraph/sourcegraph/issues/12899


bobheadxi (approved):
thank you!

PR #1443, opened 2020-08-25
Propose CE goals

nicksnyder:
This is really rough, but I wanted to suggest some ideas for goals for the CE team.


dadlerj:
I defer to @juliasourceress on these discussions, I believe she is actively working on this. I don't have any context on this addition.

PR #1452, opened 2020-08-26
Add "Open and transparent" as a value

nicksnyder:
Our openness is already documented in the handbook, and we live by this every day, so I think it is worth calling out here in our values.

### DIFF
--- a/company/values.md
+++ b/company/values.md
@@ -27,3 +27,10 @@ We prefer to [build teams, not workgroups](https://dzone.com/articles/workgroups
 
 As a team's size and scope grows, it (unintentionally) tends to become a workgroup. Resist this by aggressively focusing, or by splitting it into smaller teams.
 
+## Open and transparent
+
+We are an [open company](open_source_open_company.md).
+
+Being open and transparent builds trust with our customers and within our team. It makes it easy for us to partner and collaborate with our customers so we make Sourcegraph the best that it can be. It enables our teammates to work autonomously to solve tough problems while maintaining alignment.
+
+Transparency is hard. It requires extra effort and excellent communication. It means it is possible to observe things that are inconsistent or incomplete. We think this is good because it gives everyone visibility into the process, not just the outcome, and an opportunity to contribute to that process at any point in time.


chrispine (approved):
I really like this, thanks 😄

sqs (approved):

aidaeology (approved):
LGTM

company/values.md
```diff
@@ -27,3 +27,10 @@ We prefer to [build teams, not workgroups](https://dzone.com/articles/workgroups
 
 As a team's size and scope grows, it (unintentionally) tends to become a workgroup. Resist this by aggressively focusing, or by splitting it into smaller teams.
 
+## Open and transparent
+
+We are an [open company](open_source_open_company.md).
+
+Being open and transparent builds trust with our customers and within our team. It makes it easy for us to partner and collaborate with our customers so we make Sourcegraph the best that it can be. It enables our teammates to work autonomously to solve tough problems while maintaining alignment.
+
+Transparency is hard. It requires extra effort and excellent communication. It means it is possible to observe things that are inconsistent or incomplete. We think this is good because it gives everyone visibility into the process, not just the outcome, and an opportunity to contribute to that process at any point in time.
```
slimsag, line 36: I think we can add (or link to) some more specific examples of what we want here.

e.g. "Transparency is hard. It requires extra effort and excellent communication." reads a _little bit_ like it is perhaps ok to not be transparent/open by default and do the transparency bit after the fact ("we discussed internally, here are the results publicly") and I think many folks coming from other organizations will have this interpretation of transparency/open-ness. let's chat more about this

slimsag (approved):

company/values.md
beyang, line 36: Suggested addition:

> We favor transparency by default. There are scenarios in which private conversations are appropriate (e.g., ensuring the privacy of teammates or customers, open security investigations), but absent an explicit reason for keeping communication private, we default to public communication in public channels.

Other thoughts:
* Individuals have a right and expectation to privacy, but teams don't (too extreme?)
* In some cases, people keep stuff private, because they want to create "breathing room" for a new thought or idea to develop before exposing it to wider critique. I think this is totally legitimate, but maybe we can encourage this to happen in the open. E.g., a person can say, "Not looking for a critical eye, but for constructive suggestions on this very new/rough idea".

beyang (approved):

nicksnyder:
I have some improvements based on the feedback from @beyang and @slimsag , but the changes are material enough that I would rather merge this as-is given the approvals and I will open up a subsequent PR.

slimsag:
Sounds good to me!

nicksnyder:
New PR: https://github.com/sourcegraph/about/pull/1515

PR #1462, opened 2020-08-28
Add search goals to team page

poojaj-tech:


handbook/engineering/search/index.md
```diff
@@ -2,9 +2,93 @@
 
 We own the end-to-end Sourcegraph search experience.
 
+Our current focus is documented in the [tracking issue for the current milestone](https://github.com/sourcegraph/sourcegraph/issues?q=is%3Aopen+is%3Aissue+label%3Atracking+label%3Ateam%2Fsearch).
+
+## Vision
+
+Sourcegraph search is:
+
+- **Fast:** Performance of showing results is fast, I can get to what I’m looking for quickly
```
nicksnyder, line 11:
```suggestion
- **Fast:** Performance of showing results is fast and I can get to what I’m looking for quickly.
```

handbook/engineering/search/index.md
```diff
@@ -2,9 +2,93 @@
...
+### Fast
+
+- **Streaming search**
+  - **Problem:** We want to make significant improvements to search that are blocked by not having the infrastructure needed.
+  - **Owner:** Keegan
+  - **Status:** In progress
+
+### Expressive
+
+- **Hierarchical search (AND/OR/NOT)**
```
nicksnyder, line 44: @rvantonder I think you had an idea for a different name for this, what did you call it?

handbook/engineering/search/index.md
```diff
@@ -2,9 +2,93 @@
...
+  - **Problem:** We have customers who want to migrate from OpenGrok and other search tools, and they want to be able to do searches that are available in those tools.
+  - **Outcomes:** Users can run complex searches on Sourcegraph.
```
nicksnyder, line 46:
```suggestion
  - **Outcomes:** OpenGrok eagerly migrate to Sourcegraph.
```

handbook/engineering/search/index.md
```diff
@@ -2,9 +2,93 @@
 
 We own the end-to-end Sourcegraph search experience.
 
+Our current focus is documented in the [tracking issue for the current milestone](https://github.com/sourcegraph/sourcegraph/issues?q=is%3Aopen+is%3Aissue+label%3Atracking+label%3Ateam%2Fsearch).
```
nicksnyder, line 5: This probably belongs under the `## Short term` heading

rvantonder:
> First pass lgtm. Let's get feedback from the team.

Not sure if this is ready for feedback from the team in general, I only saw this because I was mentioned. When ready, please let us know about feedback by sharing in #search or bring it out of draft and assign search team or CC `@sourcegraph/search` in the comments it should stay in draft mode.

handbook/engineering/search/index.md
```diff
@@ -2,9 +2,93 @@
...
 
+## Vision
+
+Sourcegraph search is:
+
+- **Fast:** Performance of showing results is fast and I can get to what I’m looking for quickly.
+- **Expressive:** It is possible to construct a search query to find exactly the results I’m looking for. The tools, syntax, filters, are provided to construct the queries I want to write.
+The answer to “Can I do ‘X’ with Sourcegraph?” will always be yes.
+- **Easy to use:** Users can figure out how to find what they are looking for, the syntax is clear, new users understand what options they have for searching.
+- **Sharable:** Sourcegraph search creates network effects that provide compounding value with scale.
```
rvantonder, line 13: I agree with the value of making search shareable. It's not clear how this fits into the vision. "network effects that provide compounding value with scale." -> sounds very buzz wordy biz speak. I would venture the average user developer doesn't care ("so what?") and for Sourcegraph engineers working on search, I'm not sure this attribute guides the work (does it mean we should create links to share?). I think this could be a tag line for investors, but I'm not sure that's the target audience of this page. Maybe it's intended to mean something else, in which case, a rephrase?

handbook/engineering/search/index.md
```diff
@@ -2,9 +2,93 @@
 
 We own the end-to-end Sourcegraph search experience.
 
+## Vision
+
+Sourcegraph search is:
+
+- **Fast:** Performance of showing results is fast and I can get to what I’m looking for quickly.
+- **Expressive:** It is possible to construct a search query to find exactly the results I’m looking for. The tools, syntax, filters, are provided to construct the queries I want to write.
+The answer to “Can I do ‘X’ with Sourcegraph?” will always be yes.
```
rvantonder, line 11: Either:

> We _strive_ for a search experience that is...

 to preface

> The answer to “Can I do ‘X’ with Sourcegraph?” will always be yes.

or, 

> Sourcegraph search is...

  followed by

 > The answer to “Can I do ‘X’ with Sourcegraph?” _should_ always be yes.

The current phrasing just feels matter-of-fact (it is so), which seems inaccurate, when I think we are going for something aspirational (we continually strive to make it so) or a desired trait (according to a specification or desired behavior), rather than "it is definitely so".

handbook/engineering/search/index.md
```diff
@@ -2,9 +2,93 @@
...
+## Long term (~1 year)
+
+Sourcegraph search is far and away the best code search solution, search results are fast, and I can find what I’m looking for quickly.
+
+## Medium term (3-6 months)
+
+In the next 6 months we want to be halfway to the long term goal. To achieve this, we need to accomplish the following outcomes:
+
+- Getting to search results faster through a better search experience.
+- What do we need to do to get to the speed of grep.app?
```
rvantonder, line 28: I don't know that grep.app is faster. It is perhaps perceptually faster because I think it only searches a couple of repos by default. The goal here might be to rather deliver a faster search experience, in which case, maybe we should state this in terms of desired time-to-result for users, and what scope of repos we want to be included in such an experience. Overall I'm not a big fan of defining goals by naming alternative solutions, could we extract the attributes that those solutions have and define our goals in terms of them? what is it about the speed of grep.app that appeals?

handbook/engineering/search/index.md
```diff
@@ -2,9 +2,93 @@
...
+- Scale fast search to 500k repositories.
+
+## Short term (1-3 months)
+
+Our current focus is documented in the [tracking issue for the current milestone](https://github.com/sourcegraph/sourcegraph/issues?q=is%3Aopen+is%3Aissue+label%3Atracking+label%3Ateam%2Fsearch).
+
+### Fast
+
+- **Streaming search**
+  - **Problem:** We want to make significant improvements to search that are blocked by not having the infrastructure needed.
```
rvantonder, line 38: I think this goal also ties back to the 'make search feel faster/more responsive' in line with grep.app.

handbook/engineering/search/index.md
rvantonder: Search expressions. For the general audience, I wrote up why I'm adopting this terminology in https://github.com/sourcegraph/sourcegraph/issues/13126. "Search expressions" doesn't "have" to be the product label we slap onto this (we could simply go with "Sourcegraph's Query Language" or "Sourcegraph Queries" that's expanded over time. As far as internal goals and conveying the work, search expressions is more accurate than previous terms :-)

>

handbook/engineering/search/index.md
```diff
@@ -2,9 +2,93 @@
...
+  - **Owner:** Keegan
+  - **Status:** In progress
+
+### Expressive
+
+- **Hierarchical search (AND/OR/NOT)**
```
rvantonder, line 44:
```suggestion
- **Search Query Language Extensions**
```

handbook/engineering/search/index.md
```diff
@@ -2,9 +2,93 @@
...
+  - **Problem:** We have customers who want to migrate from OpenGrok and other search tools, and they want to be able to do searches that are available in those tools.
```
rvantonder, line 45:
```suggestion
  - **Problem:** For Big Code it is imperative that users can find and filter the code, files, repositories, and commits that they work on.
```

How about this? The OpenGrok functionality is solved (from an expressivity POV for and/or queries). If we still have customers who are looking for this, let's get in touch and let them know.

handbook/engineering/search/index.md
```diff
@@ -2,9 +2,93 @@
...
+  - **Outcomes:** Users can run complex searches on Sourcegraph.
```
rvantonder, line 46:
```suggestion
  - **Outcomes:** Users can express relations on code, files, repositories, and commits in search queries to more effectively filter the data they need
```

"Relations" here meaning, e.g., "repo has file" "repo has commit", ... etc.

handbook/engineering/search/index.md
```diff
@@ -2,9 +2,93 @@
...
+  - **Owner:** Rijnard
+  - **Status:** In progress
+- **Improve syntax for existing filters like `repohasfile`**
```
rvantonder, line 49:
```suggestion
- **Develop and implement a query language extension to express relations (i.e., rules) on code, files, repositories, and commits**
```

handbook/engineering/search/index.md
rvantonder: Addendum: I see now shareable refers to the code monitoring/saved searches work at the bottom :-) I feel like the code monitoring/search API use is much more aligned with "code insights" than core search concerns (code insights also very much benefits from search functionality, and we have requests to support certain filters, for example). The "improve diff search performance" aspect of code monitoring is core to the "make search fast", but I'm not sure about the rest.

rvantonder:
Done looking :-)

handbook/engineering/search/index.md
```diff
@@ -2,9 +2,93 @@
...
+  - **Owner:** Stefan
+  - **Status:** In progress
+
+### Easy to use
+
+- **Get to files and repositories more quickly**
+  - **Problem:** Finding files and repositories often takes multiple tries to get the syntax correct or creative solutions to get the result users are looking for.
+  - **Outcomes:** Users leverage file and repo filters to quickly get to the code they are searching for.
+  - **Owner:** Stefan
+  - **Status:** In progress
```
stefanhengl, line 65: Does this refer to the wildcard syntax I am working on? If yes, I am not sure that the wildcard syntax addresses the problem statement. The assumption would be that users fail on regex and not on the filter syntax. Is that really the case? If we just want to increase the adoption of repo and file filters then maybe this is more of a UX topic?

handbook/engineering/search/index.md
```diff
@@ -2,9 +2,93 @@
...
+- What do we need to do to get to the speed of grep.app?
+- Scale fast search to 500k repositories.
+
+## Short term (1-3 months)
+
+Our current focus is documented in the [tracking issue for the current milestone](https://github.com/sourcegraph/sourcegraph/issues?q=is%3Aopen+is%3Aissue+label%3Atracking+label%3Ateam%2Fsearch).
+
+### Fast
+
+- **Streaming search**
```
stefanhengl, line 37: In my opinion "Streaming search" is more a solution than a goal

handbook/engineering/search/index.md
```diff
@@ -2,9 +2,93 @@
...
 We own the end-to-end Sourcegraph search experience.
 
+## Vision
+
+Sourcegraph search is:
+
+- **Fast:** Performance of showing results is fast and I can get to what I’m looking for quickly.
+- **Expressive:** It is possible to construct a search query to find exactly the results I’m looking for. The tools, syntax, filters, are provided to construct the queries I want to write.
+The answer to “Can I do ‘X’ with Sourcegraph?” will always be yes.
+- **Easy to use:** Users can figure out how to find what they are looking for, the syntax is clear, new users understand what options they have for searching.
```
attfarhan, line 12: > Users can figure out how to find what they are looking for,

This feels a bit vague, and already true. If users dug deep enough into our docs, they would indeed be able to find out how to use our search. I think the goal is to make our product intuitive enough that new users know what to do on their first visit.

The rest of it, I really like. Perhaps a rephrasing:

```suggestion
- **Easy to use:**  Users quickly understand how to conduct a search and what options are available for searching. The search syntax is clear and intuitive.
```

attfarhan (approved):

handbook/engineering/search/index.md
poojaj-tech: I really like this suggestion:
_The search syntax is clear and intuitive._ 

Some parts of this were left open-ended intentionally: 

> How to find

through conducting a search, using the search results UI more effectively, consuming multiple notification channels

> What they are looking for

code, repositories, files, metadata, etc.

What do you think of this rephrasing:
Users can quickly understand how to find what they are looking for and what options are available for searching. The search syntax is clear and intuitive.

handbook/engineering/search/index.md
```diff
@@ -2,9 +2,93 @@
...
+          - **Status:** In progress
+      - It is hard for users to quickly get to code they care about.
+          - **Owner:** Farhan, Juliana
+          - **Status:** In progress
+    - **Outcomes:**
+      - New users are introduced to Sourcegraph and are able to more quickly see the value and run searches more quickly.
+      - Users can run searches over code they care about more quickly.
+      - Sourcegraph surfaces code and searches users care about.
+
+### Sharable
```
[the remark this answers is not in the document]
poojaj-tech: I think the north american spelling is "Sharable", but I agree that "Shareable" is more intuitive 🙂

handbook/engineering/search/index.md
poojaj-tech: I like _search expressions_ a lot. It feels intuitive to gather that it's about combining two or more search expressions with operators. 

If we go this route we should what a ‘search expression’ is. I think it’s easy to confuse it with phrasing that implies similar meaning in our docs today e.g. 'search patterns' used throughout the [docs](https://docs.sourcegraph.com/user/search) and there's also [regular expression search](https://docs.sourcegraph.com/user/search/queries#regular-expression-search).
There's related discussion underway in this [product terms RFC](https://docs.google.com/document/d/1SHky6nodPs1w_zRXz24jB2nq5LbMCJl1u7hcLTGHDL8/edit#heading=h.trqab8y0kufp). Search expressions in this RFC are defined as the smallest units of valid search queries (without operators AND/OR/NOT) and my understanding is the phrasing 'search pattern' should be replaced with 'search expression' in the docs.

Perhaps “Combining Search Expressions” is a clearer goal here as it describes the ability to compose individual search expressions including the ability to nest these searches.

poojaj-tech: I like "Search expressions (AND/OR/NOT)" a lot. I agree that it feels more intuitive and I can easily tell that it's about combining multiple search expressions :)

If we go this route we should define and agree on what the definition of a "search expression" is. I think it’s easy to confuse it with phrasing that implies similar meaning in our docs today e.g. "search patterns" is used throughout the [docs](https://docs.sourcegraph.com/user/search) and there's also ["regular expression search"](https://docs.sourcegraph.com/user/search/queries#regular-expression-search).
There's related discussion underway in this [product terms RFC](https://docs.google.com/document/d/1SHky6nodPs1w_zRXz24jB2nq5LbMCJl1u7hcLTGHDL8/edit#heading=h.trqab8y0kufp). Search expressions in this RFC are defined as smallest units of valid search queries (without operators AND/OR/NOT) and my understanding is the phrasing "search pattern" should be replaced with "search expression" in the docs. What do you think?

Perhaps "Combining Search Expressions" is a clearer goal as it describes the ability to compose/combine the individual search expressions (following the definition of a search expression above) including the ability to nest these searches.

handbook/engineering/search/index.md
poojaj-tech: Great suggestion, updated to add this - "OpenGrok users eagerly migrate to Sourcegraph, and can run complex searches on Sourcegraph." Let me know if you think the second part is unnecessary.

handbook/engineering/search/index.md
poojaj-tech: I can see how this phrasing for “Shareable” isn’t clear. Shareable search here refers to growing adoption of search through sharing searches that provide team and organization level value. 
Suggested rephrasing: 
> Sourcegraph searches can be easily shared, and provide team and organization wide value. This in turn creates network effects that help compound the value Sourcegraph provides with scale. 

How does this sound? Is it clear how this fits into the vision?

I definitely expect code monitoring to evolve to be a cross-functional effort (somewhat similar to code-intel aware search in the longer term). The framework Quinn Keast lays down for code monitoring in [this RFC](https://docs.google.com/document/d/1_R5DgpUkxyZilsJ9vBQm5cvRPT2udc3tZIPg2q3cnZU/edit#) does a great job explaining how this can be expanded to include campaigns and code insights in the future. 
The first leg of this work is targeted towards search queries. Enabling code monitoring for search queries helps users/teams/orgs keep track of important code changes in their codebases. Notifying users in channels outside of Sourcegraph raises awareness and adoption for these search queries + search results. Like you pointed out, performance considerations like diff search will be an important part of solving this use case for search queries.

handbook/engineering/search/index.md
```diff
@@ -2,9 +2,93 @@
...
+## Long term (~1 year)
+
+Sourcegraph search is far and away the best code search solution, search results are fast, and I can find what I’m looking for quickly.
+
+## Medium term (3-6 months)
+
+In the next 6 months we want to be halfway to the long term goal. To achieve this, we need to accomplish the following outcomes:
+
+- Getting to search results faster through a better search experience.
+- What do we need to do to get to the speed of grep.app?
```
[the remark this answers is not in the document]
poojaj-tech: Updated, is URL now.

handbook/engineering/search/index.md
poojaj-tech: Yep, this is the higher level problem statement for your work on wildcard syntax. I’m curious if you think there’s a different problem statement the wildcard syntax solves that isn’t captured?
> The assumption would be that users fail on regex and not on the filter syntax.
> We just want to increase the adoption of repo and file filters then maybe this is more of a UX topic.

Strongly agree with both of your points - 1/ validating the assumption that users fail on regex and not the filter syntax through research and 2/ understanding the UX aspect as a part of the solution space for this problem. If we go the route of understanding the UX, we can add that as sub-bullet under _Plan_ (see latest commit for change).

handbook/engineering/search/index.md
```diff
@@ -2,9 +2,99 @@
...
+Our current focus is documented in the [tracking issue for the current milestone](https://github.com/sourcegraph/sourcegraph/issues?q=is%3Aopen+is%3Aissue+label%3Atracking+label%3Ateam%2Fsearch).
+
+### Fast
+
+- **Unblock improving user perceived performance**
+  - **Problem:** We want to make significant improvements to search that are blocked by not having the infrastructure needed.
+  - **Outcome:** 
+      - Unblock the ability to add responsive and fast-loading search results 
+      - Return results faster for large result sets, e.g. indexed repositories in very large codebases
+  - **Plan:** Streaming search
```
poojaj-tech, line 41: Rephrased based on @stefanhengl and @rvantonder 's suggestions to call out that this is a core infrastructure change that unblocks the ability to improve user perceived search performance and responsiveness. My understanding is that in the short term, the first step is add this core infrastructure, and then leverage this work to improve the user experience—mentioned as #2 in Medium term goals. 
CC: @keegancsmith in case you have thoughts here.

handbook/engineering/search/index.md
poojaj-tech: If this work is solved, we can mark the status as “done” here. My understanding was that there is still some work (perhaps parser migration, UX/UI work) remaining here.
> For Big Code it is imperative that users can find and filter the code, files, repositories, and commits that they work on.

Why does big code make this imperative?

handbook/engineering/search/index.md
poojaj-tech: > "Relations" here meaning, e.g., "repo has file" "repo has commit", ... etc.

Hmm this is AND/OR search. Did you mean to suggest this for the goal below (Improve syntax for existing filters like `repohasfile`)?
> Sourcegraph is the only industrial-strength search solution that provides these capabilities.

This feels a bit vision-like or something we should aspire to be.

handbook/engineering/search/index.md
poojaj-tech: Perhaps something like "Improve search experience across different entities in the codebase." would be less specific since my guess is relations can be across many different entities. 
Would the outcome be to "help users more effectively filter the data they need"?

handbook/engineering/search/index.md
```diff
@@ -2,9 +2,93 @@
...
 ## Goals
 
-Our current goals are documented in the [search team's current tracking issue](https://github.com/sourcegraph/sourcegraph/issues?q=is%3Aopen+is%3Aissue+label%3Atracking+label%3Ateam%2Fsearch).
+_Updated 2020-08-27_
+
+## Long term (~1 year)
+
+Sourcegraph search is far and away the best code search solution, search results are fast, and I can find what I’m looking for quickly.
+
+## Medium term (3-6 months)
```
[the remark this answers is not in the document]
poojaj-tech: You’re right, these are just a starting point and definitely need more iteration. This right now captures outcome-driven goals we think will be most important to solve in the medium term. Multiple solutions can solve for each one of these goals. 
> better search experience.

"ease of use" is an important one and can be captured under this goal. We should focus on it if we think it's the best way for users to quickly find the results they are looking for.

handbook/engineering/search/index.md
stefanhengl: I agree it is more of a UX research topic right now. I suggest to remove the point `**Get to files and repositories more quickly**` entirely. Like I said above it is not clear to me that Wildcard syntax is related to this goal. I would suggest to first implement what is listed under `### Easy to use` and see whether adoption of our syntax goes up. 

Let's validate whether regex is indeed a widespread problem for our key target group before adding wildcard syntax to our short term goals.

Especially in the short term, I want to focus on the expressiveness of our query language first.

handbook/engineering/search/index.md
```diff
@@ -2,9 +2,99 @@
...
+
+In the next 6 months we want to be halfway to the long term goal. To achieve this, we need to accomplish the following outcomes:
```
lguychard, line 24: It's a bit unclear to me how the following outcomes mean we're "halfway" to the long-term goal above. After achieving these outcomes, what is missing for us to have achieved the long-term goal?

handbook/engineering/search/index.md
```diff
@@ -2,9 +2,93 @@
...
+- **Improve search experience**
+  - **Problem:**
+      - New users who are trying Sourcegraph for the first time have trouble learning the syntax and breadth of Sourcegraph features.
+          - **Owner:** Farhan
+          - **Status:** In progress
+      - It is hard for users to quickly get to code they care about.
+          - **Owner:** Farhan, Juliana
+          - **Status:** In progress
+    - **Outcomes:**
+      - New users introduced to Sourcegraph are able to more run searches and see the value of Sourcegraph more quickly.
```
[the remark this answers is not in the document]
lguychard: Suggested reword

```suggestion
      - New users introduced to Sourcegraph are able to quickly run searches that show them the value of Sourcegraph.
```

handbook/engineering/search/index.md
```diff
@@ -2,9 +2,99 @@
...
+  - **Problem:** Users want to be notified about important things going on in their code.
+  - **Outcome:** Notifications create a shared understanding and raise awareness of what’s going on in the code.
+  - **Plan**
+      - Improved diff search performance
+      - Webhooks on search results
+          - Slack integration
+          - Zapier integration
+          - Email
+  - **Owner:** TBD
+  - **Status:** Not started
```
lguychard, line 97: This short-term goal does not appear to tie back to the medium-term goal. Is this intentional?

handbook/engineering/search/index.md
```diff
@@ -2,9 +2,99 @@
...
+- **Shareable:** Sourcegraph searches are easy to share, and provide team and organization wide value. This in turn creates network effects that compound the value Sourcegraph provides with scale.
+
 ## Goals
 
-Our current goals are documented in the [search team's current tracking issue](https://github.com/sourcegraph/sourcegraph/issues?q=is%3Aopen+is%3Aissue+label%3Atracking+label%3Ateam%2Fsearch).
+_Updated 2020-08-27_
+
+## Long term (~1 year)
+
+Sourcegraph search is far and away the best code search solution, search results are fast, and I can find what I’m looking for quickly.
```
lguychard, line 20: I am wondering whether this long-term goal adds value when it comes to guiding the team's work. It's part outcome of achieving all four pillars of the vision (fast+expressive+easy to use+shareable search = Sourcegraph is far and away the best code search solution), part unquantified focus on certain pillars (search results are fast -> fast, I can find what I'm looking for quickly -> expressive + easy to use).

Is there a more concrete outcome we could/should aim for here?

lguychard (commented):
Some first thoughts

handbook/engineering/search/index.md
```diff
@@ -2,9 +2,99 @@
 
 We own the end-to-end Sourcegraph search experience.
 
+## Vision
+
+Sourcegraph search is:
+
+- **Fast:** Performance of showing results is fast and I can get to what I’m looking for quickly.
+- **Expressive:** It is possible to construct a search query to find exactly the results I’m looking for. The tools, syntax, filters, are provided to construct the queries I want to write. We strive for a search experience where the answer to “Can I do ‘X’ with Sourcegraph?” will always be yes.
```
beyang, line 10: Suggested addition::

**Scalable:** Sourcegraph scales to any private codebase and the size of the open-source universe. Users can trust that the result set is drawn from the entire set of code they care about, and if the result isn't found, it doesn't exist.

beyang:
Very excited to converge on a set of longer-term goals—thank you for writing this up and kicking this off @poojaj-tech!

keegancsmith (approved):
LGTM.

Been thinking about achieving fast indexed text search on Sourcegraph.com once it has many more repos (cc @beyang). I believe this requires doubling down on Zoekt. The amount of work to get where we want in 3-6 months is quite a bit. I think a team of two focussed on Zoekt would get us there (and when necessary the layers above it). Personally I am quite excited about that work, but to get to something great it's more than I can do in 6 months. I'd love to work closely with someone on Zoekt though :)

handbook/engineering/search/index.md
rvantonder: I think this is a good/important addition and zero objection to adding it. The one thing I'd call out here though is that the scope of a solution potentially goes beyond search. If we're talking about fast indexed search, sure, we can think in terms of horizontally scaling Zoekt, say. If we're talking about "exhaustively search for all instances so that I can replace X with Y in a campaign" it demands a search/architecture solution and scope that goes a bit beyond what I think the search team (at least relative to its current size) can focus on. I could see this being part of a core-like team that potentially branches out of a search concern.

nicksnyder: It is ok to have goals/scope beyond the capabilities of the current team. This is valuable information that can help drive growth decisions.

handbook/engineering/search/index.md
poojaj-tech: Extracted **fast-loading** and **responsive** as key attributes that make grep.app so appealing.

handbook/engineering/search/index.md
poojaj-tech: Updated to **Search Expressions**, based on conversation above
https://github.com/sourcegraph/about/pull/1462#discussion_r479377118

handbook/engineering/search/index.md
poojaj-tech: No - this isn’t intentional. The medium term goals are focussed on how search can be “fast” and “easy to use” right now as those are problems we anticipate needing to solve the most.

stefanhengl (approved):

handbook/engineering/search/index.md
nicksnyder:
```suggestion
- **Scalable:** Sourcegraph scales to any private codebase and the size of the open-source universe. Users can trust that the result set is drawn from the entire set of code they care about, and if the result isn't found, it doesn't exist.
- **Expressive:** It is possible to construct a search query to find exactly the results I’m looking for. The tools, syntax, filters, are provided to construct the queries I want to write. We strive for a search experience where the answer to “Can I do ‘X’ with Sourcegraph?” will always be yes.
```

nicksnyder:
```suggestion
- **Scalable:** Sourcegraph scales to any private codebase and the size of the open-source universe. Users can trust that the result set is drawn from the entire set of code they care about, and if the result isn't found, it doesn't exist.
- **Expressive:** It is possible to construct a search query to find exactly the results I’m looking for. The tools, syntax, filters, are provided to construct the queries I want to write. We strive for a search experience where the answer to “Can I do ‘X’ with Sourcegraph?” will always be yes.
```

handbook/engineering/search/index.md
```diff
@@ -2,9 +2,94 @@
...
+- Scale fast search to 500k repositories.
+
+## Short term (1-3 months)
+
+Our current focus is documented in the [tracking issue for the current milestone](https://github.com/sourcegraph/sourcegraph/issues?q=is%3Aopen+is%3Aissue+label%3Atracking+label%3Ateam%2Fsearch).
+
+### Fast
+
+- **Unblock improving user perceived performance**
+  - **Problem:** We want to make significant improvements to search that are blocked by not having the infrastructure needed.
```
nicksnyder, line 38: This doesn't feel like a problem statement.

What about this?

```suggestion
  - **Problem:** Our current search infrastructure relies on on-shot requests with aggressive timeouts, but it is hard to tune those timeouts for large codebases. If we make the timeouts too small, then search may not return results for needle-in-the-haystack queries (because the search timed out). If we make the timeouts too large, then search becomes slower across the board. Detecting and handling these timeouts across multiple service boundaries is also error prone and has led to bugs.
```

Is this accurate, @keegancsmith ? If not, what specific problem does streaming search solve?

handbook/engineering/search/index.md
nicksnyder: I agree, I think the vision is clear and this long term section doesn't add anything. I would delete this section.

handbook/engineering/search/index.md
nicksnyder: I agree, I think we should delete this.

```suggestion
```

handbook/engineering/search/index.md
```diff
@@ -2,9 +2,94 @@
...
+
+## Long term (~1 year)
+
+Sourcegraph search is far and away the best code search solution, search results are fast, and I can find what I’m looking for quickly.
+
+## Medium term (3-6 months)
+
+In the next 6 months we want to be halfway to the long term goal. To achieve this, we need to accomplish the following outcomes:
+
+- Getting to search results faster through a better search experience.
```
nicksnyder, line 21: I think this is too vague and should be either clarified (made into a problem statement) or deleted. What is slowing users down today?

```suggestion
```

handbook/engineering/search/index.md
keegancsmith: Sounds good. I would also emphasize the larger the timeout, the longer a user will wait to receive a small result set. Main benefit is a user will start getting results from the very first result we find, rather than waiting for the full one-shot search (which can be multiple seconds with certain queries on large instances)

handbook/engineering/search/index.md
```diff
@@ -2,9 +2,88 @@
...
 ## Goals
 
-Our current goals are documented in the [search team's current tracking issue](https://github.com/sourcegraph/sourcegraph/issues?q=is%3Aopen+is%3Aissue+label%3Atracking+label%3Ateam%2Fsearch).
+_Updated 2020-08-27_
+
+## Medium term (3-6 months)
+
+- Getting to search results faster through a better search experience.
+- Users perceive search results to be as fast-loading and responsive as [grep.app](https://grep.app/)
+- Scale fast search to 500k repositories.
```
nicksnyder, line 23:
```suggestion
### Scale indexed search to 500k repositories

- **Problem:** We have customers who need Sourcegraph to scale to 500k repositories. Some parts of Sourcegraph don't work well at that scale.
- **Owners:** TBD
- **Plan:** Incrementally add repositories to Sourcegraph.com until searches get slow or start breaking. Fix those things. Then continue adding repositories.
- **Success criteria:** Sourcegraph can search 500k repositories in less than 300ms. This is on par with [grep.app](https://grep.app).
- **Status:** Sourcegraph.com current has about 100k repositories.
```

lguychard (approved):
LGTM, thanks @poojaj-tech 🙂

handbook/engineering/search/index.md
```diff
@@ -2,9 +2,92 @@
...
+
 ## Goals
 
-Our current goals are documented in the [search team's current tracking issue](https://github.com/sourcegraph/sourcegraph/issues?q=is%3Aopen+is%3Aissue+label%3Atracking+label%3Ateam%2Fsearch).
+_Updated 2020-09-05_
+
+## [WIP] Medium term (3-6 months)
+
+### Scalable
+  - **Scale indexed search to 500k repositories**
```
nicksnyder, line 22: I know this is inconsistent with the short term goals below, but I actually think it is better to bring the concrete goal to the forefront (rather than categorizing) and we should do this for all goals (not just this one). If our goals are clearly written the it should be obvious how these contribute to the vision (therefore it isn't valuable to explicitly have a heading).

If you want to merge this PR as-is, we can discuss this idea in a followup PR.

```suggestion
### Scale indexed search to 500k repositories
```

handbook/engineering/search/index.md
```diff
@@ -2,9 +2,92 @@
...
+- **Expressive:** It is possible to construct a search query to find exactly the results I’m looking for. The tools, syntax, filters, are provided to construct the queries I want to write. We strive for a search experience where the answer to “Can I do ‘X’ with Sourcegraph?” will always be yes.
+- **Easy to use:** Users can quickly understand how to find what they are looking for and what options are available for searching. The search syntax is clear and intuitive.
+- **Shareable:** Sourcegraph searches are easy to share, and provide team and organization wide value. This in turn creates network effects that compound the value Sourcegraph provides with scale.
+
 ## Goals
 
-Our current goals are documented in the [search team's current tracking issue](https://github.com/sourcegraph/sourcegraph/issues?q=is%3Aopen+is%3Aissue+label%3Atracking+label%3Ateam%2Fsearch).
+_Updated 2020-09-05_
+
+## [WIP] Medium term (3-6 months)
```
nicksnyder, line 19: I am wondering if we should delete these headings and instead focus on setting timelines per goal.

nicksnyder (approved):
Looks good, I added a few more ideas for improving this but we can address in a followup PR if you want.

handbook/engineering/search/index.md
```diff
@@ -2,9 +2,92 @@
...
+    - **Outcome:** Sourcegraph can search 500k repositories in less than 300ms. This is on par with [grep.app](https://grep.app).
+    - **Plan:** Incrementally add repositories to Sourcegraph.com until searches get slow or start breaking. Fix those things. Then continue adding repositories.
+    - **Owners:** TBD
+    - **Status:** Sourcegraph.com currently has about 100k repositories.
+
+## Short term (1-3 months)
+
+Our current focus is documented in the [tracking issue for the current milestone](https://github.com/sourcegraph/sourcegraph/issues?q=is%3Aopen+is%3Aissue+label%3Atracking+label%3Ateam%2Fsearch).
+
+### Fast
```
christinaforney, line 33:
```suggestion
### Fast and scalable
```

limitedmage (approved):

beyang (approved):

handbook/engineering/search/index.md
poojaj-tech: Updating the PR to add this suggestion - I agree that clearly written goals should make it obvious how these contribute to the vision. This can be addressed for the other goals in a follow up.

handbook/engineering/search/index.md
poojaj-tech: Focussing on setting timelines per goal sounds great since that information isn’t really captured with these headings. A clear status (not started / started or estimated start if it’s available) along with the timeline will help communicate how we are thinking about prioritizing these goals.

poojaj-tech:
Thanks everyone for the great discussion and feedback on this PR! Excited to merge this as a starting point - we will continue to iterate on and evolve these goals.

@rijnard: Thanks for the really insightful comments. I’ve left the discussions threads around your comments open. Let's get together with @loic and discuss your feedback and concerns when you’re back.

PR #1479, opened 2020-08-28
Engineering leadership status updates

nicksnyder:
I can't sustain attending each team's weekly meeting, but I still need to know what is going on. I also don't think it will work for me to subscribe and follow each team's tracking issue each month because the updates there are too granular.

My proposal is for each EM to send a weekly email as documented, but I am open to other feedback and suggestions.

### DIFF
--- a/handbook/engineering/leadership/index.md
+++ b/handbook/engineering/leadership/index.md
@@ -79,6 +79,8 @@ Engineering managers send an update to engineering-leadership-status@sourcegraph
 - What are the plans for this week?
 - What issues should my peer-team/manager be aware of?
 
+If you have sensitive information to communicate, please send a separate email directly to the relevant people.
+
 ## Book recommendations
 
 Here are some books that multiple people on the team have read and recommend. If you haven't read these, [use your education budget](../../people-ops/travel.md#professional-development-and-education) and read them!
--- a/handbook/engineering/leadership/index.md
+++ b/handbook/engineering/leadership/index.md
@@ -73,7 +73,7 @@ Our long term vision is to build an organization that attracts and retains the b
 
 We have a 1hr sync with product and engineering every Tuesday at 9am PT ([meeting notes](https://docs.google.com/document/d/10fEh3Cw0ENKjFYDP-4OoLZbOZTVAtMpifc2WM_1mGVU/edit)).
 
-Engineering managers send an update to engineering-leadership-status@sourcegraph.com by 10am PT every Monday (sending it earlier, like EOD Friday is ok too). The purpose of the update is to keep interested parties (including the VP of Engineering, and peer managers) informed of what is going on in each part of our organization. The format of this update is at the discretion of the sender, but it should be prose that generally answers the following questions:
+Engineering managers send an update to engineering-leadership-status@sourcegraph.com by 12pm (noon) PT every Monday (sending it earlier, like EOD Friday is ok too). The purpose of the update is to keep interested parties (including the VP of Engineering, and peer managers) informed of what is going on in each part of our organization. The format of this update is at the discretion of the sender, but it should be prose that generally answers the following questions:
 
 - What important things happened last week?
 - What are the plans for this week?
--- a/handbook/engineering/leadership/index.md
+++ b/handbook/engineering/leadership/index.md
@@ -73,7 +73,7 @@ Our long term vision is to build an organization that attracts and retains the b
 
 We have a 1hr sync with product and engineering every Tuesday at 9am PT ([meeting notes](https://docs.google.com/document/d/10fEh3Cw0ENKjFYDP-4OoLZbOZTVAtMpifc2WM_1mGVU/edit)).
 
-Engineering managers send an update to engineering-leadership-status@sourcegraph.com by 12pm (noon) PT every Monday (sending it earlier, like EOD Friday is ok too). The purpose of the update is to keep interested parties (including the VP of Engineering, and peer managers) informed of what is going on in each part of our organization. The format of this update is at the discretion of the sender, but it should be prose that generally answers the following questions:
+Engineering managers send an update to [engineering-leadership-status@sourcegraph.com](https://groups.google.com/a/sourcegraph.com/g/engineering-leadership-status) by 12pm (noon) PT every Monday (sending it earlier, like EOD Friday is ok too). The purpose of the update is to keep interested parties (including the VP of Engineering, and peer managers) informed of what is going on in each part of our organization. The format of this update is at the discretion of the sender, but it should be prose that generally answers the following questions:
 
 - What important things happened last week?
 - What are the plans for this week?
--- a/handbook/engineering/leadership/index.md
+++ b/handbook/engineering/leadership/index.md
@@ -77,7 +77,7 @@ Engineering managers send an update to engineering-leadership-status@sourcegraph
 
 - What important things happened last week?
 - What are the plans for this week?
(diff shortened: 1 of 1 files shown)


handbook/engineering/leadership/index.md
```diff
@@ -69,6 +69,18 @@ Our long term vision is to build an organization that attracts and retains the b
 - Owner: TBD.
 - Status: Not started. Needs more discussion.
 
+## Communication
+
+We have a 1hr sync with product and engineering every Tuesday at 9am PT ([meeting notes](https://docs.google.com/document/d/10fEh3Cw0ENKjFYDP-4OoLZbOZTVAtMpifc2WM_1mGVU/edit)).
+
+Engineering managers send an update to engineering-leadership-status@sourcegraph.com by 10am PT every Monday (sending it earlier, like EOD Friday is ok too). The purpose of the update is to keep interested parties (including the VP of Engineering, and peer managers) informed of what is going on in each part of our organization. The format of this update is at the discretion of the sender, but it should be prose that generally answers the following questions:
```
chrispine, line 76: Could we make it noon instead? We just moved the Campaigns Sync meeting to start at 10a as requested, and the company meeting follows immediately after. I'd like to post my status update after the Sync meeting.

If that doesn't work, then I'd like to reschedule the Campaigns Sync to happen earlier.

nicksnyder:
```suggestion
Engineering managers send an update to engineering-leadership-status@sourcegraph.com by 12pm (noon) PT every Monday (sending it earlier, like EOD Friday is ok too). The purpose of the update is to keep interested parties (including the VP of Engineering, and peer managers) informed of what is going on in each part of our organization. The format of this update is at the discretion of the sender, but it should be prose that generally answers the following questions:
```

handbook/engineering/leadership/index.md
```diff
@@ -69,6 +69,18 @@ Our long term vision is to build an organization that attracts and retains the b
 - Owner: TBD.
 - Status: Not started. Needs more discussion.
 
+## Communication
+
+We have a 1hr sync with product and engineering every Tuesday at 9am PT ([meeting notes](https://docs.google.com/document/d/10fEh3Cw0ENKjFYDP-4OoLZbOZTVAtMpifc2WM_1mGVU/edit)).
```
chayim, line 74: Rather than specify the time of the meeting, how about maintaining just the document link.  On a personal note, I was hoping to move this meeting - and wouldn't want to maintain the handbook as a result of times shifting.

handbook/engineering/leadership/index.md
chayim, line 76: What do you think about combining this with a nudge to complete; this could be as simple as the existing web-team's daily check in nudge.

handbook/engineering/leadership/index.md
```diff
@@ -69,6 +69,18 @@ Our long term vision is to build an organization that attracts and retains the b
...
+
+Engineering managers send an update to engineering-leadership-status@sourcegraph.com by 10am PT every Monday (sending it earlier, like EOD Friday is ok too). The purpose of the update is to keep interested parties (including the VP of Engineering, and peer managers) informed of what is going on in each part of our organization. The format of this update is at the discretion of the sender, but it should be prose that generally answers the following questions:
+
+- What important things happened last week?
+- What are the plans for this week?
+- What issues should my peer-team/manager be aware of?
```
chayim, line 80: I'd like to propose adding the additional question (or similar):

- What if anything can we improve upon in the upcoming week?

chayim (approved):
These changes make sense overall - I have some small questions located herein.

handbook/engineering/leadership/index.md
chrispine: If it's a thing you _are going to_ improve, then it feels like it should go into "plans for this week". If it's a thing you'd like us to discuss, then it doesn't feel like a status update, and probably belongs in our weekly sync.

Do you have a specific thing in mind that you think should go into a new section?

chrispine, line 80:
```suggestion
- What should my peer-team/manager be aware of?
```

handbook/engineering/leadership/index.md
nicksnyder: I think each EM can be responsible for setting their own reminders (Google Cal event, `/remind me Post weekly update at 9am PT` in Slack DM channel to yourself).

handbook/engineering/leadership/index.md
nicksnyder: I think it is valuable to document the time here (whatever it may be) because it can be helpful for future teammates (e.g. future web team manager) to know. It is easy to update the docs here if we update the meeting day/time on a recurring basis. This doesn't prevent us from rescheduling ad hoc on any given week.

handbook/engineering/leadership/index.md
chrispine: Thanks :-)

handbook/engineering/leadership/index.md
nicksnyder: I agree with Chris, and will also add that if something was painful and you want it to improve, expressing that pain would fit into "What should my peer-team/manager be aware of?"

I tried to make it clear in the docs that I do not expect the updates to answer exactly and only these questions, so the list of questions here isn't that important, just a guide.

chayim: I guess it could fall into the _What issues should my peer-team/manager be aware of?
_.  There's a reality that I always feel that if explicit improvements aren't called out as a target, I feel that they end up being buried in the day-to-day.

chrispine: I do agree with you there. But I'm not sure a status update is the place for it. In my 1:1s, two of my standard questions are "What could have gone better since our last 1:1?" (after asking what went particularly well since our last 1:1), and "What could I be doing more or less of?" The first is oriented toward personal/team improvements for the engineer, and the second for my own improvement.

I'm sure there are other places we could bake this into our routines. I'd like to know what's worked well for you in the past, but maybe we should take this conversation out of Github :-)

lguychard (approved):

pecigonzalo (approved):

handbook/engineering/leadership/index.md
```diff
@@ -69,6 +69,18 @@ Our long term vision is to build an organization that attracts and retains the b
 - Owner: TBD.
 - Status: Not started. Needs more discussion.
 
+## Communication
+
+We have a 1hr sync with product and engineering every Tuesday at 9am PT ([meeting notes](https://docs.google.com/document/d/10fEh3Cw0ENKjFYDP-4OoLZbOZTVAtMpifc2WM_1mGVU/edit)).
+
+Engineering managers send an update to engineering-leadership-status@sourcegraph.com by 12pm (noon) PT every Monday (sending it earlier, like EOD Friday is ok too). The purpose of the update is to keep interested parties (including the VP of Engineering, and peer managers) informed of what is going on in each part of our organization. The format of this update is at the discretion of the sender, but it should be prose that generally answers the following questions:
```
nicksnyder, line 76:
```suggestion
Engineering managers send an update to [engineering-leadership-status@sourcegraph.com](https://groups.google.com/a/sourcegraph.com/g/engineering-leadership-status) by 12pm (noon) PT every Monday (sending it earlier, like EOD Friday is ok too). The purpose of the update is to keep interested parties (including the VP of Engineering, and peer managers) informed of what is going on in each part of our organization. The format of this update is at the discretion of the sender, but it should be prose that generally answers the following questions:
```

Issue #1483, opened 2020-08-28
Changes to handbook css aren't deployed


nicksnyder:
I merged https://github.com/sourcegraph/about/pull/1475, but the CSS in not in production.

I see the updated CSS when running `make serve` locally.

nicksnyder:
It looks like it was just cached for 1 hour.

PR #1485, opened 2020-08-28
Bust the cache

aileenrose:
Addressing:  https://github.com/sourcegraph/about/issues/1483


nicksnyder:
oh, it works now. The cache looks to be 1 hour (I thought I hard refreshed, but maybe not). We don't need to merge this to close the issue.

PR #1486, opened 2020-08-29
Remove arbitrary 'asdf' wordings from the footer of the podcast page

joshuabezaleel:
I found that there was a `asdf` word written at the bottom of the [podcast page](https://about.sourcegraph.com/podcast/), which I presume is placed unintentionally. Should my assumption turn out to be wrong, I will delete this particular PR. Thank you so much for the amazing podcast and looking forward to more of its episodes! 🙂 
Screen Shot 2020-08-29 at 10 24 36 AM


sqs:
Haha, thanks for catching this!

PR #1521, opened 2020-09-04
Add CE to Eng handover docs

dadlerj:
Given the recent discussion, I wanted to get this written down. This is a WIP document, and will be living for some time, especially as more clarity comes to the roles and responsibilities of the CE team. However, I thought it would be beneficial to document the "current state", even if it's not long-lived.

I only added one "new" thing to the process here: a priority tagging system that CE can use to communicate to Eng. Curious to hear your thoughts.

CC @nicksnyder @christinelovett @tistru for visibility

### DIFF
--- a/handbook/ce/ce_to_eng_handover.md
+++ b/handbook/ce/ce_to_eng_handover.md
@@ -1,0 +1,35 @@
+# CE-to-Engineering issue handover process
+
+When a prospect or customer files a new support ticket or raises a new issue, CE and Engineering should follow the steps below to ensure the issue is handled successfully.
+
+## Principles
+
+**CE is always the first line of defense.**
+
+**Engineering should only feel a responsibility to get involved if tagged in by CE.**
+
+**However, once someone is "assigned" to a ticket (whether formally, or they informally take over the conversation), it is up to them to either (1) see the issue through to resolution or (2) assign a new owner.**
+
+Exceptions to the principles above:
+
+- This does not apply to our public GitHub issue tracker; instead, it only applies to [official support channels](support.md). Issues filed in GitHub are the responsibility of Engineering and/or Product.
+- Certain customers pay for dedicated support from a member of the Engineering team. Responding to issues filed by these customers is a shared responsibility for the assigned Engineer and CE (whoever sees it first).
+- If an engineer sees a new question or issue come in from a company that they've already been introduced to, or if the question is in their direct area of expertise, they are encouraged to jump in directly.
+
+Otherwise, if an engineer sees a new question or issue come in and they are interested in responding, they should just check with CE first (by posting in the #ce Slack channel).
+
+## How CE hands issues to Engineering
+
+1. The assigned CE will first reply to the prospect or customer that reported the issue.
+1. If the issue is clearly a bug or a feature request (rather than a question that can be clarified or answered on the spot), [the CE will file or add on to a GitHub issue](customer_issues.md).
+1. The CE will add a prioritization label to the issue, from `user/p0` to `user/p4`, based on a combination of (1) the severity of the issue, and (2) the prioritization of the reporting company. These labels mean the following:
+  1. `user/p0`: The issue results in the company's Sourcegraph instance being unusable and the company is a [Tier 1 prospect or customer](../sales/index.md#segmentation).
+  1. `user/p1`: The issue results in partial loss of functionality or serious disruption and the company is a [Tier 1 or Tier 2 prospect or customer](../sales/index.md#segmentation).
+  1. `user/p2`: The issue results in minimal loss of functionality or annoyance and the company is a [Tier 1 or Tier 2 prospect or customer](../sales/index.md#segmentation), or the issue is more serious, but applies to a [Tier 3 prospect or customer](../sales/index.md#segmentation).
+  1. `user/p3`: The feature request is a "nice-to-have" for a [Tier 1 or Tier 2 prospect or customer](../sales/index.md#segmentation), or the issue results in minimal loss of functionality or annoyance for a [Tier 3 prospect or customer](../sales/index.md#segmentation).
+  1. `user/p4`: The feature request is not urgent or the issue is minor.
+
+## Engineering responsibilities
+
+1. If an Engineer agrees to take on an issue or a ticket, they must be willing to follow-through on the problem until it is addressed. If they are not willing or able to do so, they must notify the CE as soon as possible so someone else can be assigned.
+1. If an Engineer is unsure of how to respond, they should reach out to CE in the #ce channel for feedback and support.
--- a/handbook/ce/ce_to_eng_handover.md
+++ b/handbook/ce/ce_to_eng_handover.md
@@ -8,7 +8,7 @@ When a prospect or customer files a new support ticket or raises a new issue, CE
 
 **Engineering should only feel a responsibility to get involved if tagged in by CE.**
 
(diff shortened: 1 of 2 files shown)


handbook/ce/ce_to_eng_handover.md
```diff
@@ -0,0 +1,35 @@
...
+
+When a prospect or customer files a new support ticket or raises a new issue, CE and Engineering should follow the steps below to ensure the issue is handled successfully.
+
+## Principles
+
+**CE is always the first line of defense.**
+
+**Engineering should only feel a responsibility to get involved if tagged in by CE.**
+
+**However, once someone is "assigned" to a ticket (whether formally, or they informally take over the conversation), it is up to them to either (1) see the issue through to resolution or (2) assign a new owner.**
```
nicksnyder, line 11: Can we change the wording so it is clear that if someone "informally" takes over the conversation, then they should be formally assigned the ticket? I want to avoid grey areas on who is responsible.

handbook/ce/ce_to_eng_handover.md
```diff
@@ -0,0 +1,35 @@
...
+
+Otherwise, if an engineer sees a new question or issue come in and they are interested in responding, they should just check with CE first (by posting in the #ce Slack channel).
+
+## How CE hands issues to Engineering
+
+1. The assigned CE will first reply to the prospect or customer that reported the issue.
+1. If the issue is clearly a bug or a feature request (rather than a question that can be clarified or answered on the spot), [the CE will file or add on to a GitHub issue](customer_issues.md).
+1. The CE will add a prioritization label to the issue, from `user/p0` to `user/p4`, based on a combination of (1) the severity of the issue, and (2) the prioritization of the reporting company. These labels mean the following:
+  1. `user/p0`: The issue results in the company's Sourcegraph instance being unusable and the company is a [Tier 1 prospect or customer](../sales/index.md#segmentation).
+  1. `user/p1`: The issue results in partial loss of functionality or serious disruption and the company is a [Tier 1 or Tier 2 prospect or customer](../sales/index.md#segmentation).
```
nicksnyder, line 27: So a tier 2 customer can never cause a p0?

dadlerj: This is just a quick and dirty set of definitions though, and Julia needs to weigh in. The definition of what the eng team is supposed to do when a p0 lands is also not clear here, so Gonza should weigh in on that side! This is just a first attempt at providing guidance to new CE team members based on how I internally think about communicating prioritization currently.

pecigonzalo (commented):
Thanks for taking this initial draft Dan. I will start working on the incident response updates and backlink here from the PR.

handbook/ce/ce_to_eng_handover.md
```diff
@@ -0,0 +1,35 @@
...
+
+**Engineering should only feel a responsibility to get involved if tagged in by CE.**
+
+**However, once someone [takes ownership](support.md#support-owners) of a ticket (whether formally, or informally by taking over the conversation), it is their responsibility to see it through, either by (1) see the issue through to resolution or (2) assign a new owner.**
+
+Exceptions to the principles above:
+
+- This does not apply to our public GitHub issue tracker; instead, it only applies to [official support channels](support.md). Issues filed in GitHub are the responsibility of Engineering and/or Product.
+- Certain customers pay for dedicated support from a member of the Engineering team. Responding to issues filed by these customers is a shared responsibility for the assigned Engineer and CE (whoever sees it first).
+- If an engineer sees a new question or issue come in from a company that they've already been introduced to, or if the question is in their direct area of expertise, they are encouraged to jump in directly.
```
juliasourceress, line 17: An engineer should always check in with a CE before replying if it's not a conversation they're already having (but a new thread or ticket). If it's an existing thread the engineer should be the one to triage and either continue responding or let the CE owner know that despite it being the same ticket or thread, it's a new request, and they'd like to hand it back over to the CE for ownership.

handbook/ce/ce_to_eng_handover.md
```diff
@@ -0,0 +1,35 @@
...
+  1. `user/p1`: The issue results in partial loss of functionality or serious disruption and the company is a [Tier 1 or Tier 2 prospect or customer](../sales/index.md#segmentation).
+  1. `user/p2`: The issue results in minimal loss of functionality or annoyance and the company is a [Tier 1 or Tier 2 prospect or customer](../sales/index.md#segmentation), or the issue is more serious, but applies to a [Tier 3 prospect or customer](../sales/index.md#segmentation).
+  1. `user/p3`: The feature request is a "nice-to-have" for a [Tier 1 or Tier 2 prospect or customer](../sales/index.md#segmentation), or the issue results in minimal loss of functionality or annoyance for a [Tier 3 prospect or customer](../sales/index.md#segmentation).
+  1. `user/p4`: The feature request is not urgent or the issue is minor.
+
+## Engineering responsibilities
+
+1. If an Engineer agrees to take on an issue or a ticket, they must be willing to follow-through on the problem until it is addressed. If they are not willing or able to do so, they must notify the CE as soon as possible so someone else can be assigned.
```
juliasourceress, line 34: Or alternatively they can find someone else to tap in, and let the CE know.

handbook/ce/ce_to_eng_handover.md
juliasourceress: This is a great and necessary convo, and the goal of this initial PR is to document how we do this today. So for now I support @dadlerj 's documented process as a start, since this is how he's been operating. A clear next step is to iterate on this based on @pecigonzalo 's suggestions.

juliasourceress (approved):

handbook/ce/ce_to_eng_handover.md
dadlerj: Thanks Julia! Agreed with that approach.

To be transparent, I'm not _directly_ using this methodology yet, but when I started to document my process, this is roughly how I think about my own method for communicating prioritization of issues. This just adds an official tag onto it (versus my handling in a one-off way). I'll merge this now and get feedback from Tion and Christine once it's live.

@pecigonzalo thanks for the feedback. A few notes:

>I believe this list is an extension or update to the one described here https://github.com/sourcegraph/about/blob/main/handbook/ce/support.md#slas.

It's actually quite different... That page reflects the promises we make to customers—i.e., the minimum service level required per our contract—while this list reflects our internal prioritization. There clearly is an intimate connection between the two, but they're not quite the same in terms of what they specify and how we want to describe them (and never quite will be). 

>Here is a draft idea on how we could handle this, I would PR this to incident document:

This is a great conversation—"what do we do about an issue that was described as a given level of seriousness by CE"—but it is slightly different from what this doc is adding. I'd like to start with our definition of how serious an issue is, and you (and Julia and others of course) can nail down the translation of this into the "so what".

pecigonzalo: > It's actually quite different... That page reflects the promises we make to customers—i.e., the minimum service level required per our contract—while this list reflects our internal prioritization. There clearly is an intimate connection between the two, but they're not quite the same in terms of what they specify and how we want to describe them (and never quite will be).

I disagree, they are not that different given the agreed response and resolution times are directly related to the communicated prioritization as `ce/pX` and how we respond to that priority internally. The intended audience is different, but even the description wording on both is quite similar, only more detailed on this document as we have more levels.
We could provide this by having a single matrix, a linked relationship between the two or just the same name across both pages but holding different information to match each intended audience.

tistru:
Hi @dadlerj . Great work putting this together and I believe it will really helpful. I didn't have any feedback on the current version of the document. I understand we will iterate on it going forward but I am perfectly fine to operate under these guidelines going forward :)

PR #1531, opened 2020-09-08
Add "Email to Salesforce" option to handbook

dadlerj:
Add option to log emails without using Einstein. On the Ops, HR, Legal, Finance, etc. side I have a lot of sensitive emails that I'd prefer not to be auto-logged into Salesforce.


ebrodymoore (approved):
Thanks Dan!

PR #1541, opened 2020-09-09
Demo instance goal

nicksnyder:
The lack of a dedicated demo instance slowed down the Cloud team this month, so I am proposing adding this goal for the distribution team.

### DIFF
--- a/handbook/engineering/distribution/goals.md
+++ b/handbook/engineering/distribution/goals.md
@@ -9,7 +9,7 @@ Goals are continuously updated and reviewed. If you find these goals do not refl
 - **Problem:** Customer engineering needs to demo Sourcegraph to customers. Many things can be demoed on Sourcegraph.com, but some things can't. For those things, CE has been using our dogfood instance sourcegraph.sgdev.org. Engineers need to use our dogfood instance for testing, which either puts demos at risk, or delays testing when demos happen.
 - **Plan:** Create demo.sourcegraph.com that CE uses for demos. This can be a small instance with a limited number of repos that is specifically designed for demo paths. sourcegraph.sgdev.org will continue to be our large scale dogfood instance for engineering testing.
 - **Owners:** TBD
-- **Status:** Not started
+- **Status:** https://github.com/sourcegraph/sourcegraph/issues/13604
 
 ### Any engineer at Sourcegraph can create a release for all of our supported deployment types by running a single command
 
--- a/handbook/engineering/distribution/goals.md
+++ b/handbook/engineering/distribution/goals.md
@@ -4,6 +4,13 @@ Goals are continuously updated and reviewed. If you find these goals do not refl
 
 ## Medium-term goals
 
+### Create Sourcegraph demo instance
+
+- **Problem:** Customer engineering needs to demo Sourcegraph to customers. Many things can be demoed on Sourcegraph.com, but some things can't. For those things, CE has been using our dogfood instance sourcegraph.sgdev.org. Engineers need to use our dogfood instance for testing, which either puts demos at risk, or delays testing when demos happen.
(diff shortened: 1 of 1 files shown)


slimsag:
I don’t think this needs to be a goal, that seems to imply this is a larger undertaking than it actually is - there is an issue open to do this and I expect to have it done as part of 3.21, if not sooner.

pecigonzalo:
Looping the issue link in https://github.com/sourcegraph/sourcegraph/issues/13604

nicksnyder:
Glad to see this is already an issue (I didn't know that). I will leave it up to the team on whether to also track this as a goal or not.

> I expect to have it done as part of 3.21, if not sooner.

Can we add that to the 3.21 milestone to communicate that (it is ok if it happens earlier).

handbook/engineering/distribution/goals.md
```diff
@@ -4,6 +4,13 @@ Goals are continuously updated and reviewed. If you find these goals do not refl
 
 ## Medium-term goals
 
+### Create Sourcegraph demo instance
+
+- **Problem:** Customer engineering needs to demo Sourcegraph to customers. Many things can be demoed on Sourcegraph.com, but some things can't. For those things, CE has been using our dogfood instance sourcegraph.sgdev.org. Engineers need to use our dogfood instance for testing, which either puts demos at risk, or delays testing when demos happen.
+- **Plan:** Create demo.sourcegraph.com that CE uses for demos. This can be a small instance with a limited number of repos that is specifically designed for demo paths. sourcegraph.sgdev.org will continue to be our large scale dogfood instance for engineering testing.
+- **Owners:** TBD
+- **Status:** Not started
```
nicksnyder, line 12:
```suggestion
- **Status:** https://github.com/sourcegraph/sourcegraph/issues/13604
```

slimsag:
I guess my point was goals feels like it should be fairly high-level: a 15-30m task doesn't really fit that category in my view, otherwise most issues we work on should also become goals in this doc.

i digress, though, I'm happy to have this here on merit of "it feels big" and check that off as a win quickly.

slimsag:
Perhaps one way to rephrase this is that it has always been a goal, just not documented, and that managed instances has enabled us to do this in a way where we treat the CE/Sales team as a "customer" of Sourcegraph?

handbook/engineering/distribution/goals.md
```diff
@@ -4,6 +4,13 @@ Goals are continuously updated and reviewed. If you find these goals do not refl
 
 ## Medium-term goals
 
+### Create Sourcegraph demo instance
+
+- **Problem:** Customer engineering needs to demo Sourcegraph to customers. Many things can be demoed on Sourcegraph.com, but some things can't. For those things, CE has been using our dogfood instance sourcegraph.sgdev.org. Engineers need to use our dogfood instance for testing, which either puts demos at risk, or delays testing when demos happen.
+- **Plan:** Create demo.sourcegraph.com that CE uses for demos. This can be a small instance with a limited number of repos that is specifically designed for demo paths. sourcegraph.sgdev.org will continue to be our large scale dogfood instance for engineering testing.
+- **Owners:** TBD
```
pecigonzalo, line 11:
```suggestion
- **Owners:** Robert, Gonza
```

pecigonzalo (approved):
I believe that is the case, we are also working on this because `manage-instances` code enabled this to be done fairly easily and the work on dogfooding has enabled that cluster to work as intended, upgrading more frequently, which is a problem for the way it was used before for demos.

christinaforney:
It looks like this will be added as a goal. Perhaps as a later follow up after complete is to clarify the infrastructure area of ownership for the distribution team to be more specific about the instances maintained. For example:

- Internal Sourcegraph instances, including:
   1. A demo instance that is kept up-to-date and is stable for CE and Sales to use
   1. A dogfood instance for the engineering team to try out features in a Server environment (vs. Cloud)

christinaforney (approved):

nicksnyder:
we have a demo instance now, so closing this

PR #1543, opened 2020-09-09
Update hiring.md

withdavidli:
Updating information about stages in our ATS.

### DIFF
--- a/handbook/people-ops/hiring.md
+++ b/handbook/people-ops/hiring.md
@@ -62,6 +62,37 @@ For any members of a hiring team, please see [this folder](https://drive.google.
 
 See [interview process](interview_process.md).
 
+## Stages in our application tracking system
+
+| Stage name  |  Why a candidate is in this stage |
+|---|---|
+|  New lead |  This is the first stage for a sourced candidate. | 
+|  Reached out | When we have contacted a sourced candidate.  | 
+|  New applicant | This is the first stage for a candidate that applied to the role directly or referred to the role. | 
+|  Application review |  Candidates that the hiring manager should review and decide on next steps.  | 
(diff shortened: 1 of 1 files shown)


handbook/people-ops/hiring.md
```diff
@@ -62,6 +62,37 @@ For any members of a hiring team, please see [this folder](https://drive.google.
...
+|  New lead |  This is the first stage for a sourced candidate. | 
+|  Reached out | When we have contacted a sourced candidate.  | 
+|  New applicant | This is the first stage for a candidate that applied to the role directly or referred to the role. | 
+|  Application review |  Candidates that the hiring manager should review and decide on next steps.  | 
+|  Recruiter screen | Candidates that the recruiter has scheduled to screen. | 
+|  Initial screen |  First call with the hiring team. | 
+|  Assignment |  Candidates that we have scheduled to complete an assignment, or waiting for a decision after they have completed the assignment.  | 
+|  Interview |  Candidates that are in rounds of interviews after the “initial screen” and “assignment” stages.| 
+|  Debrief |  Candidates that have completed the “interview stage” and interview panel need to discuss whether we should extend an offer to the candidate. | 
+|  Reference check |  Candidates that we have decided to make an offer to. |
```
christinaforney, line 78: Something that is new-ish that we might want to change inside of Lever is that reference checks are now done during the first week of the person starting (the offer is not delayed on reference calls).

mercadon (approved):
This looks great- we should share this with all hiring managers and Cecily so we all know best practices for moving candidates through stages.

dadlerj (approved):
Great addition. If there are any quick tweaks needed since you first created this, feel free to update, but otherwise definitely better to have this merged than not!

PR #1544, opened 2020-09-09
Update engineering-manager-web.md

withdavidli:
Hi Nick,

Is the goal to move all our github documentation to the handbook?

See this job description has directions for what we do internally, and that may confuse the candidates. Following edits recommended. 

Changing the language "pick" to "will interview with" as not to confuse candidates that they get to pick who to meet. 

Technical experience github is private. Deleting the link for now and replaced with general description from other job descriptions for now.

Adding an apply here link at the top instead of just middle of the page.

PR #1545, opened 2020-09-09
Update engineering-manager-web.md

withdavidli:


nicksnyder (approved):
Was there a reason why you created this PR instead of updating https://github.com/sourcegraph/about/pull/1544?

Changes here look good, but it is missing the other wording changes from the previous PR (e.g. "Pick two:" -> "Will interview with 2 of the following:")

PR #1549, opened 2020-09-10
sourcegraph version in getting started

uwedeportivo:
need to fix deployment script to cover this about page. in the meantime here's a manual correction


aileenrose (approved):
Yahoo!!

PR #1562, opened 2020-09-12
Avoid duplicate title errors with complete titles

aileenrose:
Updated titles on onboarding and goals pages.

### DIFF
--- a/company/goals/index.md
+++ b/company/goals/index.md
@@ -1,4 +1,4 @@
-# Goals
+# Company goals
 
 Our company goals are public.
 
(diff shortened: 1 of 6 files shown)


uwedeportivo (approved):

nicksnyder (changes requested):
What problem does this solve?

Pages should be able to choose whatever titles they choose, duplication shouldn't be a problem.

aileenrose:
> What problem does this solve?
> 
> Pages should be able to choose whatever titles they choose, duplication shouldn't be a problem.

Even though these pages are directed to a new staff, as a best practice, pages should have unique titles for SEO.   SEO tools like Google Search Console will flag.   Also, great to show consistent titles in search results.

Screen Shot 2020-09-14 at 9 31 55 AM

nicksnyder (approved):

Issue #1468, opened 2020-08-28
on-call: document actions to follow up on critical alerts


bobheadxi:
It's not currently very clear what to do with alerts coming through #opsgenie - for example, blob load latency alerts have been firing frequently (https://github.com/sourcegraph/sourcegraph/issues/12942) without follow-up

Earlier today, a GitHub rate limit alert went off and was interpreted as a false-positive, which later turned out to be a real problem: https://sourcegraph.slack.com/archives/CMBA8F926/p1598615920060100

The current on-call page (https://about.sourcegraph.com/handbook/engineering/on_call) is a bit out of date and (IMO) could use a bit more detail relevant to how alerts work today - specifically, actions like:

* silencing via PR to deploy-sourcegraph-dot-com + issue tracking investigation into the alert
* adjusting alerts via a PR to sourcegraph (and/or reaching out to Distribution for help on how to better encapsulate the behaviour we want to alert on)
* improving `PossibleSolutions` with actual solutions we take, and/or linking to relevant context (https://github.com/sourcegraph/sourcegraph/issues/13436)

bobheadxi:
I think I'll open an RFC for this

bobheadxi:
I'm curious if there's a need for this anymore, the frequency of critical alerts has gone down significantly and in general everyone jumps on board pretty quickly to resolve issues and adjust alerts.

Keeping assigned, but not prioritized

bobheadxi:
Superseded by @ggilmore's work on https://github.com/sourcegraph/sourcegraph/issues/12138

PR #1568, opened 2020-09-14
Add Escalation Policy

daxmc99:
From an action item on the [ops-log](https://docs.google.com/document/d/1dtrOHs5STJYKvyjigL1kMm6u-W0mlyRSyVxPfKIOfEw/edit?disco=AAAAKQbBzN4)

### DIFF
--- a/handbook/engineering/incidents/index.md
+++ b/handbook/engineering/incidents/index.md
@@ -46,6 +46,7 @@ The goal of triage is to either quickly resolve the incident using basic procedu
     - The on-call engineer.
         - You can find out who is on-call by typing `/genie whoisoncall` in Slack.
         - If you are not able to immediately get in contact with the on-call engineer, then manually create a new OpsGenie alert by typing `/genie alert "description of incident and link to Slack thread" for ops_team`.
+        - For more information about our on-call procedures, see our [on-call](on_call.md) docs.
 
 ## Go-to-market (license and subscription) issues
 
--- a/handbook/engineering/incidents/index.md
+++ b/handbook/engineering/incidents/index.md
@@ -19,10 +19,13 @@ Incidents can be reported by anyone (e.g. customers, Sourcegraph teammates) by a
 2. Start an internal communication thread about this incident in the #dev-ops channel in Slack.
     - All subsequent communication about this issue should happen in that Slack _thread_ (not in the top level #dev-ops channel).
 3. Identify an engineer to triage the incident.
-    - If you are an engineer and available for 30 minutes, then you should [triage the incident](#triage).
-    - If you are not an engineer or are not available to triage the incident, then ask the on-call engineer to triage the incident.
+    1. If the initial alert came from our alerting infrastructure, then the alert should be routed to the responsible [team](https://sourcegraph.com/search?q=repo:github.com/sourcegraph/sourcegraph+file:monitoring/frontend.go+ObservableOwner&patternType=literal)
+
+    1. If you are an engineer and available for 30 minutes, then you should [triage the incident](#triage).
+    1. If you are not an engineer or are not available to triage the incident, then ask the on-call engineer to triage the incident.
         - You can find out who is on-call by typing `/genie whoisoncall` in Slack.
         - If you are not able to immediately get in contact with the on-call engineer, then manually create a new OpsGenie alert by typing `/genie <description of incident and link to Slack thread> with ops_team`.
+
 
 ## Triage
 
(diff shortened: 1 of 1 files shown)


daxmc99:
I am looking to iterate on this. I want to make it clear there are _very_ few cases when something would be needed to be escalated outside a mission team.

handbook/engineering/incidents/index.md
```diff
@@ -19,11 +19,14 @@ Incidents can be reported by anyone (e.g. customers, Sourcegraph teammates) by a
 2. Start an internal communication thread about this incident in the #dev-ops channel in Slack.
     - All subsequent communication about this issue should happen in that Slack _thread_ (not in the top level #dev-ops channel).
 3. Identify an engineer to triage the incident.
-    - If you are an engineer and available for 30 minutes, then you should [triage the incident](#triage).
-    - If you are not an engineer or are not available to triage the incident, then ask the on-call engineer to triage the incident.
+    1. If the initial alert came from our alerting infrastructure, then the alert should be routed to the responsible [team](https://sourcegraph.com/search?q=repo:github.com/sourcegraph/sourcegraph+file:monitoring/frontend.go+ObservableOwner&patternType=literal)
+
+    1. If you are an engineer and available for 30 minutes, then you should [triage the incident](#triage).
+    1. If you are not an engineer or are not available to triage the incident, then ask the on-call engineer to triage the incident.
```
nicksnyder, line 25: This isn't really an ordered list, it is an enumeration of options, so I think unordered is better.

```suggestion
    - If the initial alert came from our alerting infrastructure, then the alert should be routed to the responsible [team](https://sourcegraph.com/search?q=repo:github.com/sourcegraph/sourcegraph+file:monitoring/frontend.go+ObservableOwner&patternType=literal)

    - If you are an engineer and available for 30 minutes, then you should [triage the incident](#triage).
    - If you are not an engineer or are not available to triage the incident, then ask the on-call engineer to triage the incident.
```

handbook/engineering/incidents/index.md
```diff
@@ -61,6 +64,24 @@ The owner of the incident may delegate tasks to other available/working engineer
...
+2. Mission team responsible for this service
+
+_most escalation should stop here_
+
+3. VP of Engineering
+   1. (Optional) If this a security incident then the security-team should be notified as well here
+4. CTO
+5. CEO
+
+
```
nicksnyder, line 84: I don't think we need a general escalation policy right now because the most important policy is encoded in how each team handles their oncall rotation.

daxmc99: @pecigonzalo do you think its worth having a general escalation policy? I think updating this doc to the current state of team owned alerts is enough to satisfy the action item.

pecigonzalo: I think what is missing is how to escalate beyond the team. For example, for the incident on Saturday, we were unable to find what was causing the failure and would have been to have a known escalation path. 
It was lucky that it was just `/pricing` and `/about` but if this was `sourcegraph.com`, it would have been a bigger issue.

In this example, I was unable to find were an IP was served from and needed to get someone online that might know were that was set, for example Stephen, Beyang, Quinn, Nick, as you some of these details are not documented.

The questions would have been:
- Who should I call in that case?
- How should I call them?
- Should I escalate for this?

I think we used to have a SWAT team, and I would propose we add back a similar thing and add those who might know those questions (or who to ask) in that group. It should not be linked to an alert, but rather manually triggered by an engineer if in this situation.

daxmc99: I think you were still able to find the IP, though? I agree to have someone with more knowledge of this would have been helpful, but I'm not sure involving more people would have reduced the overall time to resolution. 

For your questions:
- Who should I have called? I believe notifying all on-call engineers was sufficient. 
- Via opsgenie ( `/genie whoisoncall` & `/genie alert "description of incident and link to Slack thread`) 
- I would argue if you have done both of the above you have escalated the issue to a reasonable degree. 

I would worry the SWAT team would become a permanent on-call team. That seems to go against the theme of this document which promotes containment of the unplanned work.

pecigonzalo: > I think you were still able to find the IP, though?

We found the IP, we could not find what was serving the content behind the IP, it took us 3hours to find it in a comment in code.

> Via opsgenie ( /genie whoisoncall & /genie alert "description of incident and link to Slack thread)

That would work, but there is no group to notify other engineers that would have known this, notifying other engineering teams would have not made sense.

> I would argue if you have done both of the above you have escalated the issue to a reasonable degree.

There was no other escalation to reach, when I started troubleshooting the notification had already gone through the entire escalation for distribution.

>I would worry the SWAT team would become a permanent on-call team. That seems to go against the theme of this document which promotes containment of the unplanned work.

They would be the last escalation point, if they need to be pulled in constantly, then something is missing in our documentation or processes that on-call engineers are unable to fix the problems they are on-call for, as it was the case here. Each team is escalation for the person on-call if they cant answer, most alerts should not reach this point.
The problem is if this was sourcegraph.com I would have not known what to do. Im already the the last escalation of my rotation and I cant find a fix our source of the problem, should I have called someone? else and if so, who? 
The alternative is let the site be down until Monday, if that is acceptable, then we can document that.

daxmc99:
This is starting to get stale. I think this might need input from more teams possibly in an RFC.  
Moving to closed. 
Happy to reopen though

bobheadxi:
> I think this might need input from more teams, possibly in an RFC.

I'll incorporate this into [RFC 231](https://docs.google.com/document/d/1FvNGmedbgr-KL_LAb5l4PWyPUzPPQGpzNZAdssrpJrY/edit) @daxmc99 ! Been meaning to write something up (https://github.com/sourcegraph/about/issues/1468)

PR #1573, opened 2020-09-15
Add more clarification between PDs and RFCs

quinnkeast:


christinaforney (approved):
Nice improvement, thanks!

PR #1577, opened 2020-09-15
Fine-tune typography on blog posts

quinnkeast:
This is a quick set of improvements to typography and spacing for the blog posts. Changes only target the blog posts themselves. This will make our blog posts more readable for all users.

`gatsby-remark-images` has a nasty habit of wrapping `` elements in `<p>` tags, which causes trouble with having a narrower paragraph elements without affecting the images. I've added `gatsby-remark-unwrap-images` to the `gatsby-config.js` to avoid this. As far as I can tell, this only affects the blog posts, which is what we want.

I've also made a couple of tiny formatting tweaks to @efritz 's latest blog post to un-hacky-workaround the `blockquote` formatting and make the linked image render properly. I checked through a handful of older blog posts and confirmed nothing appears broken.


sqs:
Overall looks great. Looks a bit weird with the podcast component at https://deploy-preview-1577--sourcegraph.netlify.app/podcast/michael-stapelberg/. Will check more soon.

quinnkeast:
Agreed. I'll update it shortly so that the audio component renders like this:

Screenshot 2020-09-15 at 20 35 16

sqs (approved):
Looks great!

christinaforney (approved):
Nice! Looks great!

PR #1582, opened 2020-09-16
Add alert channes to engineering onboarding

pecigonzalo:


bobheadxi (approved):
LGTM! Another niche one is `#alerts-managed-instances`, but not sure if that's something everyone should look at

pecigonzalo:
Yeah, that was my thought about it, we could add those under our distribution onboarding doc.

PR #1587, opened 2020-09-16
Adding backend infrastructure goals to Cloud team

christinaforney:
WIP: needs copy tweaking, clarification of goals/links to docs/issues.

### DIFF
--- a/handbook/engineering/cloud/index.md
+++ b/handbook/engineering/cloud/index.md
@@ -2,31 +2,52 @@
 
 The cloud team owns all work that is necessary to build, secure, scale, and operate our multi-tenant hosted version of Sourcegraph for customers that do not want to deploy Sourcegraph on-premise.
 
+The cloud team is also responsible for all [backend-infrastructure areas of ownership](../backend-infrastructure/index.md). A lot of the work for Sourcegraph Cloud is in direct support of scaling and performance of backend-infrastructure. Work will be balanced as needed to support both efforts.
+
+## Areas of ownership
+
+- Authorization and authentication
+- Repository management (gitserver, repo-updater, src-expose)
+- Data storage and access libraries
+- GraphQL API
+- src-cli
+- License generation and enforcement
+- Admin and user settings
+- Analytics
+
 ## Goals
+
+### Sourcegraph Cloud
 
 1. Any user or organization can use Sourcegraph Cloud with both the private and public code that they care about in a secure way.
 
-## Milestones
+#### Milestones
 
 1. Any user can add public code from github.com, gitlab.com and bitbucket.org
     - Code is indexed.
     - It's easy to search over only their code (vs all public code).
-2. Any user or organization can use Sourcegraph Cloud for free before it's GA
+1. Any user or organization can use Sourcegraph Cloud for free before it's GA
     - No plaintext tokens or secrets anywhere (including gitserver .git/config remotes).
     - All private repository content are only decryptable by Sourcegraph services.
         - Gitserver.
         - Searcher and code intel caches.
         - Any other service that needs access to repository contents needs to be able to decrypt what it needs on the fly.
-3. Sign-up and on-boarding of new users is fast, obvious and painless
+1. Sign-up and on-boarding of new users is fast, obvious and painless
     - Adding your repos with a nice UI flow that leverages the OAuth tokens from signing up with the code host, where you can select repositories easily.
     - Permissions and repo syncing progress is clear to the user and doesn’t take too long.
     - Progress on repo indexing is clear to the user and doesn’t take too long.
-4. Sourcegraph Cloud is Generally Available (GA)
+1. Sourcegraph Cloud is Generally Available (GA)
     - Abuse protection: API rate limiting, DDoS mitigation, limiting user accounts.
     - Scalable syncing of permissions, repos, changesets.
     - High availability, SLOs, etc.
     - Billing and subscriptions.
         - Based on size (i.e. GB) of all repositories and selected feature set.
+
+### Backend infrastructure
+
+1. Product tier enforcement
+1. Trial sign-up flow/license keys
+1. Non-Git VCS
 
 ## Contact
 
--- a/handbook/engineering/cloud/index.md
+++ b/handbook/engineering/cloud/index.md
@@ -23,20 +23,22 @@ The cloud team is also responsible for all [backend-infrastructure areas of owne
 
 #### Milestones
 
-1. Any user can add public code from github.com, gitlab.com and bitbucket.org
+1. The Sourcegraph organization and team members can add public code from GitHub.com, GitLab.com, and Bitbucket Cloud.
     - Code is indexed.
     - It's easy to search over only their code (vs all public code).
-1. Any user or organization can use Sourcegraph Cloud for free before it's GA
+1. Any user can add public code from GitHub.com, GitLab.com, and Bitbucket Cloud.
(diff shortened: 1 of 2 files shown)


tsenart:
@christinaforney: What's missing from this PR so that we can open it up for review?

christinaforney:
After our discussion in sprint planning this morning, the items we had listed under backend infrastructure felt more like roadmap items (vs. goals), so I have omitted them and added a sentence to clarify. 

I also looked more closely at the milestones and made some adjustments. Let me know what you all think and if you have any questions! Specifically, I broke out the difference between our team being able to perform an action, and then general users being able to perform an action. I think this will help make better decisions around next actions because we can focus on core functionality first, then iterate: the Sourcegraph team does not need polished UI or large scale improvements to get the right functionality in place for testing.

daxmc99 (approved):
LGTM!

asdine (approved):
👍🏼

handbook/engineering/cloud/index.md
```diff
@@ -2,32 +2,53 @@
...
+
 1. Any user or organization can use Sourcegraph Cloud with both the private and public code that they care about in a secure way.
 
-## Milestones
+#### Milestones
 
-1. Any user can add public code from github.com, gitlab.com and bitbucket.org
+1. The Sourcegraph organization and team members can add public code from GitHub.com, GitLab.com, and Bitbucket Cloud.
     - Code is indexed.
     - It's easy to search over only their code (vs all public code).
```
tsenart, line 28: @christinaforney: We added this previously because we think it's going to be important. There are multiple solutions here, but this feels like the domain of the search team. We can also punt this to the "make it usable" phase. Is there a good way for you to capture this work for that phase in the search team's roadmap?

tsenart (approved):

tsenart:
Also, @christinaforney, I think those guiding thoughts should definitely make it into the page itself. We could integrate them with the milestones themselves, and add that prose to each. What do you think?

ryanslade (approved):
LGTM

handbook/engineering/cloud/index.md
christinaforney: I agree this is is going to be important - and I think earlier in the process than later. As for punting it, perhaps having two versions of this: 1. it is possible to search over only your code (even if it's a janky implementation that our team can use), and then make it usable is more about having a clear way of knowing what you're searching over.

It's a good question, and one I would like to loop @lguychard into. I am wondering if it makes sense to put this on the search team, or if it makes sense for the Cloud team to own because it's a requirement to achieve the team's goals. I was previously assuming that the Cloud team would own some cross functional efforts, but if not, we should also talk about anything else that is required from other teams. Does this also mean that the web team should be planning to implement the UX for these efforts? Where do we draw the line and how do we make it clear to everyone?

christinaforney:
> Also, @christinaforney, I think those guiding thoughts should definitely make it into the page itself. We could integrate them with the milestones themselves, and add that prose to each. What do you think?

I updated to include a new Goals guiding principles section and added the items within the milestones as well. There are some pieces that are now unclear to me that I think would be helpful to write down. We can discuss in our meeting tomorrow!

handbook/engineering/cloud/index.md
tsenart: > As for punting it, perhaps having two versions of this (..) (even if it's a janky implementation that our team can use)

Yeah, I think that's possible. One idea: we tie in to repo groups, and create a "special" repogroup called `default` or `mine`, that refers to repos the user has added. Thoughts on this approach @sourcegraph/search?

>  Where do we draw the line and how do we make it clear to everyone

In my mind, if there's another team that owns the end product user experience of a particular area, then they need to be 100% aligned with or driving an effort in that area.

Examples: 
- Search team owns the search experience, so we need them to be 100% aligned with or driving the effort of making it easy for users to search over their public code.
- Cloud team owns admin and user settings, so we can drive changes to the user experience independently of other teams.

I'm happy to discuss this and align on whatever collaboration model we see fit, but this is how I see it.

handbook/engineering/cloud/index.md
```diff
@@ -2,31 +2,71 @@
...
+        - Code is indexed.
+        - It's possible to search over only their code (vs all public code).
+1. Any user can add public code from GitHub.com, GitLab.com, and Bitbucket Cloud.
+    - Make is usable:
+        - Adding your repos with a nice UI flow that leverages the OAuth tokens from signing up with the code host, where you can select repositories easily.
+        - Repo syncing progress is clear to the user and doesn’t take too long.
+        - Progress on repo indexing is clear to the user and doesn’t take too long.
+        - UX problems to solve (WIP):
+            - How do users know they are searching over their code vs. all Cloud code?
+            - How do we communicate status to users?
```
quinnkeast, line 45:
```suggestion
        - Connecting their first code host and adding their own public repos is intuitive and easy for new users.
        - Connecting with code hosts and and adding public repos is intuitive and easy for current users.
        - System activity and progress is easy to understand and doesn’t take too long.
          - Repo syncing
          - Repo indexing
        - Searching their own versus searching all Cloud code is intuitive.
```

handbook/engineering/cloud/index.md
```diff
@@ -2,31 +2,71 @@
...
+    - Make it work:
+        - No plaintext tokens or secrets anywhere (including gitserver .git/config remotes).
+        - All private repository content are only decryptable by Sourcegraph services.
+            - Gitserver.
+            - Searcher and code intel caches.
+            - Any other service that needs access to repository contents needs to be able to decrypt what it needs on the fly.
+1. Any user or organization can add private code to Sourcegraph Cloud for free before it's GA.
+    - Make it usable:
+        - Authorization from code hosts is enforced (e.g., organizations, teams).
+        - TODO
```
quinnkeast, line 58:
```suggestion
    - Make it usable:
        - Authorization from code hosts is enforced (e.g., organizations, teams).
        - Repository visibility and permissions on Sourcegraph is intuitive.
        - Adding private repositories is part of the same flows as adding public repositories.
        - TODO
```

handbook/engineering/cloud/index.md
christinaforney: Based on priorities/resources, it is likely the Cloud team will need to do the implementation, and I agree there needs to be close alignment/collaboration on planning/architecture with the search team.

lguychard: Having two versions sounds like a great idea:
- The Cloud team could execute on the "MVP" to unblock this goal.
- The Search team would be better equipped to tackle the end target ("users have a clear way of knowing / filtering what they're searching over, including filtering to search over their own code").

On the more general point:

> if there's another team that owns the end product user experience of a particular area, then they need to be 100% aligned with or driving an effort in that area

The distinction between "100% aligned" and "driving an effort" depends on the size of the effort needed. For small, self-contained efforts like the "special repogroup", aligning with the search team on the solution but having the Cloud team execute is probably the quickest way of moving forward.

It gets trickier when the effort is large enough that the other team would have to "drive the effort in that area" -- dependencies between teams are fragile, and I find it better for teams to be autonomously capable of making progress on their goals.

If a goal is truly large enough in scope that it would require effort from multiple teams, then it's a company goal -- we should then align _all teams_ around it, and structure/break up the execution strategy so that teams can make progress towards that goal without being mutually dependent on each other.

If a team finds itself blocked on making progress towards their goal for lack of a given skillset (eg. frontend skills), it would be more efficient to fill that gap through hiring or internal transfer than through delegation to another team.

Decoupling team priorities will keep teams focused on and engaged with business problems rather than organizational problems: "we're prioritizing this because it's the right thing to do to solve problem X" rather than "we're prioritizing this to unblock team Y".

PR #1588, opened 2020-09-16
use latest campaigns demo video

sqs:
Uses the newest video for campaigns (https://www.youtube.com/watch?v=EfKwKFzOs3E). See https://github.com/sourcegraph/sourcegraph/issues/13150.

cc @mrnugget


mrnugget:
Should we also use it on https://sourcegraph.com/campaigns? That's still showing the old video.

PR #1592, opened 2020-09-17
Add Campaigñeros logo

mrnugget:
We decided on this in today's sync. Once Erik is back we can do an even better logo.


LawnGnome (approved):
¡Si!

PR #1555, opened 2020-09-11
Move all job descriptions to handbook from careers repo

christinaforney:
- Creates company/careers.md from careers repo README (PR: https://github.com/sourcegraph/careers/pull/126)
- Moves all job descriptions from careers repo into <department>/roles/<job_descriptions>
- Add link "Go back to careers page for all open positions" to bottom of every open req.
- Creates "Roles" page for each 
- Creates people-ops/hiring and puts all hiring/interview related pages there
- Moves job description template and reference check questions from interviews repo to handbook.
- Changes about site careers footer link to go to company/careers.md, updates jobs page button to go there also.

PR #1150, opened 2020-07-03
handbook: Document 2021 Q1 Engineering org plan

tsenart:
This change set documents our current 2021 Q1 engineering org plan and will be kept up to date as it changes.

Any new feedback is always welcome! Please share it.

PR #1600, opened 2020-09-18
Add development tools team

pecigonzalo:
This still a draft which is 90% a copy of the Distribution role.

- [x] Requires: https://github.com/sourcegraph/about/pull/1555

#### Tasks
- [ ] Define vision and goals
- [ ] Define hiring requirements and interview process

/cc @nicksnyder


nicksnyder:
the diff here includes some things that aren't what you made

pecigonzalo:
@nicksnyder Its intentional, this requires #1555 which moves the JDs here.

handbook/engineering/developer-tools/index.md
```diff
@@ -0,0 +1,15 @@
+# Developer tools team
```
nicksnyder, line 1: I think the right way to document this is not to document a new team that doesn't exist, but instead add a goal to the distribution team and document this in the distribution team's growth plan (e.g. we will grow this team and plan on spinning off a dev tools team in the future.

Examples:
https://about.sourcegraph.com/handbook/engineering/web#growth-plan
https://about.sourcegraph.com/handbook/engineering/cloud#growth-plan

nicksnyder:
I merged that PR so you can rebase

handbook/engineering/developer-tools/index.md
pecigonzalo: I can merge this back to distribution, but I was actually using what we did in https://github.com/sourcegraph/about/pull/1150/ for the "Backend infrastructure team" as an example.

pecigonzalo:
This still a draft and Im still changing it.

handbook/engineering/developer-tools/index.md
nicksnyder: Ah, I didn't realize we still had that (we had some other prospective teams that I deleted and merged back into existing teams).

You can leave this if you want, but also update the distribution team to reference the growth plan.

pecigonzalo: I would like to keep it as an initial idea of that team's vision and goals, but I will add this to distribution's a distribution growth plan so its reflected there.

handbook/engineering/developer-tools/index.md
```diff
@@ -0,0 +1,15 @@
...
+
+**The developer tools team does not yet exist** and currently works as part of the [distribution team](../distribution/index.md).
+
+## Vision
+
+Create tools, systems and workflows that empower Sourcegraph engineers to work more efficiently and build quality products.
```
tsenart, line 7: This seems to have overlap with some of the goals the future [Backend Infrastructure team was supposed to have](https://about.sourcegraph.com/handbook/engineering/backend-infrastructure#vision). Is there a way to articulate this to remove this ambiguity? Or is there some actual overlap that we need to remove?

pecigonzalo: Do you mean the vision or the goals below? I think both teams will likely have similar goals, but target different parts.
As I understand the backend team will focus more on internal libraries, while this team will focus on CI/CD and dev tooling.

tsenart: I meant the vision.

> I think both teams will likely have similar goals, but target different parts.

Agreed. Can we make it clear what those differences are in the Vision statement?

> As I understand the backend team will focus more on internal libraries

Libraries and shared infrastructure.

pecigonzalo: > Libraries and shared infrastructure.

What type of shared infrastructure? Do you have something in mind?

> Can we make it clear what those differences are in the Vision statement?

Ill give it a look as I continue to update this document, is there a particular concern or a suggestion you have in mind? I dont think there is that much overlap to cause a problem at the moment.

tsenart: > What type of shared infrastructure? Do you have something in mind?

Examples: git-server and postgres.

> is there a particular concern or a suggestion you have in mind? I dont think there is that much overlap to cause a problem at the moment.

Just want to make sure we're aligned on these teams vision & long term goals and that its clear what their domains of ownership are. Would a candidate be able to understand the difference by reading the two teams pages?

pecigonzalo: I believe they would.

PR #1601, opened 2020-09-18
onboarding: ask for invite to Percy and Chromatic

tjkandala:
Question: what do we call software engineers who work on the frontend?  Common titles: frontend developer, web developer, frontend engineer, software engineer - frontend. We use "software engineer - frontend" on the hiring page, but that wouldn't sound right here ("Ask a software engineer - frontend").

Edit: Added Chromatic as well

### DIFF
--- a/handbook/engineering/onboarding.md
+++ b/handbook/engineering/onboarding.md
@@ -31,6 +31,7 @@ Your manager should complete the following steps when you join:
   - [HoneyComb.io](https://www.honeycomb.io/)
   - Ask Christina to send an invite to [Productboard](https://sourcegraph.productboard.com)
   - Ask a member of the Design team to invite as "Viewer" to [Figma](https://figma.com)
+  - Ask a frontend engineer to invite you to [Percy](https://percy.io/) (optional; recommended for frontend engineers)
 
 ## Weeks 1-3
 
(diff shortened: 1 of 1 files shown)


handbook/engineering/onboarding.md
```diff
@@ -31,6 +31,8 @@ Your manager should complete the following steps when you join:
   - [HoneyComb.io](https://www.honeycomb.io/)
   - Ask Christina to send an invite to [Productboard](https://sourcegraph.productboard.com)
   - Ask a member of the Design team to invite as "Viewer" to [Figma](https://figma.com)
+  - Ask a frontend engineer to invite you to [Percy](https://percy.io/) (optional; recommended for frontend engineers)
+  - Ask a frontend engineer to invite you to [Chromatic](https://www.chromatic.com/) (optional; recommended for frontend engineers)
```
nicksnyder, line 35: This is the EM section, so this should probably be under "Grant access to necessary services".

Is this applicable to all engineers? Should there be web specific onboarding section?

tjkandala: I originally intended to add this to the web team page, but I noticed that non-web engineers were unable to approve small UI changes on Percy or Chromatic. If we made these services visible during onboarding, they wouldn't have to wait for frontend engineers' approval for changes to copy. 

I also need to formalize + document the UI review process. Search for "chromatic" in the handbook returns 1 real result, which is the Product team onboarding page.

tjkandala: Inspired by this thread: https://sourcegraph.slack.com/archives/C07KZF47K/p1600439726004300

felixfbecker: Chromatic uses GitHub permissions, so afaik there is no way or necessity to "invite" people. Agree though that we should document the review process more. The idea here was that the @sourcegraph/design team has the authority over the UI review, while engineers have the authority over the code review. Engineers may stamp small/obvious changes too though (just like with code).

PR #1607, opened 2020-09-19
Draft Proposal to Add Team Names to RFC Metadata

Joelkw:
(This is a draft proposal. I've tagged a few people I'd love to get first eyes on this, and then may add a few more depending on your thoughts.)

The objective of the RFC is: "to be lightweight so that it can be used for many purposes (e.g. product specs, policy decisions, technical discussion), and it is optimized for facilitating collaboration and feedback." 

I don't know that adding teams helps anything when an RFC is in review, but I do think adding teams might be increasingly useful for enabling collaboration and feedback at later times, especially for new hires and as Sourcegraph continues growing (🎉). 

If you're new and trying to find RFCs that relate to your team, the best you can do is to scan titles (which requires already knowing enough to judge based on title) or search by names of people tagged (but people depart/join other teams). If we added teams to the metadata, you could search the RFCs folder for something like "Team: Web" and get only those RFCs. 

RFCs are not sources of truth (https://about.sourcegraph.com/handbook/communication#sources-of-truth), so maybe new hires shouldn't reference them anyway. I'm fine if that's the outcome, in which case I'd like to think about ways to take the "decisions" point on that source of truth page forward elsewhere – I think the decisions in abandoned RFCs in particular do contain useful information – but I'll make a new proposal and we can end the discussion here. 

If we do decide RFCs should have some sort of team meta, I don't know if this is the best way to do it. If you're a google docs magician, please show yourself. I settled on this solution because: 
- Google docs doesn't provide tagging functionality 
- Tagging addons seem to have low usage and may stop being maintained (example: https://chrome.google.com/webstore/detail/digitile-file-tag-search/cmmpmidbhloaekkpcejfmfafgchepmep)
- We already have the concept of Metadata on RFCs
- We don't use separate team folders for RFCs because they all number on the same index, so you can't have a folder for each team. I like that all RFCs are in one place and think that since the primary use case is a feedback rather than archive tool, it makes sense. 

One possible problem with this solution (suggestions here welcome!) is that enforcing exact match team names/options is difficult, but required for the google docs search to work. 

I alllllmost made this an RFC, but wanted to avoid the RFCeption.


macraig (approved):

christinaforney (approved):
I think this is a great addition, and makes it more clear who the responsible group is for the RFC (beyond the requested reviewers). It's helpful for current and new RFCs as well. 

As you mention, a core problem here is extracting the decisions that have been made within an RFC. We would really benefit from that improvement as well, and look forward to hearing more ideas and discussion about this.

PR #1618, opened 2020-09-22
Code search turned code checker

rvantonder:
---

Notes to self:

Before publishing:

- [x] page visit should trigger https://github.com/sourcegraph/sourcegraph/pull/14437
- [x] encodeURIComponent https://github.com/sourcegraph/sourcegraph/pull/14394
- [x] clicking back should restore previous history https://github.com/sourcegraph/sourcegraph/pull/14478
- [x] deny `count:1000` page visit should trigger https://github.com/sourcegraph/sourcegraph/pull/14478
- [ ] index all repos in `go-gh-100`
- [ ] Two bugs:

```
if strings.HasPrefix(:[str.], :[prefix.]) {
  :[str.] = :[str.][len(:[prefix]):]
} :[~else]
```

doesn't work

and

simple select

---

Nice to haves:

- [ ] Layout improvement (maybe 5/12 editor, 7/12 results)
- [x] comments in editor https://github.com/sourcegraph/sourcegraph/pull/14476
- [x] cmd+enter https://github.com/sourcegraph/sourcegraph/pull/14508
- [ ] `-bound-count` comby
- [ ] hover info for query
- [ ] toml config file for rewriting


blogposts/2020/code-search-turned-code-checker.md
```diff
@@ -0,0 +1,258 @@
...
+Part of an effective contribution means that we need to avoid matches in files
+that are tests, vendored, or external dependencies. While anyone can use the
+results to make contributions, it's important to be mindful of contributor
+guidelines, clearly communicate and motivate proposed changes in pull requests,
+validate that the change passes a project's tests or CI checks. And, while the
+query does exclude common test and vendored files, it's best to check that
+matches occur in files that really are part of the project.
+
+If you're interested in potentially turning these results into open source
+contributions, e-mail me at <rijnard@sourcegraph.com>, I can help.
```
rvantonder, line 241: This is an idea and still open-ended :-) Will tag and run it by Beyang and others when the post is in a more complete draft mode.

blogposts/2020/code-search-turned-code-checker.md
```diff
@@ -0,0 +1,296 @@
...
```
rvantonder, line 273: @beyang I mentioned before this idea of "help people make OSS contributions using search results", and I've narrowed down the scope of search hits that I think we would be in a good position to do that. This section of the post is open-ended, i.e., I'm fine with people reaching out to me individually, but the other way that might work better is if we had a slack channel people could join (do we still have something like that)? How do you feel about this idea generally?

sqs:
This is really cool. Here are some suggestions:

- It would be good to more clearly state why this is not just a worse staticcheck. You explain that in the text (and I get it), but it takes a lot of reading to find the exact reasons in various places.
- Related to the above, you should describe and emphasize the "lightweight workflow". You mention this once, but it should be fleshed out. Perhaps something like this (paraphrased)? "You know those times when you see a bug or a bad pattern in code, and you just fix the instance you saw? Chances are that the same problem remains unfixed in many other places. Developers generally have 2 ways to solve this problem: (1) linters (which are great for the super common, >99%-precision problems that merit writing a heavyweight lint rule); (2) better documentation, onboarding, and code review to prevent/catch these issues (which is imperfect and slow). But we're missing something in between. When you see a bug or bad pattern, you should be able to search for it across your code to see where it's lurking, and then write a fix, all interactively and without needing to be familiar with the compiler and linter APIs for the programming language. We're working to make this possible. Here's how it works, and here's how it stacks up against a popular Go lint tool (staticcheck)."
  - The "lightweight workflow" is the "you should be able to search ..." bit.
- I'm still concerned (even in that bit I wrote) that the comparison with staticcheck could make people think this is intended to replace staticcheck, no matter how much we explain that's not the goal. Could you pick some rules that would NOT make sense in staticcheck but that would make sense to find and fix using Comby patterns? The clearest/best example now is rules that apply to y/our own code, such as `actor.FromContext(ctx).UID != 0 -> actor.FromContext(ctx).IsAuthenticated()` and the custom eslint rules we have (which @felixfbecker is comfortable customizing, but most people wouldn't think to do).
- It would be nice if the linked Sourcegraph search queries (that display the search console page) have comments in them to explain what each rule is supposed to fix. (I think you've mentioned this, too.)

blogposts/2020/code-search-turned-code-checker.md
```diff
@@ -0,0 +1,296 @@
...
+
+[🔘 Example query to match known patterns in test files↗](https://sourcegraph.com/search/console?q=repo:%5Egithub%5C.com/dominikh/go-tools$%20%0Alang:go%0Afile:test%0A%0Astrings.IndexRune(...,%20...)%20%3E%20-1%0A%0Aor%0A%0Astrings.IndexRune(...,%20...)%20%3E=%200%0A%0Aor%0A%0Astrings.IndexRune(...,%20...)%20!=%20-1%0A%0Aor%0A%0Astrings.IndexRune(...,%20...)%20==%20-1%0A%0Aor%0A%0Astrings.IndexRune(...,%20...)%20%3C%200%0A%0Aor%0A%0Astrings.IndexAny(...,%20...)%20%3E%20-1%0A%0Aor%0A%0Astrings.IndexAny(...,%20...)%20%3E=%200%0A%0Aor%0A%0Astrings.IndexAny(...,%20...)%20!=%20-1%0A%0Aor%0A%0Astrings.IndexAny(...,%20...)%20==%20-1%0A%0Aor%0A%0Astrings.IndexAny(...,%20...)%20%3C%200%0A%0Aor%0A%0Astrings.Index(...,%20...)%20%3E%20-1%0A%0Aor%0A%0Astrings.Index(...,%20...)%20%3E=%200%0A%0Aor%0A%0Astrings.Index(...,%20...)%20!=%20-1%0A%0Aor%0A%0Astrings.Index(...,%20...)%20==%20-1%0A%0Aor%0A%0Astrings.Index(...,%20...)%20%3C%200%0A)
+
+Using the test files isn't a guarantee that I've implemented all the checks or
+that it's entirely precise, but it adds a lot more confidence than inferring
+patterns only from documentation.
+
+### Results
```
beyang, line 184: I think it would be interesting to show a small table comparing the definitions of these checks in `staticcheck` (presumably long/complex) vs. comby.

blogposts/2020/code-search-turned-code-checker.md
```diff
@@ -0,0 +1,296 @@
...
+downloading the 100 Go repositories to disk, and running `staticcheck` on them
+to see how the same patterns are found. For `staticcheck` to be effective, the
+project typically needs to be built first (my experience was that running
+`staticcheck` on individual files can be hit-and-miss, and understandably so). I
+didn't like the idea of doing all that work, so I punted. It does nicely motivate
+why less precise, but quicker code check workflows appeal.<sup>5</sup>
+
+### Takeaway
+
+Code search can be much more than finding your favorite provider called
```
beyang, line 224: Maybe a table showing relative advantages/disadvantages of staticcheck vs seach+comby would be evocative here.

Potential rows:
* Accuracy (syntax vs semantics)
* Scalability
* Ease of definition
* Ease of execution

blogposts/2020/code-search-turned-code-checker.md
```diff
@@ -0,0 +1,296 @@
...
+plaintext, and can't offer the fidelity of dedicated code checkers. Code
+checkers do more work to gather static information of programs, e.g., parsing
+syntax into trees and using build outputs like types and dependency graphs. It
+takes time to gather this info, it takes knowledge to write checks that use this
+info, and it takes time to hook it into your workflow. These constraints take away the
+flexibility and speed of search workflows when you're simply looking for changes
+to your favorite
+`BananaNutChocolateCake Provider`[↗](https://sourcegraph.com/search?q=BananaNutChocolateCake&patternType=literal).
+
+And yet, I can't shake the idea that there's a middle ground between crude
```
beyang, line 84: To me, this intro leaves out a huge advantage of search over staticcheck, which is scalability. With search+comby, you can execute your find-replace over the entire universe of code, whereas staticcheck requires manual setup/effort for each repository you want to cover. This difference in scale/automaticness means if I'm someone who added a new API, I would totally reach for search+comby to update existing usages of that API to take advantage of my new function. Not as easy with staticcheck, because other repository maintainers have to set it up first. Big difference in friction.

blogposts/2020/code-search-turned-code-checker.md
```diff
@@ -0,0 +1,296 @@
...
+
+Using the test files isn't a guarantee that I've implemented all the checks or
+that it's entirely precise, but it adds a lot more confidence than inferring
+patterns only from documentation.
+
+### Results
+
+Without making any claims about completeness, I was able to implement
+at least one variant for 19 out of 34 checks checks that I feel confident
```
mrnugget, line 187:
```suggestion
at least one variant for 19 out of 34 checks that I feel confident
```

mrnugget (approved):
This is really interesting, because it answers a question I've repeatedly had when looking at `staticcheck`.

I left some comments/suggestions that are all in the nitpicky/feel-free-to-ignore category.

On a general level, though, I think the structure of the blog post is a bit mixed: the first comparison between static checks & search reads as a pro/cons list, but then later there's more comparisons woven into the rest in the same pro/con tone (beginning of `## Adding more code checks`).
I think you could improve the flow by following the shape you already have and move things to their appropriate section. From what I could gather that shape is roughly this:

1. intro: what are code checks?
2. can we use code search for this? yes, with limitations, so let's explore
3. example check turned into code search
4. turning more staticcheck checks into search
4.1. approach
4.2. results
5. takeaway (here's where I'd put the pro/cons summary)
6. addendum: what do we do with our search results?

stefanhengl (commented):
I really like the post! It makes a great case why universal code search is more powerful than a purely local setup. For me personally, I would have appreciated if the post stated its purpose/intent much earlier. Right now the reader has to read quite a bit before the story becomes clear ... but that's just nit-picking ;-)

blogposts/2020/code-search-turned-code-checker.md
```diff
@@ -0,0 +1,296 @@
...
+
+and suggests instead:
+
+```go
+if strings.EqualFold(string1, string2) {
+  ...
+}
+```
+
+I find these short-and-sweet replacements are a great way to learn framework
```
lguychard, line 59: **Nit**: avoid the repetition of leading with "I find" in the first two paragraphs (it doesn't add much here).

```suggestion
These short-and-sweet replacements are a great way to learn framework
```

blogposts/2020/code-search-turned-code-checker.md
```diff
@@ -0,0 +1,296 @@
...
+some upfront configuration, the code checks are fixed in place, and productivity
+ensues (🤞).
+
+A code search can also find code snippets to replace with functions like
+`EqualFold`. In practice though, search engines generally treat code as
+plaintext, and can't offer the fidelity of dedicated code checkers. Code
+checkers do more work to gather static information of programs, e.g., parsing
+syntax into trees and using build outputs like types and dependency graphs. It
+takes time to gather this info, it takes knowledge to write checks that use this
+info, and it takes time to hook it into your workflow. These constraints take away the
```
lguychard, line 79: **nit**: refactored to not use abbreviated "info", avoid duplication, and clarify that what takes time is hooking the checks into your workflow (not the information). I find that this flows better. Hopefully for the latter that was the original intent.

```suggestion
takes time to gather this information, it takes knowledge to write checks that use it,
and it takes time to hook these checks into your workflow. These constraints take away the
```

blogposts/2020/code-search-turned-code-checker.md
```diff
@@ -0,0 +1,296 @@
...
+| Total                            | 34          |
+| Works (all variants)             | 11          |
+| Works (at least one variant)     | 19          |
+| Needs additional syntax matching | 8           |
+| Needs type info                  | 13          |
+
+<sup>Note that some checks require type info _and_ additional syntax matching.
+Also, one working variant for a check may not require type info, but another
+variant for the same check may. I.e., the values overlap and do not sum to
+Total.</sup>
```
lguychard, line 208: **nit**: since the previous and next paragraphs use "information" over "info", use it here as well for consistency. Re: may/may not, I find that starting with the affirmative flows better.

```suggestion
<sup>Note that some checks require type information _and_ additional syntax matching.
Additionally, one working variant for a check may require type information, while another
variant for the same check may not. I.e., the values overlap and do not sum to
Total.</sup>
```

blogposts/2020/code-search-turned-code-checker.md
```diff
@@ -0,0 +1,296 @@
...
+natural extension for writing better code checks, and an area of [code intelligence](https://docs.sourcegraph.com/user/code_intelligence)
+work that we're exploring.
+
+It was tempting to compare more directly with `staticcheck` output by
+downloading the 100 Go repositories to disk, and running `staticcheck` on them
+to see how the same patterns are found. For `staticcheck` to be effective, the
+project typically needs to be built first (my experience was that running
+`staticcheck` on individual files can be hit-and-miss, and understandably so). I
+didn't like the idea of doing all that work, so I punted. It does nicely motivate
+why less precise, but quicker code check workflows appeal.<sup>5</sup>
```
lguychard, line 220:
```suggestion
didn't like the idea of doing all that work, so I punted. This further highlighted
the appeal of quicker, but less precise code check workflows.<sup>5</sup>
```

blogposts/2020/code-search-turned-code-checker.md
```diff
@@ -0,0 +1,296 @@
...
+<br />
+<sup>Side note: patterns run in order and results return early after finding "enough" matches.</sup>
```
lguychard, line 238: As a reader, I find myself wanting to know what "enough" means here.

lguychard (commented):
This is very cool! 👏  I left a few non-blocking style comments.

blogposts/2020/code-search-turned-code-checker.md
beyang: I am in favor of this idea. It would be a great way to engage new users. I do not know if there is a way to enable a private Slack like ours have a public channel that anyone can join. Creating a separate community Slack channel poses the question of whether or not this would be visible enough to members of the engineering team.

What would be ideal is if we could enable anyone to join a community channel without approval, that channel is visible to members of the team, and we can restrict visibility of more sensitive channels to only Sourcegraph teammates.

rvantonder: Didn't Sourcegraph have a kind of community slack in the past? Otherwise maybe we can just make a new one, like sourcegraph-community or something and link to our internal one, like support channels. @dadlerj any thoughts on this?

blogposts/2020/code-search-turned-code-checker.md
```diff
@@ -0,0 +1,296 @@
...
+A code search can also find code snippets to replace with functions like
+`EqualFold`. In practice though, search engines generally treat code as
+plaintext, and can't offer the fidelity of dedicated code checkers. Code
+checkers do more work to gather static information of programs, e.g., parsing
+syntax into trees and using build outputs like types and dependency graphs. It
+takes time to gather this info, it takes knowledge to write checks that use this
+info, and it takes time to hook it into your workflow. These constraints take away the
+flexibility and speed of search workflows when you're simply looking for changes
+to your favorite
+`BananaNutChocolateCake Provider`[↗](https://sourcegraph.com/search?q=BananaNutChocolateCake&patternType=literal).
```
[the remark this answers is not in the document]
rvantonder: It worked before :'( I didn't create an easter egg, just found it in the wild. I can only guess the repo is gone because I tried ways to get that result again (`index:no`). Guess I'll have to settle for `ChocolateCake`.

blogposts/2020/code-search-turned-code-checker.md
```diff
@@ -0,0 +1,296 @@
...
+arguments.<sup>2</sup> The `or` keywords separate the patterns into separate
+expressions.
+
+We created a set of the top 100 Go repositories on GitHub (by stars) on
+Sourcegraph.com. We can search those by adding `repogroup:go-gh-100` to
+the query. Have a look at some of the results:
+
+[🔘 Find ways to improve code in popular Go projects ↗](https://sourcegraph.com/search/console?q=repogroup:go-gh-100%0Alanguage:go%0A-file:test%0A-file:vendor%0A-file:Godeps%0A%0Astrings.Index(...,%20...)%20%3E%20-1%0A%0Aor%0A%0Astrings.Index(...,%20...)%20%3E=%200%0A%0Aor%0A%0Astrings.Index(...,%20...)%20!=%20-1)
+<br />
+<sup>Side note: our multiline query editor is in a proof-of-concept phase.</sup>
```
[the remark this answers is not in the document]
rvantonder: It only existed as of about a ~week ago so, you're not late ;)

blogposts/2020/code-search-turned-code-checker.md
```diff
@@ -0,0 +1,296 @@
...
+it's becomes possible to write configurable code checks as self-contained search queries.
+Let's explore this idea!
+
+Here's a search query inspired by
+[a check](https://staticcheck.io/docs/checks#S1003) where `strings.Index`
+comparisons can be replaced with `strings.Contains`:
+
+```python
+language:go
+-file:test
```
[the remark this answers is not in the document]
rvantonder: why not

blogposts/2020/code-search-turned-code-checker.md
rvantonder: I don't have a clear sense how to do this with slack, we can figure it out post publishing if it makes sense, for now people can just email me and/or I will advertise it on twitter.

blogposts/2020/code-search-turned-code-checker.md
rvantonder: I like this idea in principle (thanks for suggesting it), but I feel like all of these bullets needs to be better qualified to make a valid point, and anything short of that becomes handwavy. Example:

Ease of definition: this depends a lot on who writes the definition, and who uses it. A user of staticcheck doesn't really care how easy/hard it is to define checks in that framework. Sure it's nice that they can define their own with search, but they may not be able to think of anything off the top of their mind (much nicer if someone just does the thinking for you and you push a button), so the perceived benefit to a casual user is a bit moot unless we dig into it.

If we do make have a qualified discussion, I think it detracts from a simpler takeaway ("here's an example, see this can work instead?") that applies to the audience broadly, not just people who are familiar with scalability

blogposts/2020/code-search-turned-code-checker.md
rvantonder: I agree. I want this post to be more of a deep-dive and examples of existing checks (I do get into the 'universe of code' bit more clearly in the intro now). The scenario of a custom API is great and I get into that a bit later, but it's difficult tie that benefit into this idea of code checks in the intro.

PR #1634, opened 2020-09-23
Change high agency link

macraig:
There is an article version of the Twitter thread that is easier to read


nicksnyder (approved):
Thanks!

PR #1643, opened 2020-09-24
Updated BizOps project tracking

ebrodymoore:
Update how we work


AlicjaSuska:
Great changes @ebrodymoore During the last product meeting you've mentioned that we should tag you as soon as the RFC starts. Maybe that would also be worth adding to the guidelines both here and in PM and design docs? wdyt?

ebrodymoore:
@AlicjaSuska good point, thanks for mentioning!

dadlerj (approved):
Thanks for getting this in. Sorry to do this, but could you rebase on the new version of this file (after my refactor landed last night)?

PR #1644, opened 2020-09-24
Update engineering-manager-code-intelligence.md

withdavidli:
Deleting hyperlinks to internal docs to questions. Since this is a candidate facing page, it will be confusing why it leads them to a 404 page.


nicksnyder:
I like that the JD is a source of truth for all relevant things, but I agree this is confusing since I have candidates ask about this. Can you think of a way to keep the links but make it clear that they are internal to Sourcegraph?

PR #1653, opened 2020-09-25
Add operations RFCs

dadlerj:
Interested to hear feedback! The more I looked at our RFC docs and Christina's PD docs, the more I came to realize that not creating a whole new category would be beneficial. Instead, I created this summary that focuses more on how they are the same than trying to build a new template.


ebrodymoore (approved):
LGTM - I like keeping the structure the same. Looking back over the last few months I don't see many instances where I personally would've created one but I think this is a great addition and will get more and more useful over time as the ops team grows.

PR #1660, opened 2020-09-28
move company meeting 2h earlier

sqs:
We've added a lot more people outside of the California time zone in the last year. Moving the meeting 2h earlier is better for many people now.

### DIFF
--- a/handbook/communication/company_meeting.md
+++ b/handbook/communication/company_meeting.md
@@ -1,6 +1,6 @@
 # Company meeting
 
-1. Schedule: every Monday 10:30-11:00am PST/PDT (link in the calendar event)
+1. Schedule: every Monday 8:00-8:30am PST/PDT (link in the calendar event)
    - If Monday is a holiday, the meeting is held on the next non-holiday day.
    - There is no company meeting held between December 25 and December 31. The first company meeting of the year is held on the first non-holiday Monday of the year.
 1. The link to the slides is posted in #general each Thursday before.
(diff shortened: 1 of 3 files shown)


handbook/communication/company_meeting.md
```diff
@@ -1,6 +1,6 @@
 # Company meeting
 
-1. Schedule: every Monday 10:30-11:00am PST/PDT (link in the calendar event)
+1. Schedule: every Monday 8:00-8:30am PST/PDT (link in the calendar event)
```
keegancsmith, line 3: maybe we should use UTC to avoid shifts due to daylight savings?

We still have a lot of people in PST/PDT, is 8am maybe inconvenient due to its overlap with school start times? I guess its all a tradeoff, I miss plenty of company meetings currently due to overlap with trying to get my son to sleep.

handbook/communication/company_meeting.md
```diff
@@ -1,6 +1,6 @@
...
    - If Monday is a holiday, the meeting is held on the next non-holiday day.
```
keegancsmith, line 4: I believe we don't respect that anymore, since holidays aren't global?

keegancsmith (approved):

handbook/communication/company_meeting.md
nicksnyder: There are a small number where it is documented that we move company meeting.

```suggestion
   - If Monday is a [holiday](../people-ops/holidays.md), the meeting is held on the next non-holiday day.
```

daxmc99 (approved):

handbook/communication/company_meeting.md
sqs: It's all a tradeoff. Based on people's feedback, I think this is the best for now. In general, we want it to be as early as possible for Californians (of which we have many), so that it's as non-late as possible for everyone else. I think 7am local time in California is too early, so pinning it to 8am California local time optimizes for early but not too early.

PR #1694, opened 2020-10-01
add link to how to create a blog post

chrispine:
Previously, searching the handbook for how to create a blog post didn't turn up much, because those instructions are in this repo's README. This adds links to those instructions from within the handbook itself, so people searching for "create blog post" or even just "blog post" will find them.


christinaforney:
Thanks for doing this! I propose linking in the inverse - moving the content about creating a blog post from the README into the handbook and then linking to the handbook from the README.

chrispine:
Great idea, @christinaforney!

PR #1695, opened 2020-10-01
Update onboarding to include Percy

ElizabethStirling:
Not sure if we really use Percy too often, but if we do, it should be part of onboarding. Looking for feedback on this.


daxmc99:
Frontend uses this so perhaps mention @sourcegraph/frontend-engineers ?

quinnkeast (approved):
Same as you, I'm not sure how much it's used, but if it's a tool in the stack we should make it part of onboarding so it's there.

felixfbecker (approved):

PR #1699, opened 2020-10-02
Add the rest of the languages to list

macraig:
Some of the supported languages were missing, adding them to give more clarity to users browsing the page.


christinaforney (approved):
@sqs - thoughts on alphabetizing the list of languages? It would make it easier to quickly find a specific language I think. Was there a reason for the original order here? I imagine popularity?

sqs:
The order was rough popularity. It would be weird to have ALGOL and ActionScript and Ada and AppleScript first if it becomes comprehensive.

We could just cap it at a certain number so no obscure languages are included, and then link to “and N other languages”. And then alphabetize the ones we do show.

PR #1702, opened 2020-10-02
explain wellness benefit carry-forward process

limitedmage:
As explained earlier on [Slack](https://sourcegraph.slack.com/archives/CQAGQKC4A/p1601513041017300?thread_ts=1600388710.002700&cid=CQAGQKC4A)


limitedmage:
Wellness expensing is automatic now so no need for this.

PR #1718, opened 2020-10-06
Add to gcp/projects information

pecigonzalo:
Guide engineers to use https://github.com/sourcegraph/infrastructure/pull/2155 for creating new projects.


pecigonzalo:
@daxmc99 Maybe there is something we can clarify here related to your comment in https://github.com/sourcegraph/infrastructure/pull/2155

PR #1719, opened 2020-10-06
Add additional information about internal services

pecigonzalo:
Related issue: https://github.com/sourcegraph/sourcegraph/issues/13917

Some of the [tooling](https://github.com/sourcegraph/infrastructure/tree/main/dogfood/kubernetes/tooling) where not described in our documentation, this adds a reference about them and a link to their definition.

### DIFF
--- a/handbook/engineering/distribution/internal_infrastructure.md
+++ b/handbook/engineering/distribution/internal_infrastructure.md
@@ -2,15 +2,33 @@
 
 ## Code host testing instances
 
-We maintain a collection of code hosts for testing purposes.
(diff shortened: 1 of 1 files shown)


bobheadxi (approved):
Nice! I feel like we can probably include a link to https://about.sourcegraph.com/handbook/engineering/deployments as well here since a lot of that page is also internal services (internal sourcegraph instances)

pecigonzalo:
Will do!

pecigonzalo:
@bobheadxi added, is that what you had in mind?

bobheadxi:
Yep, LGTM - thanks!

PR #1723, opened 2020-10-06
G Suite was renamed to Google Workspace

sqs:
https://workspace.google.com/


website/static/_redirects
```diff
@@ -55,7 +55,7 @@
 /docs/extensions https://docs.sourcegraph.com/extensions 301!
 /docs/features/api https://docs.sourcegraph.com/api/graphql 301!
 /docs/features/browser-extension https://docs.sourcegraph.com/integration/browser_extension 301!
-/docs/features/browser-extension/g-suite-installation https://docs.sourcegraph.com/integration/google_gsuite 301!
+/docs/features/browser-extension/g-suite-installation https://docs.sourcegraph.com/integration/google_workspace 301!
```
limitedmage, line 58: Is this intended? Do we want a new redirect here for the old route?

Issue #1714, opened 2020-10-06
Add F5 Networks case study to website


lorihey:
Here is the content:

https://docs.google.com/document/d/101VIT2m1O8Bdg7R4PU3ll-wjM_6SiwwseDxQqfAoaMw/edit?usp=sharing

We'll need a new case study page created in the usual format, and we need to add a link to it from this page:
https://about.sourcegraph.com/case-studies

We are waiting for Satish' photo, it should be coming soon.

Note, their logo guidelines are here: https://www.f5.com/pdf/f5/f5-creative-guidelines.pdf

PR #1734, opened 2020-10-07
Add F5 case study

aileenrose:
https://github.com/sourcegraph/about/issues/1714

DRAFT Pull Request - Awaiting headshot for Satish Surapaneni from F5, if available.

For review:
https://deploy-preview-1734--sourcegraph.netlify.app/case-studies/
https://deploy-preview-1734--sourcegraph.netlify.app/case-studies/f5-streamlines-collaboration-globally/


lorihey:
A few things:
1. Since Satish is the only one speaking, we don't need to use his name after every quote -- we usually do that when there are several people being quoted. So the quotes can stand alone, without attribution -- like we did in this case study: https://about.sourcegraph.com/case-studies/we-are-thorn
2. Please add an extra line of space at the end of the case study, after the last paragraph. It's too close to the black footer. You can match the same spacing as on this case study: https://about.sourcegraph.com/case-studies/we-are-thorn
Thanks!

aileenrose:
> A few things:
> 
> 1. Since Satish is the only one speaking, we don't need to use his name after every quote -- we usually do that when there are several people being quoted. So the quotes can stand alone, without attribution -- like we did in this case study: https://about.sourcegraph.com/case-studies/we-are-thorn
> 2. Please add an extra line of space at the end of the case study, after the last paragraph. It's too close to the black footer. You can match the same spacing as on this case study: https://about.sourcegraph.com/case-studies/we-are-thorn
>    Thanks!

@lorihey Updated.

lorihey (approved):

lorihey:
Looks good.

PR #1738, opened 2020-10-09
Updating company values

christinaforney:
@nicksnyder and I felt that there was a gap in our values that would help with hiring decisions, growth conversations, process decisions, and more. We also felt that a short, clear list of values would help more clearly communicate the values in a way that can be digested and remembered. The details are valuable for clarifying the shorter list and adding more color to what this means to us but isn't necessary for communicating the high level understanding of values.

The newly added values have empty sub-sections and TO DOs. I propose we merge as is, and can follow up with additional explanations.

Specifically, we have tweaked two of the existing values:
- 'We not I' -> 'Work as a team'
- ~'High quality' -> 'Be iterative' - this was sending the wrong message (as indicated by the need to clarify in company meeting). Instead, what we value is being iterative to creating value of high quality work.~ This still needs more work, reverting back to high quality here to move the PR forward and open up a follow up PR to clarify.

PR #1760, opened 2020-10-14
Improvements to values

christinaforney:
Follow up PR to https://github.com/sourcegraph/about/pull/1738

- [x] Make order of details sections match top-level order of values
- [x] Add suggestions from @AlicjaSuska 
- [ ] Figure out how to reconcile High-quality and iterative
- [ ] Improve wording of high agency one sentence summary to communicate sentiment from this twitter thread: https://twitter.com/shreyas/status/1276956838840295425 
- [ ] Move 'be human' to the top of the list? @lguychard 
- [ ] Make sure usage of we/us/you is clear and consistent throughout @chrispine

### DIFF
--- a/company/values.md
+++ b/company/values.md
@@ -49,7 +49,7 @@ You are responsible for finding out what high-quality work looks like and delive
 
 If we discover that we've produced something that does not meet our quality bar, we'll roll it back (for example, revert a commit, roll back a deployment, or unpublish content) until we can make it high quality. If that's not possible, we'll prioritize improving it.
 
-High-quality isn't the opposite of iterative. You can keep quality high while still being iterative by narrowing the scope (for example, by solving a smaller problem at first, or by solving it only for certain customers initially).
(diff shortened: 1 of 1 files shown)


nicksnyder:
I think we should avoid coupling reordering values with changing them as it makes the diff harder to review.

I think either reordering should happen last, or silently do that as a direct commit to the main branch.

nicksnyder:
I also think if we can get the base branch merged, then each of these could be their own PR that we can merge without coupling unrelated suggestions.

company/values.md
```diff
@@ -89,10 +89,12 @@ We document our goals and plans publicly so we can engage with customers to buil
 
 Our code and PRs are public, so lack of transparency is only obfuscation, not security. Unless something is immediately exploitable and would put customer data at risk, you should be completely transparent about it, just like any other non-security improvement.
 
+Being open and transparent also means sharing your work frequently. You should keep others informed on what you're working on, your progress, and plans. It is helpful in getting more diverse feedback and spotting dependencies between the teams and the solutions. It's your responsibility to make everyone feel welcomed to share their point of view.
```
chayim, line 92: I really like this! Perhaps remove the idea of _should_ so that this seems less suggestive - i.e it indicates our norm is that "You keep others informed"

AlicjaSuska: +1 🎉

PR #1768, opened 2020-10-15
make engineering page index clearer

slimsag:
This has become a bit of a globby mess, so I am trying to clean it up here before introducing a new page link.


nicksnyder:
Thank you for helping to improve this; I agree it is a mess :)

nicksnyder (approved):

PR #1770, opened 2020-10-16
Split deployments doc into multiple sections

pecigonzalo:


handbook/engineering/deployments/index.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Infrastructure
+
+The cloud resources (including clusters, DNS configuration, etc.) on which are deployments run should be configured in the [infrastructure repository](https://github.com/sourcegraph/infrastructure), even though Kubernetes deployments are managed by various `deploy-sourcegraph-*` repositories. For information about how our infrastructure is organized, refer to [Environments](./environments.md).
+
+## deploy-sourcegraph
+
+![Renovate downstream](https://github.com/sourcegraph/sourcegraph/workflows/Renovate%20downstream/badge.svg) [![master build status](https://badge.buildkite.com/018ed23ed79d7297e7dd109b745597c58d875323fb06e81786.svg?branch=master)](https://buildkite.com/sourcegraph/deploy-sourcegraph) ![Dispatch update](https://github.com/sourcegraph/deploy-sourcegraph/workflows/Dispatch%20update/badge.svg)
```
ElizabethStirling, line 46: We should probably also update these repos to use main, rather than master, just to keep everything consistent. Not something that should block this PR, but something I wanted to call out.

ElizabethStirling (approved):
This is wonderful, thank you so much for taking the time to rework this!

uwedeportivo (approved):

handbook/engineering/deployments/playbooks.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Via BigQuery (for read-only operations)
+
+You can also query the production database via BigQuery as an external data source.
+
+See an [example query](https://console.cloud.google.com/bigquery?sq=527047051561:bfa7c7e57f884d209f261d15e4610229) to get started.
+
+**Note**: This method only permits read-only access
+
+# Manage users in k8s.sgdev.org
```
bobheadxi, line 64: Playbooks like this (and the dot-com one) imo should just go in the main deployments page, under each deployment where they used to be, to help with discoverability and avoid confusion - I think anything on separate pages like this one should not be deployment-specific

(a bit late on the review, but happy to put in a PR if you don't disagree)

pecigonzalo: I disagree, the document was quite hard to GROK as it had a mix of descriptions of the deployments, steps to manage things, etc. Maybe we could group them per-deployment in the playbook section if they are for a particular deployment, which would potentially fix both our concerns, wdyt?

I would agree to generalize those steps, as most of them should be fairly generic for the most part, but I chose in this PR to keep the same content as I was already moving a lot of things around which make the PR harder to review.

bobheadxi: In that case per-deployment playbook sections + links to those sections from the main deployments page would be a good middle ground 👍

pecigonzalo: Maybe some further context, when I moved them I also left the names there as it would make it easier to use search for `k8s.sgdev.org` and the playbook doc would pop up, helping discoverability.

PR #1784, opened 2020-10-20
blogpost: Introducing a new database for code intel

efritz:
Closes https://github.com/sourcegraph/sourcegraph/issues/14009.

This is a quick blog post to coincide with the release of 3.21 and the introduction of a second database instance, which may throw some site admins off at first. This blog post is meant to explain the reasons we chose to add a second database and how it might affect existing users.

I have also written the story detailing the weird code intelligence outage that happened in the cloud instance when applying this change, but am planning on publish it on my personal blog instead. I think the story and circumstances are interesting (and should be made available), but is not necessarily something we want to immortalize directly next to the announcement for the change that caused it. See https://github.com/efritz/blog/pull/13 for the outage story and feel free to leave comments on that PR as well.

My thinking may be skewed here, so I'm happy to hear alternative viewpoints!


macraig (approved):

efritz:
Questions about things that generally need updating after my first draft of blog posts:

- Is this enough context for the audience, or do I need to explain code intel in-depth at the intro?
- This is more of an announcement for existing site-admins in order to justify any burden on their part of adjusting their deployments more than anything. Do we need more of an outtro?

blogposts/2020/introducing-codeintel-db.md
```diff
@@ -0,0 +1,65 @@
...
+  "blog"
+]
+slug: introducing-a-new-code-intelligence-database
+heroImage: https://image.flaticon.com/icons/png/512/51/51333.png
+published: true
+---
+
+With the release of 3.21, Sourcegraph will ship with a _second_ Postgres database instance, named `codeintel-db`, which we use to store the data that powers [precise code intelligence](https://docs.sourcegraph.com/code_intelligence/explanations/precise_code_intelligence).
+
+We've [written previously](https://about.sourcegraph.com/blog/evolution-of-the-precise-code-intel-backend/) about the architectural changes we've made to the code intelligence backend. This post is yet another submission in the archaeological trail of how the code intelligence team's effort shaped Sourcegraph on a technical level, and will explain why we chose to swap out our persistence layer - a very non-trivial difference in our architecture.
```
uwedeportivo, line 15:
```suggestion
We've [written previously](https://about.sourcegraph.com/blog/evolution-of-the-precise-code-intel-backend/) about the architectural changes we have made to the code intelligence backend. This post is yet another submission in the archaeological trail of how the code intelligence team shapes Sourcegraph on a technical level. The post explains why we chose to swap out our persistence layer - a very non-trivial difference in our architecture.
```

blogposts/2020/introducing-codeintel-db.md
```diff
@@ -0,0 +1,65 @@
...
+
+Using SQLite for this data has become an operational and feature burden for the code intelligence team.
+
+1. We need to maintain a periodic process to reconcile the data on disk with the metadata we store in Postgres.
+2. We have good instructions and tips for backup and recovery of our Postgres data store, but have never officially recommended backing up other persistent volumes.
+3. We've locked ourselves into a node topology which is not embarrassingly scalable. We essentially have a singleton service over a persistent volume. In order to scale this service horizontally (which would become necessary once either the amount of data or the frequency of queries increased), we would need to shard the data across multiple volumes. This throws a lot of wrenches into our plans: we need to be able to automatically move data on increase or decrease of shard counts; we need to be able to route requests to the correct node with that data; we need to ensure that reconciliation between the data on disk and the metadata we store in Postgres is aware that there are multiple nodes where data _could_ reside.
```
uwedeportivo, line 33:
```suggestion
3. We locked ourselves into a node topology which is not easily scalable. We essentially have a singleton service over a persistent volume. In order to scale this service horizontally (which would become necessary once either the amount of data or the frequency of queries increased), we would need to shard the data across multiple volumes. This throws a lot of wrenches into our plans: we need to be able to automatically move data on increase or decrease of shard counts; we need to be able to route requests to the correct node with that data; we need to ensure that reconciliation between the data on disk and the metadata we store in Postgres is aware that there are multiple nodes where data _could_ reside.
```

blogposts/2020/introducing-codeintel-db.md
```diff
@@ -0,0 +1,65 @@
...
+2. Moving data to an existing technology allows us to re-use the same advice and documentation for backup and disaster recovery strategies. Removing the number of knobs that must be turned by site-admins is always going to be a win across the board.
+3. Moving from an embedded database to a server-client database removes the nasty singleton property on the service. Now that multiple servers can interact with the same data concurrently without risk of corruption, we can scale the services horizontally without requiring additional effort into implementing sticky sessions or complex request routing rules.
+4. Moving data into a single relational store allows us to create additional indexed views into the data. We can now query over dimensions that were not possible before without opening tends of thousands of relatively small database files.
+
+## Notes for site admins
+
+If you are an active user of LSIF, there are a few changes that may subtly affect the behavior of the code intelligence backend. Our enterprise users operate at incredibly different scales and is difficult to ensure that a change has zero impact on all existing customers.
```
uwedeportivo, line 47:
```suggestion
If you are an active user of LSIF, there are a few changes that may subtly affect the behavior of the code intelligence backend. Our enterprise users operate at incredibly different scales so it is difficult to ensure that a change has zero impact on all existing customers.
```

blogposts/2020/introducing-codeintel-db.md
```diff
@@ -0,0 +1,65 @@
...
+
+#### Data retention policy
+
+The [3.21 changelog](https://github.com/sourcegraph/sourcegraph/blob/3.21/CHANGELOG.md) details the introduction of a new environment variable, `PRECISE_CODE_INTEL_MAX_DATA_AGE`, on the precise-code-intel-bundle-manager service.
+
+Previously, the bundle manager stored all LSIF data on disk as SQLite files. This allowed us to run custodial processes which removed old data, but only when disk pressure was high. Since we've migrated this data to a Postgres instance, there is a less clear-cut method to determine how much data we need to free in order to keep the desired amount of headroom.
+
+Disks do not grow on trees and we at Sourcegraph are well aware of that fact. In order to ensure we didn't accidentally slip an infinite disk space requirement to the Sourcegraph 3.21 release, we introduced an alternate data retention policy to replace the old full-disk trigger. Any data that is older than `PRECISE_CODE_INTEL_MAX_DATA_AGE` and is not used to serve code intelligence queries for the tip of a default branch will be automatically flushed from the code intelligence database periodically.
```
uwedeportivo, line 63:
```suggestion
Disks do not grow on trees and we at Sourcegraph are well aware of that fact. In order to ensure we didn't accidentally require an infinite disk space in the Sourcegraph 3.21 release, we introduced an alternate data retention policy to replace the old full-disk trigger. Any data that is older than `PRECISE_CODE_INTEL_MAX_DATA_AGE` and is not used to serve code intelligence queries for the tip of a default branch will be automatically flushed from the code intelligence database periodically.
```

blogposts/2020/introducing-codeintel-db.md
```diff
@@ -0,0 +1,65 @@
...
+
+We believe that this will give us roughly the same behavior as the previous versions, and can be tuned to match expected behavior at your instance's scale. The default value for this environment variable is `720h` (30 days), which is likely adequate for the majority of instances. This value should be adjusted if the size of your index uploads are large and your commit frequency is high.
```
uwedeportivo, line 65:
```suggestion
We believe that this will give us roughly the same behavior as the previous versions, and can be tuned to match expected behavior at the scale of your instance. The default value for this environment variable is `720h` (30 days), which is likely adequate for the majority of instances. This value should be adjusted if the size of your index uploads are large and your commit frequency is high.
```

uwedeportivo (approved):

blogposts/2020/introducing-codeintel-db.md
```diff
@@ -0,0 +1,65 @@
...
+2. Moving data to an existing technology allows us to re-use the same advice and documentation for backup and disaster recovery strategies. Removing the number of knobs that must be turned by site-admins is always going to be a win across the board.
+3. Moving from an embedded database to a server-client database removes the nasty singleton property on the service. Now that multiple servers can interact with the same data concurrently without risk of corruption, we can scale the services horizontally without requiring additional effort into implementing sticky sessions or complex request routing rules.
+4. Moving data into a single relational store allows us to create additional indexed views into the data. We can now query over dimensions that were not possible before without opening tends of thousands of relatively small database files.
```
nicksnyder, line 43: I think this section conflates reasons why we are making this change now, and potential benefits we will experience in the future. I think it is important to keep these things separate.

The conditional future tense used throughout this passage gives it won't be until some future point that we actually benefit from this change. As I reader I am left wondering "why make this change now?".

blogposts/2020/introducing-codeintel-db.md
```diff
@@ -0,0 +1,65 @@
...
+4. There is no efficient way to implement certain data access patterns we may need in the future. For example, looking for data matching a certain identifier _name_ would require us to open every SQLite database on disk (and, no, _batching it_ is not a solution).
+
+Instead of forging ahead and tackling the complexity we've made for ourselves, we decided to take a step back and re-evaluate whether or not our previous architecture choices were meeting our current needs. And it seems like they did not.
+
+Moving this data into Postgres seems like an obvious choice that will (eventually) be able to reduce or completely remove all of the burdens listed above.
```
nicksnyder, line 38: "Seems like an obvious choice" and "(eventually)" sound wishy-washy and aren't confidence inspiring statements. Are you not confident this was the right decision?

blogposts/2020/introducing-codeintel-db.md
```diff
@@ -0,0 +1,65 @@
...
+2. Moving data to an existing technology allows us to re-use the same advice and documentation for backup and disaster recovery strategies. Removing the number of knobs that must be turned by site-admins is always going to be a win across the board.
+3. Moving from an embedded database to a server-client database removes the nasty singleton property on the service. Now that multiple servers can interact with the same data concurrently without risk of corruption, we can scale the services horizontally without requiring additional effort into implementing sticky sessions or complex request routing rules.
+4. Moving data into a single relational store allows us to create additional indexed views into the data. We can now query over dimensions that were not possible before without opening tends of thousands of relatively small database files.
+
+## Notes for site admins
```
nicksnyder, line 45: Do we want this blog post to be the source of truth for information like this?

cc @christinaforney

efritz: This information is already available in the upgrade docs and the changelog, I just wanted to call it out as an extra precaution.

efritz: We already did have a question in a support channel about these values, so it's apparently pretty easy to glaze over some of the release notes.

blogposts/2020/introducing-codeintel-db.md
efritz: Will strengthen!

blogposts/2020/introducing-codeintel-db.md
efritz: I'll separate them (at least) and (maybe) remove the weaker one.

nicksnyder: I think it is ok to talk about hypothetical future benefits as perks, as long as that is separate from and after the reasons why we are doing it now.

blogposts/2020/introducing-codeintel-db.md
nicksnyder: I think the right way to call it out is to link to a source of truth. If those documents don't have sufficient information (customers have asked questions) then I think we should add context to the source of truth, not add more information here in this blog post.

efritz: There are already links to https://docs.sourcegraph.com/admin/external_database and https://docs.sourcegraph.com/admin/updates in this text.

As for the question we got already:

```
does this use the pgsql instance we configured?  it looks like it is trying to stand up another db for this.... if that is the case, i need to find the settings to configure it to use the HA pgsql instance we have placed
```

I think the addition just took them by surprise, because their intuition about how to update their deployment was correct.

Here is the text in the upgrading section:

```
This release introduces a second database instance, codeintel-db. If you have configured Sourcegraph with an external database, then update the CODEINTEL_PG* environment variables to point to a new external database as described in the external database documentation. Again, these must not point to the same database or the Sourcegraph instance will refuse to start.
```

nicksnyder: So my question is, why not delete all the content and just keep the links? If there is any content that you hesitate to delete, ask yourself if it belongs somewhere else (e.g. inside one of those links).

efritz: The information under the data retention policy I think is useful, but would not fit nicely inside the changelog (where the parameter is introduced).

Maybe this isn't the best place for it, but I don't think the changelog is the best places for it either.

nicksnyder:
I still have some feedback and I think it would be most effective to chat live tomorrow.

blogposts/2020/introducing-codeintel-db.md
christinaforney: Let me know if you need more input on this post (I skimmed through it, and my concerns are the same that Nick mentions above - making sure the context is in the appropriate location).

efritz: I'm assuming Nick and I will discuss this online in our 1-1 tomorrow, but I'll tap you if we discuss it earlier in case you're interested.

Strum355 (approved):

gbrik:
It might be cool to put the benchmarking data? I feel like I know devs who would be surprised that SQLite could keep up with Postgres and devs who would be surprised that Postgres could keep up with SQLite.

blogposts/2020/introducing-codeintel-db.md
```diff
@@ -0,0 +1,68 @@
...
+- We have good instructions and tips for backup and recovery of our Postgres data store, but have never officially recommended backing up other persistent volumes.
+- We locked ourselves into a node topology which is not easily scalable. We essentially have a singleton service over a persistent volume. In order to scale this service horizontally (which would become necessary once either the amount of data or the frequency of queries increased), we would need to shard the data across multiple volumes. This throws a lot of wrenches into our plans: we need to be able to automatically move data on increase or decrease of shard counts; we need to be able to route requests to the correct node with that data; we need to ensure that reconciliation between the data on disk and the metadata we store in Postgres is aware that there are multiple nodes where data _could_ reside.
+- There is no efficient way to implement certain data access patterns we may need in the future. For example, looking for data matching a certain identifier _name_ would require us to open every SQLite database on disk (and, no, _batching it_ is not a solution).
```
gbrik, line 34: wdym by batching in this context?

blogposts/2020/introducing-codeintel-db.md
```diff
@@ -0,0 +1,68 @@
...
+
+Instead of forging ahead and tackling the complexity we've made for ourselves, we decided to take a step back and re-evaluate whether or not our previous architecture choices were meeting our current needs. And it seems like they did not.
+
+Moving this data into Postgres was an obvious choice that greatly reduced the burdens listed above.
+
+- Moving data to an existing technology allows us to re-use the same advice and documentation for backup and disaster recovery strategies. Removing the number of knobs that must be turned by site-admins is always going to be a win across the board.
```
gbrik, line 40: Change `existing` to something like "already used elsewhere in our tech stack"?

gbrik: also, either 'removing knobs' or 'lowering the number of knobs'

gbrik: also, did this change reduce the number of knobs, or just the number of types of knobs?

blogposts/2020/introducing-codeintel-db.md
efritz: Clarified.

blogposts/2020/introducing-codeintel-db.md
efritz: #1: Updated.
#2: Good catch.
#3: Net number of knobs is lower. We'll no longer have to worry about resource allocation in the bundle manager query path, and we'll be able to remove most of the janitor settings (once we get S3/GCS/MinIO to replace the rest of the bundle manager, there will be like 1-2 settings at all for this).

PR #1221, opened 2020-07-16
distribution: add monitoring architecture page

slimsag:
@bobheadxi noted to me over Slack:

image

I think this is a great idea and something that I have long intended to put together - but I think an RFC may be the wrong forum for it because many of these decisions have already been made. I am therefor proposing the following:

1. The current monitoring architecture and decisions we have already made are documented and live in this document. For any question @pecigonzalo has asked me or you @bobheadxi about "why are we doing this this way?" I would like this document to explain it
2. For changes to the current monitoring architecture and decisions we are considering, those should be RFCs (or, if small changes they can be lightweight `Proposal: ...` GitHub issues, like https://github.com/sourcegraph/sourcegraph/issues/12010).

@bobheadxi can you own filling this out from here with all the context I've transferred to you, any questions Gonza has asked you, etc.? And then I'll follow-up and add any context I may have as well after

Rendered: https://github.com/sourcegraph/about/blob/monitoring-architecture/handbook/engineering/distribution/observability/monitoring_architecture.md

PR #1783, opened 2020-10-20
observability: remove outdated prometheus page, move to monitoring

bobheadxi:
I did a search for Prometheus stuff and came across this very outdated page - I've moved relevant bits into the new documentation, and removed the page and references to it completely 

Also cleaned up some headers to make them easier to reference

PR #1785, opened 2020-10-20
observability docs: move file to where it's referenced

rvantonder:
https://github.com/sourcegraph/about/pull/1221 broke docsite build. It looks like this file should live where I moved it in this PR see https://github.com/sourcegraph/about/runs/1279046251#step:4:23


bobheadxi:
Oh, I started to try and fix the references in #1783 😅

rvantonder:
@bobheadxi OK, I looked at the docsite logs and it's this file--based on the change, the referenced file should live where I moved it, and this PR unbreaks the build (which is really the part I care about :-) it's either this PR or revert the other). Can I get a stamp here please, and you can fix up anything else that needs Real Fixing (TM) in your PR, if needed?

bobheadxi (approved):

bobheadxi:
(we might want to make the docsite build a required check too)

rvantonder:
agree!

PR #1787, opened 2020-10-20
distribution: Deployment pipeline UX goal

pecigonzalo:
There is a related [RFC-239](https://docs.google.com/document/d/1-qPAITDZ1uvOqo4lUBK0qmUr4YhXLOp7KcS6hAdwSA4/edit#heading=h.trqab8y0kufp) which is different in scope, as that conversation is about deploying continuously or not to sourcegraph.com, while this goal is around making the deployment to X environment more transparent.

Related Slack thread: https://sourcegraph.slack.com/archives/CJX299FGE/p1601450437095100

### DIFF
--- a/handbook/engineering/distribution/goals.md
+++ b/handbook/engineering/distribution/goals.md
@@ -15,7 +15,7 @@ Goals are continuously updated and reviewed. If you find these goals do not refl
   - We are able to rollback to or redeploy a previous version.
   - Stretch: We can correlate deployments to Grafana metrics and Stackdriver logs.
   - TBD: Branches can trigger a parallel `sourcegraph-frontend` deployment.
-  - TBD: We can trigger a deployment via a `/` command in Slack.
+  - TBD: We can trigger rollbacks and deployments via a `/` command in Slack.
 - **Milestones**: TBD
 
 ### Any engineer at Sourcegraph can create a release for all of our supported deployment types by running a single command
--- a/handbook/engineering/distribution/goals.md
+++ b/handbook/engineering/distribution/goals.md
@@ -14,7 +14,7 @@ Goals are continuously updated and reviewed. If you find these goals do not refl
   - Deployments display information about the artifact they are deploying.
   - We are able to rollback to or redeploy a previous version.
   - Stretch: We can correlate deployments to Grafana metrics and Stackdriver logs.
-  - Stretch: Branches can trigger a parallel `sourcegraph-frontend` deployment.
+  - TBD: Branches can trigger a parallel `sourcegraph-frontend` deployment.
   - TBD: We can trigger a deployment via a `/` command in Slack.
 - **Milestones**: TBD
 
--- a/handbook/engineering/distribution/goals.md
+++ b/handbook/engineering/distribution/goals.md
@@ -8,7 +8,7 @@ Goals are continuously updated and reviewed. If you find these goals do not refl
 
 Creating a new release for our deployments is currently a semi-automated process, which requires several manual steps and synchronizing our versioned artifacts (Sourcegraph, Kubernetes manifests, docker-compose manifests, etc). We want to enable any engineer to perform a release as often as needed, to enable this we want to make releasing Sourcegraph a simple, automated process.
 
-- **Owner**: Stephen, Dave, Robert
+- **Owner**: Dave, Robert
 - **Status**: In progress. Estimated completion by end of 2020.
 - **Outcomes**:
   - Releases can be triggered by a single manual step.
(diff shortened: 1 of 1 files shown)


handbook/engineering/distribution/goals.md
```diff
@@ -4,11 +4,21 @@ Goals are continuously updated and reviewed. If you find these goals do not refl
 
 ## Medium-term goals (~3-6 Months)
 
+### [Improve deployment pipeline UX](https://github.com/orgs/sourcegraph/projects/96)
+
+- **Owner**: TBD
+- **Status**: Not started. Unknown amount of work.
+- **Outcomes**:
+  - Its simple to identify if a commit has been deployed to an environment
+  - We can trigger a deployment via a `/` command in Slack
```
bobheadxi, line 13: I was under the impression that the `/`-command thing was primarily for fast rollbacks, which I think is a nice idea - I think `/`-command just for deployment seems like a crutch for what should be a more transparent deployment process (what we should aim for, ie the point above, which should probably encapsulate "how my commit gets deployed is obvious")

handbook/engineering/distribution/goals.md
```diff
@@ -4,11 +4,21 @@ Goals are continuously updated and reviewed. If you find these goals do not refl
 
 ## Medium-term goals (~3-6 Months)
 
+### [Improve deployment pipeline UX](https://github.com/orgs/sourcegraph/projects/96)
+
+- **Owner**: TBD
+- **Status**: Not started. Unknown amount of work.
+- **Outcomes**:
+  - Its simple to identify if a commit has been deployed to an environment
```
bobheadxi, line 12:
```suggestion
  - It is simple to identify when a commit has been deployed to an environment
```

I feel like _if_ is a bit too easy to solve (for example, you can already kind of tell from the build version in `https://sourcegraph.com/users/$ME/settings` ), but much more interesting would be knowing _when_ a commit has been deployed (for example, via a comment on the PR, or through Buildkite, or Slack, or etc.) which seems like it could be nice for triage as well (a question that seems to come up often is, "did something deploy at x time?")

bobheadxi (commented):
some quick thoughts

tsenart:
🥳

handbook/engineering/distribution/goals.md
tsenart: Yeah, agree, plus, I'd like to see the history of deployments easily.

pecigonzalo: >I feel like if is a bit too easy to solve 

Thanks, yeah that is what I meant, not only when as in "timeline" but rather, I merged, is my commit already in the environment?

>like to see the history of deployments easily.

Lets add that!

handbook/engineering/distribution/goals.md
pecigonzalo: It depends on how you look at it, I think a simpler https://github.blog/2015-06-02-deploying-branches-to-github-com/ is highly desirable and gives a lot of visibility onto what is deploying, when it finishes, etc.

handbook/engineering/distribution/goals.md
pecigonzalo: Unresolving to highlight the conversation

handbook/engineering/distribution/goals.md
bobheadxi: I think we can the visibility on deployment while keeping the automated fashion of "what's in master is what is out there" (which I think is a lot less confusing than the model in the blog post, which is "any branch could be out there" - I'm also curious about whether they still use that model, given the post is 5 years old and some of the repos linked are now archived 🤔 )

pecigonzalo: I dont know if they do, and I would not map it 1 to 1, I think a simplified version of that works quite well. Its similar to what was shared here https://technology.riotgames.com/news/how-riot-games-uses-slack

ryanslade: 👍 from me on keeping it automatic. 

I think we should strive to have anything committed to `main` be ready for production and therefore can be deployed to any environment. We can use feature flags to actually enable / disable it per environment. For example if we want to demo a new feature on demo.sourcegraph.com before it's public.

ryanslade: I'm a bit torn between whether we actually need a `/`command. If deployment is automatic then I don't think we need it for that. Fast rollbacks are nice, but it means that we are out of sync with what is committed vs what is deployed. If CI was fast enough (I know that's a stretch) then a rollback could be done via a git revert.

pecigonzalo: There are 2 types of rollback from my point of view.

**Roll forward**: Revert the code, but you are actually just creating a new artifact that removes a feature/change
**Roll back**: Redeploy an existing artifact

ryanslade: In that case in an ideal world I would prefer Roll Forward 😄

keegancsmith: Roll back seems like the same thing, but is "instant" vs needing any artifacts to be built. I don't think roll forward is achievable.

Note: our infrastructure is in a repo. So we will always be tracking the `release` branch of that. So for example a roll back would a roll forward in the infra repo :P

keegancsmith (commented):
LGTM, glad we will work on this. I like how it is documented, via what our end state is for the developer/release engineer without dictating any implementation details.

handbook/engineering/distribution/goals.md
tsenart: A roll back could also just be a direct git revert without a PR?

pecigonzalo: IMO its still a "roll forward" as you have to rebuild the artifacts, you are creating a new version that lacks a feature, not restoring an older version.

bobheadxi (commented):
🚀

bobheadxi (approved):
LGTM 😁

tsenart (approved):

handbook/engineering/distribution/goals.md
```diff
@@ -4,11 +4,25 @@ Goals are continuously updated and reviewed. If you find these goals do not refl
 
 ## Medium-term goals (~3-6 Months)
 
+### [Improve deployment pipeline UX](https://github.com/orgs/sourcegraph/projects/96)
```
nicksnyder, line 7: Is there a problem statement beyond the negation of the desired outcomes?

pecigonzalo: There is, I forgot to add it. Ill create another PR to include that.

PR #1446, opened 2020-08-25
Tech writer onboarding

kghopson:
@christinaforney I couldn't figure out how to fix the issue with the numbering in Local Environment - Windows section, but please look over the rest (especially the Mac sections)

PR #1801, opened 2020-10-21
Improvements to onboarding docs

christinaforney:
Cleaning up and organizing information from https://github.com/sourcegraph/about/pull/1446

Changes:
- General
  - The general onboarding list was long and hard to read. I moved it to it's own page and cleaned up the sections.
  - Added a section for before starting at Sourcegraph
  - Cleaned up spending company money to remove redundancy and organize/format
- Product
  - Added resources and section to product onboarding about contributing to the docs
  - Extracted some processes to the technical writer page
  - ~Created a developing with Windows guide (TBD if this should be kept and/or moved)~ Removed


handbook/product/technical_writing/windows_development.md
```diff
@@ -0,0 +1,78 @@
+# Using a windows environment
```
christinaforney, line 1: @davejrt and @pecigonzalo - you were both mentioned as using PC's in the guide. Is the information on the Windows doc helpful to you at all? Do you think it will be useful to other teammates? Going forward, anyone on the product team is expected to be running a unix environment (windows is not acceptable).

pecigonzalo: WSL2 Is a UNIX environment, I would not be working directly over Windows, but I think its fully acceptable to use a Windows PC with WSL.

christinaforney: Ok, makes sense. I think your recommendation to always use a Linux environment and just follow those instructions is sufficient. I don't want to keep documentation that will lead to more confusion. I'll delete this page.

Thank you for all the clarification!

PR #1803, opened 2020-10-21
Fixed image src for Scott Raney in About page

Strum355:


ryanslade:
Oops, deployed instead of approving :)

PR #1819, opened 2020-10-22
Add PgHero docs

daxmc99:
Adds docs for PgHero accessing pgHero on Sourcegraph Cloud


handbook/engineering/deployments/postgresql.md
```diff
@@ -0,0 +1,27 @@
...
+
+### Sourcegraph.com specific
+
+You can also directly view database in [GCP](https://console.cloud.google.com/sql/instances?project=sourcegraph-dev)
+
+We currently run two seperate databases. The `sg-cloud` database is the primary database and the `sg-cloud-code-intel` is used by the code-intel team.
+
+Use the cloud sql proxy to connect to the database
+```
+  cloud_sql_proxy -instances=sourcegraph-dev:us-central1:sg-cloud-732a936743=tcp:5555
```
unknwon, line 17: From my experience, the portion `732a936743` in `sg-cloud-732a936743` changes every time it restarts? If so, how can we reliable get the correct name to use here?

daxmc99: Are you sure? This value is encoded in the terraform state and this name changing would cause terraform to report a state difference (and cause some [issues](https://sourcegraph.sgdev.org/search?q=sg-cloud-732a936743&patternType=literal) ). If this was a k8s pod that would be true but this is the cloud sql connection id and it should not change.

unknwon: I am not sure :) That's why asked.

> If this was a k8s pod that would be true but this is the cloud sql connection id and it should not change.

I see, I got messed up with k8s 😂

PR #1829, opened 2020-10-23
Add evaluating values to hiring process

christinaforney:
I think we should immediately start having every interviewer follow this process. 

@withdavidli - is it possible to add this as a separate section/field? It would be better if we didn't have to copy and paste every time.

### DIFF
--- a/handbook/people-ops/hiring/evaluating_values.md
+++ b/handbook/people-ops/hiring/evaluating_values.md
@@ -2,27 +2,17 @@
 
 We expect every teammate, team, and our company to live up to the [Sourcegraph values](../../../company/values.md). It is therefore important that we evaluate every candidate on the values to determine how successful they will be within Sourcegraph.
 
-It is the responsibility of every interviewer to think about these values when interviewing a candidate. When submitting feedback for a candidate, copy the list of values into your notes and score the candidate out of 5 based on the rubric.
+We are actively working on creating a formal process for values interviewing ([Ops RFC 7](https://docs.google.com/document/d/1d1a53dEP-k5Qsjv7BOAGq9lg4RL4Br7wg-weglJjHwY/edit)) to create a structured set of questions that are used consistently across the organization to evaluate value alignment.
 
 ## Rubric
 
+In the short-term, the Product team is experimenting with having a values score card inside of feedback forms that uses the following rubric:
+
 | Score | Description |
 | --- | --- |
-| -1 | Exhibited behavior that is contrary to this value. |
-| 0 | No evidence towards or against this value (neutral). |
-| 1 | Demonstrated this value with evidence. |
-| 2 | Exceptionally demonstrated this value, with abundant evidence. |
+| 👎👎 | Exhibited behavior that is contrary to this value. |
+| 👎 | No evidence towards or against this value (neutral). |
+| 👍 | Demonstrated this value with evidence. |
+| 👍👍 | Exceptionally demonstrated this value, with abundant evidence. |
 
-## Values to copy into feedback
-
-Use this plaintext format for easy addition to the Lever feedback area.
-
-```markdown
-- Be customer driven:
-- Work as a team:
-- High agency:
-- High quality:
-- Open and transparent:
-- Continuously grow:
-- Be a good human:
-```
+There is a text box for filling in any notes about the above value ratings.
--- a/handbook/people-ops/hiring/evaluating_values.md
+++ b/handbook/people-ops/hiring/evaluating_values.md
@@ -1,0 +1,29 @@
+# Evaluating values in an interview
+
+We expect every teammate, team, and our company to live up to the [Sourcegraph values](../../../company/values.md). It is therefore important taht we evaluate every candidate on the values to determin how successful they will be within Sourcegraph.
+
+It is the responsibility of every interviewer to think about these values when interviewing a candidate. When submitting feedback for a candidate, copy the list of values into your notes and score the candidate out of 5 based on the rubric.
+
+## Rubric
+
+Scored out of 5:
+
+- 1 - Exhibited behavior that is contrary to this value
+- 2 - I did not have any evidence towards or against this value (neutral)
+- 3 - Gave the impression they have this value, but lacking evidence
+- 4 - Demonstrated this value with evidence
+- 5 - Exceptionally demonstrated this value, with abundent evidence
+
+## Values to copy into feedback
+
+Use this plaintext format for easy addition to the Lever feedback area.
+
+```markdown
+- Be customer driven:
+- Work as a team:
+- High agency:
+- High quality:
+- Open and transparent:
+- Continuously grow:
+- Be a good human:
+```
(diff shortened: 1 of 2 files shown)


withdavidli:
@christinaforney Is this for all interview roles? We can add this to the interview forms. Would you like the placement to be at the bottom or top of the form?

Given the way this commit is written I'm thinking format should be:

[name of value measured] -  drop down menu for [numeric score] [explanation of score]

christinaforney:
@withdavidli Yes! That format sounds perfect. I would put it at the bottom of the form so that the interview module feedback is at the top.

handbook/people-ops/hiring/evaluating_values.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Rubric
+
+Scored out of 5:
+
+- 1 - Exhibited behavior that is contrary to this value
+- 2 - I did not have any evidence towards or against this value (neutral)
+- 3 - Gave the impression they have this value, but lacking evidence
+- 4 - Demonstrated this value with evidence
+- 5 - Exceptionally demonstrated this value, with abundent evidence
```
nicksnyder, line 15: I like that you did this in your interview feedback, but I found these scores hard to quickly digest.

What do you think of this?

```suggestion
| Score | Description |
| --- | --- |
| -1 | Exhibited behavior that is contrary to this value. |
| 0 | No evidence towards or against this value (neutral), or there were both positive and negative examples. |
| 1 | Demonstrated this value with evidence. |
| 2 | Exceptionally demonstrated this value, with abundant evidence. |
```

We could also expand this

handbook/people-ops/hiring/evaluating_values.md
```diff
@@ -0,0 +1,29 @@
+# Evaluating values in an interview
+
+We expect every teammate, team, and our company to live up to the [Sourcegraph values](../../../company/values.md). It is therefore important taht we evaluate every candidate on the values to determin how successful they will be within Sourcegraph.
+
+It is the responsibility of every interviewer to think about these values when interviewing a candidate. When submitting feedback for a candidate, copy the list of values into your notes and score the candidate out of 5 based on the rubric.
+
+## Rubric
+
+Scored out of 5:
```
nicksnyder, line 9: I like that you did this in your interview feedback, but I found these scores hard to quickly digest.

What do you think of this?

```suggestion
| Score | Description |
| --- | --- |
| -1 | Exhibited behavior that is contrary to this value. |
| 0 | No evidence towards or against this value (neutral). |
| 1 | Demonstrated this value with evidence. |
| 2 | Exceptionally demonstrated this value, with abundant evidence. |
```

handbook/people-ops/hiring/evaluating_values.md
nicksnyder, line 15: I like that you did this in your interview feedback, but I found these scores hard to quickly digest.

What do you think of this?

```suggestion
| Score | Description |
| --- | --- |
| -1 | Exhibited behavior that is contrary to this value. |
| 0 | No evidence towards or against this value (neutral). |
| 1 | Demonstrated this value with evidence. |
| 2 | Exceptionally demonstrated this value, with abundant evidence. |
```

nicksnyder: When scoring, I would want to have a brief explanation that points to the points of evidence for or against (which can be explained in more detail elsewhere in the feedback)

nicksnyder: Other note: I deleted "Gave the impression they have this value, but lacking evidence" because I think it risks bias and also I don't think we need more granularity on the margin. We want clear evidence and the rubric should highlight when we don't have that.

handbook/people-ops/hiring/evaluating_values.md
```diff
@@ -0,0 +1,29 @@
...
+## Values to copy into feedback
+
+Use this plaintext format for easy addition to the Lever feedback area.
+
+```markdown
+- Be customer driven:
+- Work as a team:
+- High agency:
+- High quality:
+- Open and transparent:
```
nicksnyder, line 26: Does good communication fall under this?

handbook/people-ops/hiring/evaluating_values.md
christinaforney: Love this suggestion!

handbook/people-ops/hiring/evaluating_values.md
christinaforney: Yes, I think so. Where can we clarify this?

nicksnyder: My instinct is that we should come up with rubric that is informed by our values, but not necessarily exactly the values 1-1, because that will make it easier to evaluate candidates.

For example:

- Customer(/outcome/metrics?) driven
- Collaborative (I think this covers both work as a team and be a good human)
- High agency
- Iterative
- Effective communication (clear, succinct)
- Growth mindset
- Authentic and kind

christinaforney: After setting this aside for a bit, I actually think feeling like we need to move away from our values to other rubrics means that our values are not quite right. We should live by and evaluate base on our values and use those as the rubric, otherwise this will lead to confusion and misalignment over time.

nicksnyder: I am fine with that. I think just having more alignment/training/details on what these value mean will help (in the values handbook page)

christinaforney: I completely agree. I don't think it hurts to add this change (or the questions) in the meantime while we work on the RFC proposal to get that in place. Perhaps adding a note here to indicate that the work you mention is in progress?

christinaforney:
@mercadon @nicksnyder - see updated with what we just discussed.

handbook/people-ops/hiring/evaluating_values.md
```diff
@@ -0,0 +1,18 @@
...
+In the short-term, the Product team is experimenting with having a values score card inside of feedback forms that uses the following rubric:
+
+| Score | Description |
+| --- | --- |
+| 👎👎 | Exhibited behavior that is contrary to this value. |
+| 👎 | No evidence towards or against this value (neutral). |
+| 👍 | Demonstrated this value with evidence. |
+| 👍👍 | Exceptionally demonstrated this value, with abundant evidence. |
+
+There is a text box for filling in any notes about the above value ratings.
```
nicksnyder, line 18:
```suggestion
| 👎👎 | Abundant evidence that this person 's behavior isn't aligned with this value. |
| 👎 | Evidence that this person 's behavior isn't aligned with this value.  |
| 👍 | Demonstrated this value with evidence. |
| 👍👍 | Exceptionally demonstrated this value, with abundant evidence. |

There is a text box for filling in any notes about the above value ratings.

If your interview did not get signal on a value (or if you got conflicting signals), then you can choose to not rate the value and instead share your thoughts in the open text box.
```

nicksnyder (approved):

PR #1874, opened 2020-10-28
publicly disclosing current security policies

chayim:


ElizabethStirling:
Not sure how I feel about this, since much of this is not enforced, implemented, or verified in any way, AFAICT. I may be misunderstanding, but this seems more aspirational than we want the handbook to be. Let's chat about this.

dadlerj:
Having this in our handbook would be amazing, but I don't think I'm the right person to review the specific content. I agree with @ElizabethStirling 's comment as well, but I imagine there's some subset of this content that can be merged today, so we can have these pages as frameworks that we can add to as our security/IT/compliance functions begin to verify and enforce more of this over time.

chayim:
I'm going to slice them up, and start merging as appropriate.

nicksnyder:
ping (merge or close)

chayim:
Immediately after the Vanta meeting I'll know the answer. I want things in
our handbook, but if Vanta has templates and outputs that are more useful,
we'll use those.

chayim:
Opting to close for now as we'll issue these with Vanta **and** publish those to the Handbook. Closing and pruning out.

PR #1878, opened 2020-10-29
Removed Tion and added Jonah to CE team

macraig:


juliasourceress (approved):

juliasourceress:
Thanks for this update & good catch, @macraig! cc @justdueck to add into our general process if you don't mind.

PR #1890, opened 2020-10-30
Add documentation on Prometheus and Cloudflare 

ElizabethStirling:


handbook/engineering/observability/query.md
```diff
@@ -0,0 +1,15 @@
+# Manual searching developer guide
+
+> **Note:** Looking for _how to monitor Sourcegraph?_ See the [observability documentation](https://docs.sourcegraph.com/admin/observability).
+
+This document gives a quick overview of how to manually search for data that isn't currently represented or displayed in [Grafana dashboards](https://docs.sourcegraph.com/admin/observability/metrics). Note that, in most cases, writing manual queries indicates that you should be [adding monitoring](https://about.sourcegraph.com/handbook/engineering/observability/monitoring#adding-monitoring) so that the next person who needs to find that data doesn't need to repeat your work.
```
slimsag, line 5:
```suggestion
This document gives a quick overview of how to manually search for data that isn't currently represented or displayed in [Grafana dashboards](https://docs.sourcegraph.com/admin/observability/metrics). If the data you are querying is to monitor the health of Sourcegraph, you should be [adding monitoring](https://about.sourcegraph.com/handbook/engineering/observability/monitoring#adding-monitoring) instead so that the next person and site admins/customers alike don't need to repeat your work. 
```

Rephrasing slightly because we only want data in our monitoring stack if it can be alerted on, non-alertable (read: non-actionable) data can be queried ad-hoc (so this doc is great) but is usually a symptom of us not having a good way to alert on that data for some reason.

slimsag (approved):
Great, thanks for documenting this!

handbook/engineering/observability/query.md
```diff
@@ -0,0 +1,15 @@
+# Manual searching developer guide
```
bobheadxi, line 1: There is already some stuff about finding Prometheus metrics in https://about.sourcegraph.com/handbook/engineering/observability/monitoring#prometheus-and-alertmanager which you linked to - I would personally strongly prefer trying to keep the monitoring stuff in one place, and just have this guide there, so it can be maintained in one place.

For example, we have ["Adding monitoring"](https://about.sourcegraph.com/handbook/engineering/observability/monitoring#adding-monitoring), maybe we can have a "Finding monitoring" section? 😄

handbook/engineering/observability/cloudflare.md
```diff
@@ -0,0 +1,37 @@
+# Sourcegraph Cloudflare monitoring developer guide
```
bobheadxi, line 1: Can you add a brief description and link to this in the [Sourcegraph Cloud section in the architecture](https://about.sourcegraph.com/handbook/engineering/observability/monitoring_architecture#sourcegraph-cloud)?

handbook/engineering/observability/query.md
ElizabethStirling: Makes sense, thanks for the feedback!

handbook/engineering/observability/query.md
ElizabethStirling: Good call, thanks!

handbook/engineering/observability/cloudflare.md
ElizabethStirling: Added

PR #1891, opened 2020-10-31
Draft PR to move the web team roadmap

Joelkw:
I'm not sure what makes sense here in the long term, but copied the Cloud structure for now and kicked off because of [this thread](https://sourcegraph.slack.com/archives/C0C324C91/p1603917865276000). 

The thing I don't love is that we're linking to the iteration google doc and board in two places now. But I didn't want the goals page to not have direct links to all goals, and I also think it made sense to have the iterations linked to in the "iteration process" section of the web team index page. I'm happy to leave this PR in draft for a bit if there's another change coming that Christina may have alluded to in her slack message in that thread.


christinaforney (approved):
This looks great! There will be more changes (as you mention) to clean this up a bit more, but this is a good incremental step.

felixfbecker (approved):

PR #1921, opened 2020-11-02
Document distribution private Slack

pecigonzalo:
We will be experimenting with a private Slack channel for shortening the loop on some topics that come out of retrospectives or other team communication (scheduling, how we plan, how/when we do updates, etc). The channel will be #distribution-team but our default communication is still #distributioneers and most communication should happen over this channel.

### DIFF
--- a/handbook/engineering/distribution/index.md
+++ b/handbook/engineering/distribution/index.md
@@ -4,14 +4,14 @@
 
 The distribution team is responsible for making Sourcegraph easy to deploy, scale, monitor, and debug. We solve the most challenging problems our customers face deploying Sourcegraph on-premise in a variety of environments, and on [Sourcegraph.com](https://sourcegraph.com/search) - the largest Sourcegraph instance in the world.
 
-## Communication
+## Contact
 
 - Slack: #distributioneers channel or @distribution
 - File issues: [team/distribution](https://github.com/sourcegraph/sourcegraph/issues/new?labels=team/distribution) label
 
-Our team has two Slack channels, one public (#distributioneers) and one private (#distribution-team). **Our default is to use the public channel**.
+## Team communication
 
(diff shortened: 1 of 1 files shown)


handbook/engineering/distribution/index.md
```diff
@@ -4,11 +4,15 @@
 
 The distribution team is responsible for making Sourcegraph easy to deploy, scale, monitor, and debug. We solve the most challenging problems our customers face deploying Sourcegraph on-premise in a variety of environments, and on [Sourcegraph.com](https://sourcegraph.com/search) - the largest Sourcegraph instance in the world.
 
-## Contact
+## Communication
```
bobheadxi, line 7: It seems `Contact` is a standard header in all the team pages, we might want to avoid changing this

handbook/engineering/distribution/index.md
```diff
@@ -4,11 +4,15 @@
...
 
 - Slack: #distributioneers channel or @distribution
 - File issues: [team/distribution](https://github.com/sourcegraph/sourcegraph/issues/new?labels=team/distribution) label
 
+Our team has two Slack channels, one public (#distributioneers) and one private (#distribution-team). **Our default is to use the public channel**.
+
+The #distribution-team channel will be used for discussing internal team topics, like scheduling, retrospectives, how we do updates, etc.
```
bobheadxi, line 14: We should have a "Teams communication" section instead, like Campaigns does (they also reference a private channel): https://about.sourcegraph.com/handbook/engineering/campaigns#team-communication

pecigonzalo: Ill move it. Just for reference, I thought it would be good to have this at the top and in the same section given its just a small paragraph. This is also the reasoning behind the `Contact -> Communication`.

bobheadxi (approved):

PR #1952, opened 2020-11-05
Add roadmap items

macraig:
Split milestones and roadmap items.

### DIFF
--- a/handbook/engineering/code-intelligence/goals.md
+++ b/handbook/engineering/code-intelligence/goals.md
@@ -12,37 +12,39 @@ Progress on adoption and usage is tracked in our [Looker dashboard](https://sour
 
 **Milestones:**
 
-1. 🔄 Deliver precise Go code intelligence to 3 customers.
-1. 🔄 Deliver precise C++ code intelligence to 3 customers.
+1. 🔄 Deliver precise Go code intelligence to 3 organizations.
+    - ✅ Deliver to [*N<sub>1</sub>*][N1]
+    - 🔄 Deliver to [*N<sub>2</sub>*][N2]
+    - Deliver to [*N<sub>3</sub>*][N3]
+1. 🔄 Deliver precise C++ code intelligence to 3 organizations.
+    - Deliver to [*N<sub>4</sub>*][N4]
+    - Deliver to [*N<sub>5</sub>*][N5]
+    - Deliver to [*N<sub>6</sub>*][N6]
 1. 🔄 Support automatically indexing source code within Sourcegraph.
+    - 🔄 Provide best effort auto-indexing for supported languages [RFC 201](https://docs.google.com/document/d/1NPQs1s814LZjNXjPuavqC1N7hZR192DNtmSBmAeH9UY/edit)
 1. 🔄 Support incremental indexing of Go source code.
-1. 🔄 Deliver precise Java code intelligence to 3 customers.
-1. Deliver precise TypeScript code intelligence to 3 customers.
+    - 🔄 Add incremental indexing for monorepo support [RFC 170](https://docs.google.com/document/d/1NPu0Vc7FpdoYwCrtpnu-8KB4OPbw7L0KBTqw96JVc8w/edit#)
+1. 🔄 Deliver precise Java code intelligence to 3 organizations.
+    -  🔄 Add native Gradle support for lsif-java
+    - Deliver to [*N<sub>7</sub>*][N7]
+    - Deliver to [*N<sub>8</sub>*][N8]
+    - Deliver to (TBD)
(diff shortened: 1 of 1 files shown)


nicksnyder (commented):
Just curious what the motivation here in splitting these things is.

Keep an eye on the amount of work in progress. Right now it looks like we have 5 non-trivial things being worked on for 2.5 engineers (and that seems high).

efritz (approved):

macraig:
The split is just to reflect the difference between the milestones and the roadmap items (steps to complete the milestones).

I think that the amount looks high but at the same time some of them aren't full-time efforts (delivery to customers for example). Beyang is helping with customer delivery outreach so we have a bit more team capacity on that particular effort. Does that make more sense?

efritz:
> Keep an eye on the amount of work in progress. Right now it looks like we have 5 non-trivial things being worked on for 2.5 engineers (and that seems high).

Some of these in-progress items are delivery items which have a lot of down time while waiting for customer responses. Would it be better to section these types of tasks off into something that shows it's not taking 100% attention from the owner?

Maybe there's a better way to think about this class of tasks.

nicksnyder:
> Some of these in-progress items are delivery items which have a lot of down time while waiting for customer responses.

My worry would be that we are not being proactive enough here at getting these things delivered to customers. If we don't get a quick response from customers 1, 2, 3, then can we find customers 4, 5, 6 to reach out to? If customers aren't eager to adopt what we are building, is that signal that we aren't building the right thing?

The fact that these things do have a long cycle time is a problem in and of itself, that I think is worth fixing and not skipping over. This is probably a problem more for you to think about @macraig .

> Would it be better to section these types of tasks off into something that shows it's not taking 100% attention from the owner?

No, I think we need to focus on getting them done instead.

> The split is just to reflect the difference between the milestones and the roadmap items (steps to complete the milestones).

Yeah, I guess my question is that if the roadmap items are the steps to complete the milestones, then why not nest the steps under the milestones that they serve? Is the value that the roadmap can more accurately interleave the order of steps from different milestones?

macraig:
> The fact that these things do have a long cycle time is a problem in and of itself, that I think is worth fixing and not skipping over. This is probably a problem more for you to think about @macraig.

Sure, we'll also add it to our team sync conversation to decide on next steps.

> Yeah, I guess my question is that if the roadmap items are the steps to complete the milestones, then why not nest the steps under the milestones that they serve? Is the value that the roadmap can more accurately interleave the order of steps from different milestones?

Yes, but to your point nesting may be a clearer way to have it represented in the docs as we'll also have the roadmap in Productboard. Will revert to nested version and just link Productboard in the roadmap section (and later post an image of the Gantt chart so it's public)

handbook/engineering/code-intelligence/goals.md
```diff
@@ -1,39 +1,50 @@
...
+1. 🔄 Deliver precise Go code intelligence to 3 organizations.
+    - ✅ Deliver to [*N<sub>1</sub>*][N1]
+    - 🔄 Deliver to [*N<sub>2</sub>*][N2]
+    - Deliver to [*N<sub>3</sub>*][N3]
+1. 🔄 Deliver precise C++ code intelligence to 3 organizations.
+    - Deliver to [*N<sub>4</sub>*][N4]
+    - Deliver to [*N<sub>5</sub>*][N5]
+    - Deliver to [*N<sub>6</sub>*][N6]
 1. 🔄 Support automatically indexing source code within Sourcegraph.
+    - 🔄 Provide best effort auto-indexing for supported languages [RFC 201](https://docs.google.com/document/d/1NPQs1s814LZjNXjPuavqC1N7hZR192DNtmSBmAeH9UY/edit)
```
efritz, line 24: I've just marked this as implemented. I could use some help breaking out additional tasks as it's currently in the _long tail_ of making sure the service is reliable and useful.

efritz (approved):

christinaforney (approved):
I think this is an improvement, and in the spirit of iterating think we should get this merged and work through the current issues - that there is too much in progress and if we're not making progress/waiting on those things, we should be finding ways to be more proactive about it. @macraig - we can strategize together if you'd like!

handbook/engineering/code-intelligence/goals.md
macraig: Great! I'll change it to Done and add a TBD next step

macraig:
> I think this is an improvement, and in the spirit of iterating think we should get this merged and work through the current issues - that there is too much in progress and if we're not making progress/waiting on those things, we should be finding ways to be more proactive about it. @macraig - we can strategize together if you'd like!

Yes that would be great @christinaforney ! We talked about it a bit yesterday in our team sync and there are a few ideas we can discuss.

PR #1963, opened 2020-11-05
clarified campaigns roadmap; other handbook page tweaks

chrispine:
These changes:

- clarify our teams roadmapping process
- add links to roadmap docs/pages
- update our team's section of the product roadmap
- remove the section on our "new" UX
- move documentation links to bottom of page

### DIFF
--- a/handbook/engineering/campaigns/index.md
+++ b/handbook/engineering/campaigns/index.md
@@ -2,18 +2,6 @@
 
 <img src="https://storage.googleapis.com/sourcegraph-assets/campaigneros.png"  align=right alt="Campaigñeros Logo">
 
-User-facing documentation: https://docs.sourcegraph.com/campaigns
-
-Developer documentation: https://docs.sourcegraph.com/dev/background-information/campaigns
-
-### New, work-in-progress UX
-
-We're working on a new flow and a new way for people to create campaigns based on the beta feedback. See [#10921](https://github.com/sourcegraph/sourcegraph/pull/10921) and the following new doc pages (which are "pre-written" prior to being implemented to get us all on the same page about the changes we'll be making):
-
-- [Campaigns main doc page](https://docs.sourcegraph.com/@campaigns-new-flow/user/campaigns)
-- [Hello World Campaign guide](https://docs.sourcegraph.com/@campaigns-new-flow/user/campaigns/hello_world_campaign)
-- [Campaigns design doc](https://docs.sourcegraph.com/@campaigns-new-flow/dev/campaigns_design)
-
 ## Vision
 
 **Find code that needs to be changed and change it by running code**.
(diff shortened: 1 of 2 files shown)


handbook/engineering/campaigns/index.md
```diff
@@ -82,6 +70,18 @@ Our users use more than just their code host and Sourcegraph. They use ticket tr
 
 We want to discover what external systems our users want to use campaigns with, and ensure that we can integrate with them. For example, an organization that uses JIRA will likely want to be able to link tickets to campaigns and have the state updated as the campaign is executed.
 
+## Roadmap
+
+Our team's public roadmap is part of our larger [Product roadmap](https://about.sourcegraph.com/handbook/product/roadmap#campaigns).
+
+### Process
```
chrispine, line 77:
```suggestion
### Roadmap Process
```

LawnGnome (approved):

handbook/product/roadmap.md
```diff
@@ -55,11 +55,14 @@ section Web
 
 1. **Customer outreach to improve adoption of campaigns**
 1. User credentials ([RFC 242](https://docs.google.com/document/d/1SqoWWm1xs82QibrWwYsXmpmgweN6EpcKt1qXrRBjjlU/edit)), which will allow non-site-admins to create campaigns
+1. Better burndown charts
 1. Versioning/releasing of src-cli with respect to sg/sg
-1. Respect GitHub GraphQL rate limits
 1. Improved documentation of src-cli login process
+1. Add filtering/searching to campaign and changeset lists
+1. Ability to auto-merge changesets, depending on various conditions
+1. Ability to add default reviewers to changesets
```
eseliger, line 63: Maybe these two shouldn't be on the same level as the others as I think it's not yet clear what these two involve and they might be pushed back if too much effort(?)

eseliger (approved):

handbook/product/roadmap.md
chrispine: While that is true, these are currently our best guess as to what we want to tackle next, and I wanted this to be as accurate as we currently know. As we learn more (and possibly push one or both of these back), I will keep this list updated.

handbook/engineering/campaigns/index.md
```diff
@@ -82,6 +70,18 @@ Our users use more than just their code host and Sourcegraph. They use ticket tr
 
 We want to discover what external systems our users want to use campaigns with, and ensure that we can integrate with them. For example, an organization that uses JIRA will likely want to be able to link tickets to campaigns and have the state updated as the campaign is executed.
 
+## Roadmap
+
+Our team's public roadmap is part of our larger [Product roadmap](https://about.sourcegraph.com/handbook/product/roadmap#campaigns).
```
nicksnyder, line 75: @christinaforney We are moving toward inlining roadmaps in team pages, right?

handbook/engineering/campaigns/index.md
```diff
@@ -82,6 +70,18 @@ Our users use more than just their code host and Sourcegraph. They use ticket tr
...
+
+### Roadmap Process
+
+Our roadmapping process is a team effort. As we receive customer feedback, and as we come up with cool new features, we record these in one of three places. Customer feedback goes into our [team Productboard page](https://sourcegraph.productboard.com/feature-board/2104383-campaigns). Well-defined tasks live in our [team backlog](https://github.com/sourcegraph/sourcegraph/issues?q=is%3Aopen+is%3Aissue+label%3Ateam%2Fcampaigns+milestone%3ABacklog). And larger, less defined ideas live in the bottom section of our [private roadmap doc](https://docs.google.com/document/d/1zRTfK6mENKicfLwDaWgLk1dBvQVKDg-J7pwjGg8tpps/edit#), below the _fog of war_ line.
```
mrnugget, line 79: Side-note (not a blocker): does the private roadmap doc need to be private? I don't think there's sensitive information in there and potential candidates often enjoy looking at the roadmap. I'd rather have us "default to public" than "default to private and maybe make it public"

mrnugget (approved):

handbook/engineering/campaigns/index.md
chrispine: My default is "private when it relates to customer data" which is something that was very much drilled into me from my last job (to the point where we couldn't even say "$COMPANY is a customer of ours"). 😄 That roadmap doc has information about feature requests from several customers. So until I hear otherwise, I'm going to stick with that default.

Otherwise, I totally agree with you, and would have no problem making it public.

nicksnyder: More context:

Things we don't want to be public include customer names (I see some in that doc), company financials or sensitive metrics (e.g. revenue, I don't see any such things).

Even if customer X is public (i.e. they are on our homepage), we would not want to attribute specific feedback/quotes to that customer (since they might not have given permission to us to share that quote).

We are open, but we assume that our customers are not open.

mrnugget: Got it! Thanks for both answers :)

chrispine: > We are open, but we assume that our customers are not open.

Great way to put that, thanks!

handbook/engineering/campaigns/index.md
chrispine: I'm going to go ahead and merge this, but am happy to move things around later based on feedback here.

christinaforney: Yes. Updated for all teams here: https://github.com/sourcegraph/about/pull/1984

PR #1754, opened 2020-10-12
Experiment: Stephen becomes internal contributor, focusing on high-impact low-cost wins

slimsag:
I will be taking on a new direction at Sourcegraph focusing more heads-down on product development, specifically on high-value low-cost wins. This is an experiment, and means I will be effectively departing my regular responsibilities on the Distribution team and elsewhere.

More context: https://sourcegraph.slack.com/archives/C0EPTDE9L/p1602530986086900

PR #1104, opened 2020-06-25
distribution roadmap

slimsag:
This is an early draft, I hope to finish this by EOW.

I suggest reading these two first:

- https://github.com/sourcegraph/about/blob/sg/distribution-roadmap/handbook/engineering/distribution/index.md#why-the-distribution-team-not-the-sre-team-infrastructure-team-etc
- https://github.com/sourcegraph/about/blob/sg/distribution-roadmap/handbook/engineering/distribution/product/index.md

Then I suggest reading the actual product roadmap:

- https://github.com/sourcegraph/about/blob/sg/distribution-roadmap/handbook/engineering/distribution/product/roadmap.md

@pecigonzalo please spam me with as many questions/comments as you have, or we can go over this live soon :)

PR #1973, opened 2020-11-07
distribution: proposed updates to team goals

slimsag:
Prior to switching to my new role, I wanted to re-review Distribution team goals and ensure that I did not have any major suggestions. Ultimately, I trust @pecigonzalo and the rest of the team here though :)

Meta: I didn't have time to _really_ review and read these goals until now. That is a big signal that (a) I am an outlier or (b) others probably haven't either. We should probably find a way to talk about / review these more frequently.

### DIFF
--- a/handbook/engineering/distribution/goals.md
+++ b/handbook/engineering/distribution/goals.md
@@ -108,7 +108,7 @@ Many customers of Sourcegraph today are still running a single-container `source
   - RFC approval.
   - ...TBD...
 
-### Upgrades across multiple Sourcegraph versions should be easier
+### Support upgrading across multiple Sourcegraph versions
 
 Upgrading from 3.13 -> 3.17 requires you perform 4 individual upgrades today (3.14 -> 3.15 -> 3.16 -> 3.17) which is extremely painful and time consuming for site admins, especially so given how time consuming our upgrade process is in general. We would like to make upgrades across multiple Sourcegraph versions easier.
 
--- a/handbook/engineering/distribution/goals.md
+++ b/handbook/engineering/distribution/goals.md
@@ -4,7 +4,7 @@ Goals are continuously updated and reviewed. If you find these goals do not refl
 
 ## Medium-term goals (~3-6 Months)
 
-### [Improve deployment pipeline UX](https://github.com/orgs/sourcegraph/projects/96)
+### [Improve internal deployment pipeline UX](https://github.com/orgs/sourcegraph/projects/96)
 
 Our existing deployment pipelines to our Sourcegraph instances (such as Sourcegraph Cloud) has several usability problems - for example, it is hard for engineers to identify when a commit was deployed to an environment or which deployment is currently running in a particular environment. We want to improve the deployment experience, making sure we can deploy with confidence and can easily understand in which stage of the pipeline a change currently is.
 
@@ -20,12 +20,12 @@ Our existing deployment pipelines to our Sourcegraph instances (such as Sourcegr
   - TBD: We can trigger rollbacks and deployments via a `/` command in Slack.
 - **Milestones**: TBD
 
-### Any engineer at Sourcegraph can create a release for all of our supported deployment types by running a single command
+### Any engineer at Sourcegraph can create a release for all of our supported deployment types by running a single action
 
 Creating a new release for our deployments is currently a semi-automated process, which requires several manual steps and synchronizing our versioned artifacts (Sourcegraph, Kubernetes manifests, docker-compose manifests, etc). We want to enable any engineer to perform a release as often as needed, to enable this we want to make releasing Sourcegraph a simple, automated process.
 
 - **Owner**: Dave, Robert
-- **Status**: In progress. Estimated completion by end of 2020.
+- **Status**: In progress. Estimated completion by end of FY21-Q1.
 - **Outcomes**:
   - Releases can be triggered by a single manual step.
   - All supported deployment types are released at the same time with the same command.
@@ -37,7 +37,7 @@ Creating a new release for our deployments is currently a semi-automated process
   - [Releases can be done in a single day](https://github.com/orgs/sourcegraph/projects/90). **In progress**
   - [Enable continuous regression tests on `main`](https://github.com/orgs/sourcegraph/projects/90). **In progress**
   - [Ensure relevant engineers are notified of broken builds](https://github.com/orgs/sourcegraph/projects/90). **In progress**
-  - Releases can be done automatically with a single `/` command via Slack. _Estimated: 2020_
+  - Releases can be done automatically with a single action (e.g. CLI command, `/` command in Slack, etc.). _Estimated: 2020_
 
 ### [Upgrades between releases are easy to perform](https://github.com/orgs/sourcegraph/projects/71)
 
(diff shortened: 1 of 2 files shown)


handbook/engineering/distribution/goals.md
```diff
@@ -4,7 +4,7 @@ Goals are continuously updated and reviewed. If you find these goals do not refl
 
 ## Medium-term goals (~3-6 Months)
 
-### [Improve deployment pipeline UX](https://github.com/orgs/sourcegraph/projects/96)
+### [Improve internal deployment pipeline UX](https://github.com/orgs/sourcegraph/projects/96)
```
slimsag, line 7: Just a minor clarification, because I thought this was talking about the improved deployment process for customers at first (Dhall).

handbook/engineering/distribution/goals.md
```diff
@@ -20,12 +20,12 @@ Our existing deployment pipelines to our Sourcegraph instances (such as Sourcegr
   - TBD: We can trigger rollbacks and deployments via a `/` command in Slack.
 - **Milestones**: TBD
 
-### Any engineer at Sourcegraph can create a release for all of our supported deployment types by running a single command
+### Any engineer at Sourcegraph can create a release for all of our supported deployment types by running a single action
```
slimsag, line 23: I think "a single command", and "slack command" at that, is not so important. I suspect that we would be OK with this being a single action of some sort (whether that is a CLI command, creating a GitHub issue, etc.) as long as the rest is automated.

handbook/engineering/distribution/goals.md
```diff
@@ -20,12 +20,12 @@ Our existing deployment pipelines to our Sourcegraph instances (such as Sourcegr
...
 
 Creating a new release for our deployments is currently a semi-automated process, which requires several manual steps and synchronizing our versioned artifacts (Sourcegraph, Kubernetes manifests, docker-compose manifests, etc). We want to enable any engineer to perform a release as often as needed, to enable this we want to make releasing Sourcegraph a simple, automated process.
 
 - **Owner**: Dave, Robert
-- **Status**: In progress. Estimated completion by end of 2020.
+- **Status**: In progress. Estimated completion by end of FY21-Q1.
```
slimsag, line 28: I updated this based on your earlier comment in https://github.com/sourcegraph/about/pull/1754#discussion_r503560574 @pecigonzalo

handbook/engineering/distribution/goals.md
```diff
@@ -64,17 +64,17 @@ We will start by looking at our Kubernetes deployment and working on an easier u
   - Design customization workflow.
     - Potentially Kustomize would still be used for last-mile changes and non-standard derivations.
 
-### Improve the debugging and troubleshooting process
+### Further standardize the debugging and troubleshooting process
```
slimsag, line 67: I am not content with how under-specified this goal is, I would suggest specifying it further (as I have done here) or removing it from this page until we can specify it further.

I also felt the prior description could leave anyone coming across this page to believe that we have no consistent tooling to debug issues, but this does a great disservice to the fact that we have been investing in exactly that (standardized Grafana, Jaeger, alerting, etc. across all deployments) for the past year and have made great strides there :)

I agree there is more to do, though, so I do think this is a good goal - we should just try to specify it further. I have tried to do that here.

handbook/engineering/distribution/goals.md
```diff
@@ -64,17 +64,17 @@ We will start by looking at our Kubernetes deployment and working on an easier u
...
 - **Outcomes**:
   - We can categorize and capture the amount of effort spent on different incident types.
-  - We can provide a straightforward set of tools to collect initial debugging and deployment information.
-  - TBD
+  - We provide a straightforward set of processes/tools to collect initial debugging and deployment information (configuration, type, size, diff from upstream, recently-firing alerts, logs, etc.)
+  - Alerts in services owned by other teams at Sourcegraph are well-defined, extensive, and have excellent documentation site admins can understand.
+  - The tooling we rely on to debug our own infrastructure is aligned with the tooling we use to debug customer deployments.
 - **Milestones**: TBD
 
 ### [Split Cloud infrastructure into separate GCP projects](https://github.com/orgs/sourcegraph/projects/92)
```
slimsag, line 80: What is the status of this? It seems almost completely done except some docs?

Also, can we create a "past goals" section in this doc (or in a `past_goals.md` file) which includes all the goals we successfully did thus far? I think that is a great resource for looking historically, and for sharing with candidates when we start hiring more folks.

Check out https://github.com/sourcegraph/about/pull/1104/files#r519076983 which includes my review of the roadmap I put out earlier this year - it's really cool to see in retrospect how many of these points we did make a lot of progress on. We should find a way to communicate this (internally and to the rest of the org)

handbook/engineering/distribution/goals.md
```diff
@@ -93,6 +93,31 @@ We will initially focus on reducing the time it takes to collect troubleshooting
   - ~~Create code to bootstrap new projects.~~ Done: 3.21
   - Document project and folder usage guidelines.
 
+### Site admins use reliable methods like Docker Compose or Kubernetes for production deployments
```
slimsag, line 96: I copied this over from https://github.com/sourcegraph/about/pull/1104/files - I think this is a valuable goal worth calling out still here (even if it turns out to be something I work on)

handbook/engineering/distribution/goals.md
```diff
@@ -93,6 +93,31 @@ We will initially focus on reducing the time it takes to collect troubleshooting
...
+- **Outcomes**:
+  - Customer deployments become more reliable in performance, stability, uptime, etc.
+  - No customers are using the unreliable `sourcegraph/server` for their production deployment.
+  - Customers upgrade to Docker Compose or Kubernetes deployments without any major trouble.
+- **Milestones**:
+  - ~~RFC creation.~~
+  - RFC approval.
+  - ...TBD...
+
+### Upgrades across multiple Sourcegraph versions should be easier
```
slimsag, line 111: I copied this over from https://github.com/sourcegraph/about/pull/1104/files - I think this is a valuable goal worth calling out still here (even if it turns out to be something I work on)

handbook/engineering/distribution/goals.md
slimsag: I created another PR to add a past_goals.md doc - I think it's super interesting! https://github.com/sourcegraph/about/pull/1974

handbook/engineering/distribution/goals.md
pecigonzalo: This was "intentionally" under-specific as it was what we knew at the time, and its a TBD goal in many of its aspects.
This goal as discussed at the time was not about our monitoring tools, it was about a set of tools/scripts/etc we can provide to customers to bundle their logs/metrics/etc and send them to us (configuration, type, size, diff from upstream, etc) and the process to capture what issues are related to, so we can prioritize improving X.

I think this change fundamentally changes the goal and would split this into its own PR.

handbook/engineering/distribution/goals.md
pecigonzalo: The status is reflected in the goal, its partially missing `Document project and folder usage guidelines.` and filling this outcome `We can set spending limits for dynamic environments.`. I dont think it should be moved yet, unless we choose to de-scope those.

handbook/engineering/distribution/goals.md
pecigonzalo:
```suggestion
### Support upgrading across multiple Sourcegraph versions
```
I think the current header could be mixed with `Upgrades between releases are easy to perform`.

pecigonzalo (commented):
LGTM in general, I left 👍 or commends in the sections I think we need to change.

Thanks for bringing this over.

handbook/engineering/distribution/goals.md
slimsag: In this case, I will keep both the original goal here in its under-specified state *and* the new goal I have listed here, and we can improve/merge them later if that sounds good

handbook/engineering/distribution/goals.md
slimsag: Got it, `We can set spending limits for dynamic environments.` makes it more clear to me there is more than just "follow up on docs" here, any objection to me adding that to "milestones"?

handbook/engineering/distribution/goals.md
```diff
@@ -93,6 +93,31 @@ We will initially focus on reducing the time it takes to collect troubleshooting
   - ~~Create code to bootstrap new projects.~~ Done: 3.21
   - Document project and folder usage guidelines.
```
slimsag, line 94:
```suggestion
  - Document project and folder usage guidelines.
  - Set spending limits for dynamic environments.
```

handbook/engineering/distribution/goals.md
slimsag: https://github.com/sourcegraph/about/pull/1973/files#r520010960

handbook/engineering/distribution/goals.md
slimsag: nvm, I will just keep the old one - having both is confusing.

uwedeportivo (approved):

daxmc99 (approved):

handbook/engineering/distribution/goals.md
christinaforney, line 28: FYI our current quarter ending Feb 2021 is FY22-Q1. See @dadlerj's recent change to the definition: https://about.sourcegraph.com/handbook/communication#writing

PR #1998, opened 2020-11-13
releases: remove milestone close, some tags, config compile

bobheadxi:
* Tags no longer need to be created, milestones no longer need to be closed, due to changes in https://github.com/sourcegraph/sourcegraph/pull/15743 (closes https://github.com/sourcegraph/sourcegraph/issues/15323)
* Config no longer needs to be compiled, due to changes in https://github.com/sourcegraph/sourcegraph/pull/15729 (closes https://github.com/sourcegraph/sourcegraph/issues/15562)

### DIFF
--- a/handbook/engineering/releases/patch_release_issue_template.md
+++ b/handbook/engineering/releases/patch_release_issue_template.md
@@ -16,6 +16,10 @@ Arguments:
 
 ---
 
+**Note:** All `yarn run release ...` commands should be run from folder `dev/release`.
+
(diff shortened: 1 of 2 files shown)


pecigonzalo:
> Tags no longer need to be created
This is because they are created automagically and not because they are not longer needed correct?

bobheadxi:
> This is because they are created automagically and not because they are not longer needed correct?

Yep! This was added in https://github.com/sourcegraph/sourcegraph/pull/15743

PR #2001, opened 2020-11-13
Add goal to support large scale organizations

poojaj-tech:
Search goal and milestones update to create a ~6 month vision/plan. 

TODO:

- [x] Review with team in Monday sync on 2020-11-16
- [ ] Fix goal title (see comment from Christina)
- [ ] Update the problem statement to describe the current experience of users at large scale organizations
- [ ] Finish the Outcomes section
- [ ] Milestones should be written as achievable points in time as we work towards the end goal
  - [ ] Some milestones will require gathering baseline metrics
  - [ ] Milestones should be linked to ProductBoard sections with customer data.

### DIFF
--- a/handbook/engineering/search/goals.md
+++ b/handbook/engineering/search/goals.md
@@ -30,7 +30,7 @@ TODO: What does this look like? Describe this in prose using descriptions from t
 - Search is optimized for the cloud, where hundreds of a company’s private repos will live alongside hundreds of thousands of public open source repos.
 - We have a learnable, comprehensive code search query language; performant and reliable search across all result types we offer; a polished product that handles failure cases properly; and a personalized experience for users that surfaces their most relevant code.
 
-#### Milestones (currently unordered)
+#### Milestones (WIP: currently unordered and incomplete)
 
 - All basic search types (indexed, regular expression, structural) are fast at this scale.
     - What is fast?
--- a/handbook/engineering/search/goals.md
+++ b/handbook/engineering/search/goals.md
@@ -1,57 +1,44 @@
+
+
+
 # Search goals and priorities
 
 ## Goals
 
-Goals help us achieve our [vision](index.md#vision) for search.
+### Support large scale organizations
 
-### Mature query language
-
-**Problem:** The Sourcegraph query language does not support the growing complexity of searches being run by our users.
-
-Examples:
-
-- I want to only show examples from repositories that have been modified in the last 3 months so I see the most recent solutions.
-- I want to find repositories that have not been modified in over a year so I can deprecate them.
-- TODO: add more examples.
-
-**Success criteria**
-
-TODO table of use cases / capabilities we want to achieve, comparing Sourcegraph with other search tools
-
-**Planned work**
-
-1. Search expressions
-1. Search language rules engine
-
-### Users trust Sourcegraph for fast and reliable search results at scale
-
-**Problem:** Search is too slow or frequently times out at the scale some of our customers care about, and at the scale we want Sourcegraph Cloud to reach. In large-scale codebases, running the same search several times yields varying result counts, with no insights into the underlying reasons. Customers cannot trust that their codebase was exhaustively searched.
-
-**Success criteria**
-
-- **By EOY 2020**: Latency for the set of search queries run by search-blitz is less than TODO Nms with 500k repositories indexed on Sourcegraph Cloud
-    - 2020-11-05: TODO
-- **By EOY 2020**: P99 time to first result is less than TODO Nms across all search types on Sourcegraph Cloud
-    - 2020-11-05: TODO
+**Definition of large scale:**
+Customers with ([looker](https://sourcegraph.looker.com/looks/436))
+- Git Directory size > 100 GB and < 50 TB
+- \> 1 billion LOC and < 5 billion LOC
 
 
-**Planned work**
 
-1. Scale indexed text search to 500k repositories
-1. Streaming search
+**Problem:** Large customers expect search to be fast at their scale. Customers expect to be able to search over their codebase intuitively, and want their first search result quickly. For larger organizations, customers care about getting all search results for important changes in their codebase. Supporting search at this scale also helps work towards supporting the Cloud growth efforts.
 
-### Sourcegraph proactively notifies of important code changes
+**Outcome:** What does this look like? Describe this in prose using descriptions from the team: [Search team 6 month vision](https://docs.google.com/document/d/1iiYCKK5D2PTVzzFmTF1OHl5SNLVkYfbOfyrCcoYM_24/edit#heading=h.bi6mdia4vr7w) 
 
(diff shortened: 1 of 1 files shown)


handbook/engineering/search/goals.md
```diff
@@ -2,64 +2,52 @@
 
 ## Goals
 
-Goals help us achieve our [vision](index.md#vision) for search.
+### Support large scale organizations
```
christinaforney, line 5: I would like us to come up with a better title for the goal, ideally so it is more time based (has a clear "ending"). Also, this almost makes it seem like we don't support large scale _at all_, which is not true. We want this to capture that we are targeting an excellent experience and raising the bar for large organizations. I can't think of a better suggestion at the moment, would love some ideas.

handbook/engineering/search/goals.md
```diff
@@ -2,64 +2,52 @@
...
+- Total lines of code (LOC) between 1 billion and 5 billion
 
-**Success criteria**
+See [looker](https://sourcegraph.looker.com/looks/436) for current customer data.
 
-TODO table of use cases / capabilities we want to achieve, comparing Sourcegraph with other search tools
+#### Problem
 
-**Planned work**
+Large customers expect search to be fast at their scale. Customers expect to be able to search over their codebase intuitively, and want their first search result quickly. For larger organizations, customers care about getting all search results for important changes in their codebase. Supporting search at this scale also helps work towards supporting the Cloud growth efforts.
```
christinaforney, line 20: Rather than talking about expectations, let's talk about the current experience, listing current metrics. 

For example, something like the following (but worded better to sound more fluid and like the story of the current experience).

```
Customers with more than X LOC experience timeouts when performing searches without limiting total repositories. Currently, structural search only works when scoped to < 50 repositories. 
```

handbook/engineering/search/goals.md
```diff
@@ -2,64 +2,52 @@
...
+- Make it clear to the passing naive hacker that they couldn’t write something which replaces us in a few weekends / just use grep.
+- Our search experience is consistent, reliable, and predictable: IE Sourcegraph search does what it says and says what it does.
+- We understand more about each individual user and tailor their results so that results are ranked to promote the languages and repositories they are most likely to care about.
+- Search queries and their results become a frictionless starting point for creating all the other business objects our users will start to value… monitors, insights, campaigns and /next/.
+- Search is optimized for the cloud, where hundreds of a company’s private repos will live alongside hundreds of thousands of public open source repos.
+- We have a learnable, comprehensive code search query language; performant and reliable search across all result types we offer; a polished product that handles failure cases properly; and a personalized experience for users that surfaces their most relevant code.
 
-**Success criteria**
+#### Milestones (currently unordered)
```
christinaforney, line 33:
```suggestion
#### Milestones (WIP: currently unordered and incomplete)
```

handbook/engineering/search/goals.md
```diff
@@ -2,64 +2,52 @@
...
+Organizations with:
 
-- I want to only show examples from repositories that have been modified in the last 3 months so I see the most recent solutions.
-- I want to find repositories that have not been modified in over a year so I can deprecate them.
-- TODO: add more examples.
+- A Git directory size between 100 GB and 50 TB
+- Total lines of code (LOC) between 1 billion and 5 billion
 
-**Success criteria**
+See [looker](https://sourcegraph.looker.com/looks/436) for current customer data.
```
keegancsmith, line 16: Is there a look which just includes the customers matching the above description? I'd be interested to know how many fit into this category. My main concern is if it is too many, we wouldn't get the benefits of having specific customers in mind when deciding what to work on.

handbook/engineering/search/goals.md
```diff
@@ -2,64 +2,52 @@
...
+#### Definition of large scale
 
-Examples:
+Organizations with:
 
-- I want to only show examples from repositories that have been modified in the last 3 months so I see the most recent solutions.
-- I want to find repositories that have not been modified in over a year so I can deprecate them.
-- TODO: add more examples.
+- A Git directory size between 100 GB and 50 TB
+- Total lines of code (LOC) between 1 billion and 5 billion
```
keegancsmith, line 14: Should we explicitly call out differences between monorepos vs lots of repos? The sort of solutions that help monorepos don't normally help lots of repos. I would think focussing on monorepos (or possibly having a part of the team focus on monorepos) would make a lot of sense.

Although anything new we make needs to support both use cases (eg if we did indexed commit search).

poojaj-tech: Good call out. We used Git dir size and LOC as  metrics (versus number of repos) because it felt like a more accurate measure of scale. We should capture the problems unique to `monorepos` vs `lots of repos`. Adding it to the `Problem` section as a specific group of problems we want to target solving seems to make sense. This would mean adding it as an item to the list Christina outlines here [comment](https://github.com/sourcegraph/about/pull/2001#discussion_r523891652).

handbook/engineering/search/goals.md
poojaj-tech: Not yet, you would need to toggle Git dir size + LOC in the linked look to get this info. As mentioned in the search sync there's ~10 customers we'd define as large right now. I'll either create a separate look or add a link to a google doc with the names in a follow-up.

handbook/engineering/search/goals.md
```diff
@@ -2,64 +2,52 @@
...
 
-**Success criteria**
+See [looker](https://sourcegraph.looker.com/looks/436) for current customer data.
 
-TODO table of use cases / capabilities we want to achieve, comparing Sourcegraph with other search tools
+#### Problem
```
lguychard, line 18:
```suggestion
#### [WIP] Problem
```

handbook/engineering/search/goals.md
```diff
@@ -2,64 +2,52 @@
...
 
-**Planned work**
+Large customers expect search to be fast at their scale. Customers expect to be able to search over their codebase intuitively, and want their first search result quickly. For larger organizations, customers care about getting all search results for important changes in their codebase. Supporting search at this scale also helps work towards supporting the Cloud growth efforts.
 
-1. Search expressions
-1. Search language rules engine
+#### Outcome
```
lguychard, line 22:
```suggestion
#### [WIP] Outcome
```

handbook/engineering/search/goals.md
```diff
@@ -2,64 +2,52 @@
...
+- Make it clear to the passing naive hacker that they couldn’t write something which replaces us in a few weekends / just use grep.
+- Our search experience is consistent, reliable, and predictable: IE Sourcegraph search does what it says and says what it does.
+- We understand more about each individual user and tailor their results so that results are ranked to promote the languages and repositories they are most likely to care about.
+- Search queries and their results become a frictionless starting point for creating all the other business objects our users will start to value… monitors, insights, campaigns and /next/.
+- Search is optimized for the cloud, where hundreds of a company’s private repos will live alongside hundreds of thousands of public open source repos.
+- We have a learnable, comprehensive code search query language; performant and reliable search across all result types we offer; a polished product that handles failure cases properly; and a personalized experience for users that surfaces their most relevant code.
 
-**Success criteria**
+#### Milestones (WIP: currently unordered and incomplete)
```
lguychard, line 33:
```suggestion
#### [WIP] Milestones
```

PR #2002, opened 2020-11-13
Update frontend engineer job spec

jeanduplessis:
Refines the role's requirements and updates the interview process

### DIFF
--- a/handbook/engineering/hiring/software-engineer-frontend.md
+++ b/handbook/engineering/hiring/software-engineer-frontend.md
@@ -41,8 +41,8 @@ The following engineering teams are currently hiring for this role:
 - Large and/or numerous contributions to open source projects.
 - Published blog posts and/or tech talks about your work.
 - Experience working on small high-performing product teams, preferably tech startups.
-- Experience working with a global or otherwise multicultural team
-- Developer platform/tool industry experience
+- Experience working with a global or otherwise multicultural team.
+- Developer platform/tool industry experience.
 
 ## Learn more about us
 
--- a/handbook/engineering/hiring/software-engineer-frontend.md
+++ b/handbook/engineering/hiring/software-engineer-frontend.md
@@ -10,7 +10,7 @@ The following engineering teams are currently hiring for this role:
 ## About you
 
 - You have practice at creating high quality software balanced with a pragmatic understanding of how to make appropriate tradeoffs (e.g. cut scope) to ship quickly and iterate when necessary.
-- You are proficiency in the English language, both written and verbal, sufficient for success in a remote and largely asynchronous work environment.
+- You are proficient in the English language, both written and verbal, sufficient for success in a remote and largely asynchronous work environment.
 - You communicate clearly and empathetically, especially in writing and documentation.
 - You share our [values](../../../company/values.md), and work in accordance with those values
 - You have the ability to thrive in an all-remote and largely asynchronous organization
--- a/handbook/engineering/hiring/software-engineer-frontend.md
+++ b/handbook/engineering/hiring/software-engineer-frontend.md
@@ -59,7 +59,7 @@ Learn more about what it is like to work at Sourcegraph by reading [our handbook
    - Your past work experience. In the absence of professional work experience we would look at any open-source projects you have created or contribute to.
    - The [Nice-to-haves](#nice-to-haves) we listed above.
 1. A **Recruiter** sets up a 30 minute call to learn more about what you are looking for, tell you about Sourcegraph, and answer any questions that you have.
-1. We set up a 1 hour interview with the **Hiring Manager**, who will tell you more about the role in the team. They will ask you about your past work and accomplishments and assess your sutiability for the role.
+1. We set up a 1 hour interview with the **Hiring Manager**, who will tell you more about the role in the team. They will ask you about your past work and accomplishments and assess your suitability for the role.
 1. You complete a 3 hour [coding excercise](software-engineer-coding-exercise.md#frontend-coding-exercise) that we designed to test your technical ability and the responsibilities of the role as [listed above](#your-responsibilities).
    - _Note: while we do not disqualify candidates who do not have React and TypeScript experience from taking the excercise, you will be required to use these for the excercise._
 1. An **Engineer** and a **Designer** will review and grade your submission.
--- a/handbook/engineering/hiring/software-engineer-frontend.md
+++ b/handbook/engineering/hiring/software-engineer-frontend.md
@@ -2,15 +2,22 @@
 
 We are looking for frontend engineers who know how to build intuitive user experiences to make the power of [Universal Code Search](https://about.sourcegraph.com/product) accessible to everyone. You will have a lot of ownership to solve tough technical and UX problems in key areas of our web application and browser extensions (e.g. code search, code navigation, campaigns).
 
+The following engineering teams are currently hiring for this role:
+
+1. [Web](../web/index.md#growth-plan) - all experience levels
+2. [Cloud](../cloud/index.md#growth-plan) - senior experience level
+
 ## About you
 
 - You have practice at creating high quality software balanced with a pragmatic understanding of how to make appropriate tradeoffs (e.g. cut scope) to ship quickly and iterate when necessary.
+- You are proficiency in the English language, both written and verbal, sufficient for success in a remote and largely asynchronous work environment.
 - You communicate clearly and empathetically, especially in writing and documentation.
-- You are [high agency](../../../company/values.md#high-agency).
+- You share our [values](../../../company/values.md), and work in accordance with those values
+- You have the ability to thrive in a fully remote and largely asynchronous organization
 
 ## Your responsibilities
 
-- Work closely with a dedicated designer and PM to implement high-quality UIs.
+- Work closely with a dedicated designer and PM to implement high-quality UIs using React, SCSS and TypeScript.
 - Actively participate in the planning of the features we're going to build, e.g. by writing and reviewing RFCs.
 - Build UIs you will love to use yourself as a developer.
 - Ensure UI/UX consistency in our app.
@@ -33,7 +40,9 @@ We are looking for frontend engineers who know how to build intuitive user exper
 - Experience building browser extensions (e.g. Chrome, Firefox, Safari).
 - Large and/or numerous contributions to open source projects.
 - Published blog posts and/or tech talks about your work.
-- Experience working on small high-performing teams, preferably tech startups.
+- Experience working on small high-performing product teams, preferably tech startups.
+- Experience working with a global or otherwise multicultural team
+- Developer platform/tool industry experience
 
 ## Learn more about us
 
@@ -46,10 +55,17 @@ Learn more about what it is like to work at Sourcegraph by reading [our handbook
 ## Interview process
 
 1. You [apply here](https://jobs.lever.co/sourcegraph/b2f9a8b0-cc06-4629-81a0-0f2fa64271c7/apply).
+1. We screen your application for suitability for the role by looking at:
+   1. Your past work experience. In the absence of professional work experience we would look at any open-source projects you have created or contribute to.
+   1. Whether you have the necessary technical skills that we require.
+   1. The answers you provide to the application questions
 1. We set up a 30-minute call to learn more about what you are looking for, tell you about Sourcegraph, and answer any questions that you have.
-1. You complete a 3-hour [exercise project](software-engineer-coding-exercise.md#frontend-coding-exercise) that we designed to test the technical qualifications listed above.
-1. We review your submission and schedule a **1 hour follow-up call** to discuss it.
-1. We schedule additional interviews across multiple days.
+1. You complete a 3-hour [coding excercise](software-engineer-coding-exercise.md#frontend-coding-exercise) that we designed to test the technical qualifications and responsibilities of the role as [listed above](#your-responsibilities).
+   1. _Note: while we do not disqualify candidates who do not have React and TypeScript experience from taking the excercise, you will be required to use these for the excercise._
+1. We review your submission and schedule a **1 hour follow-up call** with an engineer and designer to discuss it.
+   1. _Note: at this point, based on the evaluation of the coding excercise assessment, we might choose to allocate you to a different team than you might have indicated a preference for during your application if we believe your skills would be more suitable in it._
+1. We set up a 1-hour interview with the **Hiring Manager** who will tell you more about the role in the team. They will ask you about your past work and accomplishments and assess your sutiability for the role.
+1. If, after the coding excercise evaluation and hiring manager interviews, we believe you to be a suitable candidate for the role, we schedule the following additional interviews across multiple days:
    - 1h **Technical experience:** We ask you about your past work and accomplishments in depth, how you worked with others, decisions you made, and what you'd do differently today.
    - 1h **Team collaboration:** We ask you about how you work and communicate in a team setting, and how you handle tricky situations.
    - 30m **VP Engineering**
--- a/handbook/engineering/hiring/software-engineer-frontend.md
+++ b/handbook/engineering/hiring/software-engineer-frontend.md
@@ -61,20 +61,15 @@ Learn more about what it is like to work at Sourcegraph by reading [our handbook
 1. An **Engineer** and a **Designer** will review and grade your submission.
 1. We schedule a **1 hour follow-up call** to discuss your submission and ask any questions we might have to clarify our understanding of your technical capability and sutiability for the role.
 1. We schedule the following additional interviews, in no particular order, across multiple days:
-   - 1 hour **Technical experience:** We ask you about your past work and accomplishments in depth, how you worked with others, decisions you made, and what you'd do differently today.
-   - 1 hour **Team collaboration:** We ask you about how you work and communicate in a team setting, and how you handle tricky situations.
+   - 1 hour **Team collaboration:** We ask you about how you work and communicate in a team setting, how you handle tricky situations, decisions you made, and what you'd do differently today.
    - 30 minutes **VP Engineering**
    - 30 minutes **CTO**
    - 30 minutes **CEO**
 1. We check your references.
 1. We make you a job offer.
 
-**Please note the following:**
-- Your application could get rejected after any one of the steps in our interview process.
-- We might choose to assign you to a different team than you might have indicated a preference for during your application if we believe your skills would be more suited for it.
-
 We want to ensure Sourcegraph is an environment that suits your working style and empowers you to do your best work, so we are eager to answer any questions that you have about us at any point in the interview process.
 
 **[Click here to apply](https://jobs.lever.co/sourcegraph/b2f9a8b0-cc06-4629-81a0-0f2fa64271c7/apply)**
 
-Go back to the [careers page](../../../company/careers.md) for all open positions.
+Go back to the [careers page](../../../company/careers.md) for all open positions or read more about our [hiring processes](../../people-ops/hiring).
--- a/handbook/engineering/hiring/software-engineer-frontend.md
+++ b/handbook/engineering/hiring/software-engineer-frontend.md
@@ -11,7 +11,7 @@ The following engineering teams are currently hiring for this role:
 
 - You have practice at creating high quality software balanced with a pragmatic understanding of how to make appropriate tradeoffs (e.g. cut scope) to ship quickly and iterate when necessary.
 - You communicate clearly and empathetically, especially in writing and documentation.
-- You share our [values](../../../company/values.md), and work in accordance with those values
+- You share our [values](../../../company/values.md), and work in accordance with those values.
 
 ## Your responsibilities
 
--- a/handbook/engineering/hiring/software-engineer-frontend.md
+++ b/handbook/engineering/hiring/software-engineer-frontend.md
@@ -53,11 +53,7 @@ Learn more about what it is like to work at Sourcegraph by reading [our handbook
 ## Interview process
 
 1. You [apply here](https://jobs.lever.co/sourcegraph/b2f9a8b0-cc06-4629-81a0-0f2fa64271c7/apply).
-1. The **Hiring Manager** screens your application for suitability for the role by looking at:
-   - The answers you provide to the application questions.
-   - The skills you have obtained in your career.
-   - Your past work experience. In the absence of professional work experience we would look at any open-source projects you have created or contribute to.
-   - The [Nice-to-haves](#nice-to-haves) we listed above.
+1. The **Hiring Manager** reviews all the information you provided on your application to determine if you meet our qualifications for this role (if there is another open role we think you would be better qualified for, we will let you know).
 1. A **Recruiter** sets up a 30 minute call to learn more about what you are looking for, tell you about Sourcegraph, and answer any questions that you have.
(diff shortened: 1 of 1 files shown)


handbook/engineering/hiring/software-engineer-frontend.md
```diff
@@ -2,15 +2,22 @@
 
 We are looking for frontend engineers who know how to build intuitive user experiences to make the power of [Universal Code Search](https://about.sourcegraph.com/product) accessible to everyone. You will have a lot of ownership to solve tough technical and UX problems in key areas of our web application and browser extensions (e.g. code search, code navigation, campaigns).
 
+The following engineering teams are currently hiring for this role:
```
jeanduplessis, line 5: @tsenart FYI as discussed. Also, note the Hiring Manager interview step in the interview process and let me know if you are happy with this.

handbook/engineering/hiring/software-engineer-frontend.md
```diff
@@ -2,15 +2,22 @@
...
+- You are proficiency in the English language, both written and verbal, sufficient for success in a remote and largely asynchronous work environment.
 - You communicate clearly and empathetically, especially in writing and documentation.
-- You are [high agency](../../../company/values.md#high-agency).
+- You share our [values](../../../company/values.md), and work in accordance with those values
+- You have the ability to thrive in an all-remote and largely asynchronous organization
 
 ## Your responsibilities
 
-- Work closely with a dedicated designer and PM to implement high-quality UIs.
+- Work closely with a dedicated designer and PM to implement high-quality UIs using React, SCSS and TypeScript.
```
jeanduplessis, line 20: @felixfbecker I call this out specifically here as I make mention of it as a requirement for the coding excercies. Keen to hear your thoughts on this and whether we should expand it with the framework we use for testing for instance.

handbook/engineering/hiring/software-engineer-frontend.md
```diff
@@ -46,10 +55,18 @@ Learn more about what it is like to work at Sourcegraph by reading [our handbook
...
+   1. Whether you have the necessary technical skills that we require.
+   1. The answers you provide to the application questions.
+   1. And the [Nice-to-haves](#nice-to-haves) we listed above.
 1. We set up a 30-minute call to learn more about what you are looking for, tell you about Sourcegraph, and answer any questions that you have.
-1. You complete a 3-hour [exercise project](software-engineer-coding-exercise.md#frontend-coding-exercise) that we designed to test the technical qualifications listed above.
-1. We review your submission and schedule a **1 hour follow-up call** to discuss it.
-1. We schedule additional interviews across multiple days.
+1. You complete a 3-hour [coding excercise](software-engineer-coding-exercise.md#frontend-coding-exercise) that we designed to test the technical qualifications and responsibilities of the role as [listed above](#your-responsibilities).
+   1. _Note: while we do not disqualify candidates who do not have React and TypeScript experience from taking the excercise, you will be required to use these for the excercise._
```
jeanduplessis, line 65: @felixfbecker this is the change to the coding exercise that makes it explicit that you should use our tech stack to complete the coding excercise.

handbook/engineering/hiring/software-engineer-frontend.md
```diff
@@ -46,10 +55,18 @@ Learn more about what it is like to work at Sourcegraph by reading [our handbook
...
+1. We review your submission and schedule a **1 hour follow-up call** with an engineer and designer to discuss it.
```
jeanduplessis, line 66: @rrhyne @AlicjaSuska @quinnkeast FYI

handbook/engineering/hiring/software-engineer-frontend.md
```diff
@@ -46,10 +55,18 @@ Learn more about what it is like to work at Sourcegraph by reading [our handbook
...
+   1. _Note: at this point, based on the evaluation of the coding excercise assessment, we might choose to allocate you to a different team than you might have indicated a preference for during your application if we believe your skills would be more suitable in it._
+1. We set up a 1-hour interview with the **Hiring Manager** who will tell you more about the role in the team. They will ask you about your past work and accomplishments and assess your sutiability for the role.
```
jeanduplessis, line 68: @withdavidli this is a new step in the interview process for the Frontend Engineer role.

@nicksnyder It's my recommendation that if a candidate does not receive a strong hire from either the hiring manager or the engineer & designer that reviewed their coding exercise that we do not proceed with the rest of the interviews. This will ensure we have a high bar for joining the team and protect the time of other teammates so they only interview candidates that are very likely to result in a positive outcome.

handbook/engineering/hiring/software-engineer-frontend.md
tsenart: LGTM, thanks!

handbook/engineering/hiring/software-engineer-frontend.md
felixfbecker, line 66: I think we should call out that a designer will review it too (not just be in the call). It'll make clear that visual fidelity, UX etc will be tested.
```suggestion
1. An engineer and a designer will review your submission and schedule a **1 hour follow-up call** to discuss it.
```

felixfbecker, line 66: Should we separate the call and the review? If we rate no hire on the submission, we don't do a call - this wording currently implies we always do the call, which may cause a candidate to not write all context into the README thinking they can discuss it in the call, which works against their favor.

felixfbecker (commented):
Good evolution, I like how we're more upfront about more things.

handbook/engineering/hiring/software-engineer-frontend.md
```diff
@@ -2,15 +2,22 @@
...
 
+The following engineering teams are currently hiring for this role:
+
+1. [Web](../web/index.md#growth-plan) - all experience levels
+2. [Cloud](../cloud/index.md#growth-plan) - senior experience level
+
 ## About you
 
 - You have practice at creating high quality software balanced with a pragmatic understanding of how to make appropriate tradeoffs (e.g. cut scope) to ship quickly and iterate when necessary.
+- You are proficiency in the English language, both written and verbal, sufficient for success in a remote and largely asynchronous work environment.
```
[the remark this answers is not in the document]
nicksnyder: I think we can definitely delete this line, and can potentially delete others as well.

```suggestion
- You are proficiency in the English language, both written and verbal, sufficient for success in a remote and largely asynchronous work environment.
```

From our handbook:

https://about.sourcegraph.com/handbook/people-ops/hiring/job_description_guidelines#qualifications

> Avoid general skills that is expected of everyone we hire, or is expected for everyone in that role. For example, communication skills, being collaborative, able to write clean and reusable code for engineering roles, or writing skills for writer roles.

handbook/engineering/hiring/software-engineer-frontend.md
nicksnyder: > It's my recommendation that if a candidate does not receive a strong hire from either the hiring manager or the engineer & designer that reviewed their coding exercise that we do not proceed with the rest of the interviews. This will ensure we have a high bar for joining the team and protect the time of other teammates so they only interview candidates that are very likely to result in a positive outcome.


100% agree with this sentiment, but I have questions on implementation. For example, I don't think it is a good outcome to simply add another 1hr interview to our process.

Example alternatives that I would be more comfortable with:
1. Don't add this step, but instead make the first HM call this call (make it 1 hour and do a heavy screen). You can always cut the call short if it isn't going well. We are in a strong hiring position, we have a public handbook, so our first call with a candidate doesn't have to be pitching the candidate as much, we can go in and immediately start asking tough questions. Set candidate expectations in the email template when we reach out to them to schedule this call (e.g. we expect you have already read the handbook).
1. If we add a second HM step, it needs to be a structured interview just like other interviews (documented questions and expectations). To a first approximation it could be as simple as "HM does technical experience interview" so this would be net-neutral change to candidates. In any case, if the HM is going to do another 1 hr interview, we need to ensure it doesn't duplicate signal from subsequent interviews.

felixfbecker: @nicksnyder I'm confused why you say "second HM step". This is the first and only time the HM gets to meet the candidate in the process afaict?

nicksnyder: Oh, it wasn't clear from the JD that a recruiter does the first 30m call (step 3). If that is true, then it should be explicit.

> We set up a 30-minute call to learn more about what you are looking for, tell you about Sourcegraph, and answer any questions that you have.

If this is a recruiter call, then I think the HM should speak with the candidate BEFORE the coding exercise.

handbook/engineering/hiring/software-engineer-frontend.md
jeanduplessis: The thought is that while we don't require you to have professional React and TypeScript experience, the reality is that we exclusively use that in our frontend and we want to assess candidates on their ability to utilize these to create a frontend solution. It places an expectation on the candidate that to pass our coding exercise you need to show us that you can make use of these technologies to deliver your solution. This also ensures we can calibrate our reviews of candidates better since we are not comparing different tech stacks.

handbook/engineering/hiring/software-engineer-frontend.md
jeanduplessis: I've clarified the roles and moved the HM call to be before the coding exercise.

handbook/engineering/hiring/software-engineer-frontend.md
```diff
@@ -46,10 +55,18 @@ Learn more about what it is like to work at Sourcegraph by reading [our handbook
...
+   1. _Note: while we do not disqualify candidates who do not have React and TypeScript experience from taking the excercise, you will be required to use these for the excercise._
+1. We review your submission and schedule a **1 hour follow-up call** with an engineer and designer to discuss it.
+   1. _Note: at this point, based on the evaluation of the coding excercise assessment, we might choose to allocate you to a different team than you might have indicated a preference for during your application if we believe your skills would be more suitable in it._
+1. We set up a 1-hour interview with the **Hiring Manager** who will tell you more about the role in the team. They will ask you about your past work and accomplishments and assess your sutiability for the role.
+1. If, after the coding excercise evaluation and hiring manager interviews, we believe you to be a suitable candidate for the role, we schedule the following additional interviews across multiple days:
```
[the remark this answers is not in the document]
jeanduplessis: I've updated the text to indicate that we can reject a candidate at any point in our interview process.

jeanduplessis:
@felixfbecker @nicksnyder I've updated the text based on the feedback received so far, please review again. 
The specific commit that deals with the feedback implementation can be seen here: [`ac561e9` (#2002)](https://github.com/sourcegraph/about/pull/2002/commits/ac561e9f893a6f3b33f9c81395c86ea280603053)

handbook/engineering/hiring/software-engineer-frontend.md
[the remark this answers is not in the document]
jeanduplessis: @nicksnyder Thanks for pointing out the handbook guidance on this. I'm not 100% sure I agree with it completely yet, but that's a discussion for a different PR. To move this PR forward I've removed the reference to English language proficiency.

rrhyne (approved):

handbook/engineering/hiring/software-engineer-frontend.md
```diff
@@ -2,15 +2,20 @@
...
+
+1. [Web](../web/index.md#growth-plan) - all experience levels
+2. [Cloud](../cloud/index.md#growth-plan) - senior experience level
+
 ## About you
 
 - You have practice at creating high quality software balanced with a pragmatic understanding of how to make appropriate tradeoffs (e.g. cut scope) to ship quickly and iterate when necessary.
 - You communicate clearly and empathetically, especially in writing and documentation.
-- You are [high agency](../../../company/values.md#high-agency).
+- You share our [values](../../../company/values.md), and work in accordance with those values
```
felixfbecker, line 14:
```suggestion
- You share our [values](../../../company/values.md), and work in accordance with those values.
```

felixfbecker (approved):
LGTM!

handbook/engineering/hiring/software-engineer-frontend.md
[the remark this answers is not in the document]
felixfbecker: @nicksnyder note that the link you shared is for the "Qualifications" section, but this is the "About you" section, which I had added to communicate to the candidate (not our recruiters) what we expect of them. I think it's very important to be explicit here, even repititive if necessary, because this is what candidates will read (not the linked handbook page). I do want to be able to be explicit here that we expect candidates to e.g. write "clean and reusable code" (like your linked section says to avoid in "Qualifications") or that they need to be able to communicate well in English as Jean proposed, because we actually get a lot of applications from candidates that do not meet those (seemingly obvious) criteria.

handbook/engineering/hiring/software-engineer-frontend.md
```diff
@@ -46,18 +53,30 @@ Learn more about what it is like to work at Sourcegraph by reading [our handbook
...
-   - 30m **VP Engineering**
-   - 30m **CTO**
-   - 30m **CEO**
+1. The **Hiring Manager** screens your application for suitability for the role by looking at:
+   - The answers you provide to the application questions.
+   - The skills you have obtained in your career.
+   - Your past work experience. In the absence of professional work experience we would look at any open-source projects you have created or contribute to.
+   - The [Nice-to-haves](#nice-to-haves) we listed above.
+1. A **Recruiter** sets up a 30 minute call to learn more about what you are looking for, tell you about Sourcegraph, and answer any questions that you have.
+1. We set up a 1 hour interview with the **Hiring Manager**, who will tell you more about the role in the team. They will ask you about your past work and accomplishments and assess your sutiability for the role.
```
nicksnyder, line 62:
```suggestion
1. We set up a 1 hour interview with the **Hiring Manager**, who will tell you more about the role in the team. They will ask you about your past work and accomplishments and assess your suitability for the role.
```

handbook/engineering/hiring/software-engineer-frontend.md
```diff
@@ -46,18 +53,30 @@ Learn more about what it is like to work at Sourcegraph by reading [our handbook
...
+1. You complete a 3 hour [coding excercise](software-engineer-coding-exercise.md#frontend-coding-exercise) that we designed to test your technical ability and the responsibilities of the role as [listed above](#your-responsibilities).
+   - _Note: while we do not disqualify candidates who do not have React and TypeScript experience from taking the excercise, you will be required to use these for the excercise._
+1. An **Engineer** and a **Designer** will review and grade your submission.
+1. We schedule a **1 hour follow-up call** to discuss your submission and ask any questions we might have to clarify our understanding of your technical capability and sutiability for the role.
+1. We schedule the following additional interviews, in no particular order, across multiple days:
+   - 1 hour **Technical experience:** We ask you about your past work and accomplishments in depth, how you worked with others, decisions you made, and what you'd do differently today.
```
nicksnyder, line 68: Do we still need this or is this duplicative of the HM interview? I want to ensure that candidates aren't telling the same story over and over again.

handbook/engineering/hiring/software-engineer-frontend.md
```diff
@@ -46,18 +53,30 @@ Learn more about what it is like to work at Sourcegraph by reading [our handbook
...
-   - 1h **Technical experience:** We ask you about your past work and accomplishments in depth, how you worked with others, decisions you made, and what you'd do differently today.
-   - 1h **Team collaboration:** We ask you about how you work and communicate in a team setting, and how you handle tricky situations.
-   - 30m **VP Engineering**
-   - 30m **CTO**
-   - 30m **CEO**
+1. The **Hiring Manager** screens your application for suitability for the role by looking at:
+   - The answers you provide to the application questions.
+   - The skills you have obtained in your career.
+   - Your past work experience. In the absence of professional work experience we would look at any open-source projects you have created or contribute to.
+   - The [Nice-to-haves](#nice-to-haves) we listed above.
```
nicksnyder, line 60: nit (nonblocking): It feels like the JD itself is the definition of what we are looking for so it is redundant to elaborate here.

```suggestion
1. The **Hiring Manager** reviews all the information you provided on your application to determine if you meet our qualifications for this role (if there is another open role we think you would be better qualified for, we will let you know).
```

handbook/engineering/hiring/software-engineer-frontend.md
```diff
@@ -46,18 +53,30 @@ Learn more about what it is like to work at Sourcegraph by reading [our handbook
...
+   - 1 hour **Technical experience:** We ask you about your past work and accomplishments in depth, how you worked with others, decisions you made, and what you'd do differently today.
+   - 1 hour **Team collaboration:** We ask you about how you work and communicate in a team setting, and how you handle tricky situations.
+   - 30 minutes **VP Engineering**
+   - 30 minutes **CTO**
+   - 30 minutes **CEO**
 1. We check your references.
 1. We make you a job offer.
 
+**Please note the following:**
+- Your application could get rejected after any one of the steps in our interview process.
```
nicksnyder, line 77: This is documented in our handbook: https://about.sourcegraph.com/handbook/people-ops/hiring#respecting-time

Do you think there is a benefit to inlining this sentiment on our job descriptions? To me if feels like something that applies to every job, so doesn't really belong inline in a JD.

handbook/engineering/hiring/software-engineer-frontend.md
```diff
@@ -46,18 +53,30 @@ Learn more about what it is like to work at Sourcegraph by reading [our handbook
...
+- We might choose to assign you to a different team than you might have indicated a preference for during your application if we believe your skills would be more suited for it.
```
nicksnyder, line 78: I proposed an edit above to mention that we might reroute candidates doing application review.

nicksnyder (approved):

withdavidli (approved):

handbook/engineering/hiring/software-engineer-frontend.md
jeanduplessis: @nicksnyder I was wondering about this myself and agree it could lead to too much repetition. I'd say it's not needed based on the HM interview that proceeds it and other team members or frontend engineers can join the team collaboration interview rather.

handbook/engineering/hiring/software-engineer-frontend.md
jeanduplessis: Since we have it in the handbook I'm happy to remove it from here.

handbook/engineering/hiring/software-engineer-frontend.md
felixfbecker: I'd be worried about removing this interview. It has a great documented script, has a trained interviewer pool of engineers, is very well calibrated and the only other somewhat technical assessment other than the coding exercise (which is also still new and not well calibrated). I would feel less confident in making offers to candidates if this interview didn't exist.

jeanduplessis: @felixfbecker I don't think the idea is to remove the value this interview brought to the overall decision making, but rather that it falls on the hiring manager to cover this in their interview with the candidate. 

I found myself repeating a lot of information during my interviews at Sourcegraph, and previous roles for that matter, and it's always bugged me that companies with many interview steps don't make sure the context from previous interviews are shared so that candidates don't have to repeat themselves.

felixfbecker: Imo the Team Collaboration interview is not equivalent to Technical Experience:
- [Team Collaboration](https://github.com/sourcegraph/interviews/blob/master/engineering/team-collaboration.md)
- [Technical Experience](https://github.com/sourcegraph/interviews/blob/master/engineering/technical-experience.md)

In Team Collaboration there is no test for communicating with a _technical_ audience, continuous learning, mentorship nor interest/passion. There is maybe some overlap in the communication part, but the important difference here was that one is communication with a technical audience and the other with a non-technical audience (and that's why the interview panels are engineers for Technical Experience and non-engineers for Team Collaboration).

It _may_ be the case that a situation a candidate went through is a suitable example for a question in either interview but they would still be talking about separate aspects of that situation and ideally they can come up with separate, targeted examples (i.e. it's on them if they only have a single story and it would probably not work in their favor, unless that story is faceted enough to serve as an example for all the different questions).

If you care strongly about removing this and want to keep this change, @jeanduplessis what change do you propose to the Team Collaboration interview panel and what is the script for the HM call (so we make sure we're not removing any of these assessments from the overall interview process)?

felixfbecker: One side-effect of this would also be that the candidate would meet 2 less people of the team in the interview process (e.g. only one non-technical team member).

felixfbecker: > I found myself repeating a lot of information during my interviews at Sourcegraph

This is a valuable data point and I'd be curious to here more about this from both your and interviewer's perspectives (as I wasn't part of your interview panel).

> it's always bugged me that companies with many interview steps don't make sure the context from previous interviews are shared so that candidates don't have to repeat themselves.

As an interviewer at Sourcegraph, I have also previously wished we would share more context between interviews so we could know what to dig in more on, but what I heard was that it's intentional that we don't get to see other's feedback until the debrief to not introduce bias and that interview modules should be stand-alone and be able to each independently test a subset of the overall requirements. Curious if you agree with that?

jeanduplessis: > One side-effect of this would also be that the candidate would meet 2 less people of the team in the interview process (e.g. only one non-technical team member).

I would imagine we could incorporate the technical team members as part of the team collaboration session?

> This is a valuable data point and I'd be curious to here more about this from both your and interviewer's perspectives (as I wasn't part of your interview panel).

As the candidate, I found I had to share my work background and experience quite a bit. I can't exactly recall if it was any specific question that triggered that, but I do recall it happening frequently enough that I made a mental note of wanting to address it.

In terms of sharing context, I think what we want to avoid is sharing opinions about a candidate if there are other interviews still taking place. However, I see this as being different from a summary of a person's background and experience (i.e. more colorful version that their CV) being passed on as context for future interviewers. In fact, I would expect those interviewers to review the context and possibly ask follow-up questions of the candidate based on it. What I like to do as well is highlight areas I want another interviewer to dig into further, either because I am uncertain about something, or I didn't have time to get deeper into it.

nicksnyder: General principles:
- We need to ensure our hiring process captures all the signal that we care about.
- We need to ensure candidates aren't repeating themselves.

We generally accomplish both of these things by writing down the questions we ask in each interview so we can audit them to ensure nothing is duplicated.

In this diff we added a 1hr HM interview, and deleted the 1h technical experience interview. I think the current gap in the documentation is that we don't have guidance for the 1hr HM interview. Felix has concerns that we will no longer be getting important signal that we used to get from the technical experience interview.

@jeanduplessis Can you please open a new PR that either (1) documents the HM is doing the technical experience interview (to a first approximation) or (2) lists the questions that the HM is going to ask so we can audit it to ensure we get the right signal.


****


The second point is sharing candidate background across interviews. I agree this would be valuable but there are some challenges we would need to figure out.

1. We don't want to share the interviewer's judgement about the background, just the background itself.
2. When interviews are back-to-back, there is no time for interviewers to (A) write down the relevant info to share and (B) review feedback from previous interviews. (A) could be solved with transcription software like otter.ai. (B) could only be solved by scheduling gaps between interviews.

Another option is to move more of our interview process to "answer these written questions" (which anyone on the team can review) and then we have a live discussion.

PR #2006, opened 2020-11-13
Remove full stack role

nicksnyder:
## Situation

We recently created a full stack engineering role (https://about.sourcegraph.com/handbook/engineering/hiring/software-engineer-full-stack) to fill a perceived need that we had on multiple teams. That need was “Code intel, search, cloud teams have mostly backend work, but still have frontend needs, but not enough to warrant a full time frontend dev”.

## Complication

Search successfully hired someone into this role and is now focused on backend hiring. Cloud successfully hired someone into this role, and also realized that they have a lot more frontend work in front of them such that they actually could still use another dedicated frontend engineer. Code intel needs to grow, but (1) we don’t have a hiring manager to own this pipeline and (2) the quality of candidates we got for code intel team through the full stack role wasn’t as good as the candidates that we got through the more targeting code intel JD (which we took down since we don’t have a HM).

## Question

What should we do with the full stack role?

## Answer

My proposal is that we should close the full stack role. Cloud and Search don’t need to hire for this role any more, which would only leave code intel. We aren’t going to make progress on the role until Owen (EM code intel) starts on the 23rd, and it is probably better for him to write a fresh JD anyway that is specific for what code intel team needs.


efritz (approved):

chrispine (approved):
👍 Campaigns will start looking for a full-stack engineer soon, around when Owen joins. I think it makes sense for us to work on that together and start fresh.

tsenart (approved):

unknwon:
Side-note:

@tsenart: was "tsenard" a typo of your GitHub handle (in the CODENOTIFY)?

lguychard (approved):

PR #2009, opened 2020-11-14
Explicitly state frontend code exercise requirements

jeanduplessis:
@felixfbecker how do you feel about the added language of making things a bit more explicit for the frontend exercise? By stating what we look at in the review of the solution it helps the candidate understand the expectation we have of them and allows them to understand better how to allocate their time and what to focus on.

### DIFF
--- a/handbook/engineering/hiring/software-engineer-coding-exercise.md
+++ b/handbook/engineering/hiring/software-engineer-coding-exercise.md
@@ -32,7 +32,7 @@ Things we consider when reviewing your solution:
 - Adherance to the design
 - Functional completeness
 - Semantically correct and accesible markup
-- Whether the structure of the components and code separate UI and logic concerns
+- Whether the code is well-structured
 - Test coverage
 
 If we decide to move forward, we schedule a one hour follow-up call to discuss your submission.
--- a/handbook/engineering/hiring/software-engineer-coding-exercise.md
+++ b/handbook/engineering/hiring/software-engineer-coding-exercise.md
@@ -23,10 +23,17 @@ If we decide to move forward, we will schedule a 30-minute followup call to disc
 We designed this exercise to measure your understanding of building web UIs.
 
 - You will choose a three hour timeframe to independently work on the exercise. You can choose any day and time; we just schedule an email.
-- The exercise asks you to implement a UI in a scaffold TypeScript React app.
+- The exercise asks you to implement a UI in a scaffolded TypeScript React app.
 - We provide you with the code of the scaffold app, the URL of a working GraphQL API to fetch data from and a link to a Figma design for the UI.
-- You are free in how you chose to implement it and can change the existing code in any way you'd like.
+- You are free in how you chose to implement your solution, and can change the existing code in any way you'd like, however be aware the scaffold app is there to save you time.
 - You can use your own development environment and lookup documentation on the internet.
+
+Things we consider when reviewing your solution:
+- Adherance to the design
+- Functional completeness
+- Semantically correct and accesible markup
+- Whether the structure of the components and code separate UI and logic concerns
+- Test coverage
 
 If we decide to move forward, we schedule a one hour follow-up call to discuss your submission.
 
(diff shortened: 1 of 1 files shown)


nicksnyder (approved):

handbook/engineering/hiring/software-engineer-coding-exercise.md
```diff
@@ -25,9 +25,16 @@ We designed this exercise to measure your understanding of building web UIs.
 - You will choose a three hour timeframe to independently work on the exercise. You can choose any day and time; we just schedule an email.
 - The exercise asks you to implement a UI in a scaffold TypeScript React app.
 - We provide you with the code of the scaffold app, the URL of a working GraphQL API to fetch data from and a link to a Figma design for the UI.
-- You are free in how you chose to implement it and can change the existing code in any way you'd like.
+- You are free in how you chose to implement your solution, and can change the existing code in any way you'd like, however be aware the scaffold app is there to save you time.
```
felixfbecker, line 28: This seems to contradict your other change to require React, because it says the scaffold can be changed in any you like (which would include changing it to not use React). Is that intentional? In any case we should realign the two documents to not be contradictory.

felixfbecker (commented):
I generally like being more explicit than we were before, but I wonder to what degree we're "giving away the answer in the task", or may even mislead. Would like to get @lguychard's thoughts too.
Somewhat related here's an example of clearer expectations I wrote without mentioning any "implementation details" (I had this change locally, not meant as a direct counter proposal): https://github.com/sourcegraph/commit-page-exercise/pull/4/files

handbook/engineering/hiring/software-engineer-coding-exercise.md
jeanduplessis: I've removed this statement as I don't think it adds any value to the candidate based on the current expectation we outline.

jeanduplessis:
> I generally like being more explicit than we were before, but I wonder to what degree we're "giving away the answer in the task", or may even mislead. Would like to get @lguychard's thoughts too.
> Somewhat related here's an example of clearer expectations I wrote without mentioning any "implementation details" (I had this change locally, not meant as a direct counter-proposal): [sourcegraph/commit-page-exercise#4 (files)](https://github.com/sourcegraph/commit-page-exercise/pull/4/files)

I like your wording, it speaks to the intent I had of helping candidates understand what to spend their time on and the expectation, however, you said it more eloquently.

In terms of "giving away the answer in the task": I always feel that it's better to be clear on your expectation of a candidate; set them up for success without giving the solution. When a new teammate joins Sourcegraph we don't expect them to guess our standards and expectations. They have our Practices & Philosophy as specified in the handbook and have an existing codebase to reference, so my thought is always why won't we share the same with a prospective candidate? Interviewing is stressful enough that I don't want to test how good someone is under pressure, I want to know whether they can deliver a "well-rounded" solution as you put it.

handbook/engineering/hiring/software-engineer-coding-exercise.md
felixfbecker: Sounds good. Maybe the only thing we would wanna keep communicate is "you can add dependencies" because I think that's a common question for coding exercises

handbook/engineering/hiring/software-engineer-coding-exercise.md
```diff
@@ -25,9 +25,13 @@ We designed this exercise to measure your understanding of building web UIs.
 - You will choose a three hour timeframe to independently work on the exercise. You can choose any day and time; we just schedule an email.
 - The exercise asks you to implement a UI in a scaffold TypeScript React app.
 - We provide you with the code of the scaffold app, the URL of a working GraphQL API to fetch data from and a link to a Figma design for the UI.
-- You are free in how you chose to implement it and can change the existing code in any way you'd like.
 - You can use your own development environment and lookup documentation on the internet.
```
felixfbecker, line 28:
```suggestion
- You can use your own development environment, lookup documentation on the internet and add dependencies.
```

felixfbecker (approved):

felixfbecker:
> In terms of "giving away the answer in the task": I always feel that it's better to be clear on your expectation of a candidate; set them up for success without giving the solution. When a new teammate joins Sourcegraph we don't expect them to guess our standards and expectations. They have our Practices & Philosophy as specified in the handbook and have an existing codebase to reference, so my thought is always why won't we share the same with a prospective candidate? Interviewing is stressful enough that I don't want to test how good someone is under pressure, I want to know whether they can deliver a "well-rounded" solution as you put it.

I agree with this and I actually want to make a change to encode that better, in particular adding a bit more "example code" to the scaffold as "a point of reference" as you put it. For example, I had already added Jest as a dependency, but there was no reference test, and I am convinced that the things I would look for are more in _what_ they test than in that they test at all.

PR #2024, opened 2020-11-17
Reword search large-scale organizations goal

lguychard:
As per https://github.com/sourcegraph/about/pull/2001/files#discussion_r523890984, suggests a rewrite to our goal that is more time-based: if we identify the "core use cases", we have reached our goal when large-scale organizations can effectively rely on them.


lguychard:
Merging as-is to iterate, happy to take post-merge suggestions.

PR #2030, opened 2020-11-17
Provide feedback on onboarding

nicksnyder:
We don't have a survey, so updating the document to have the new hire provide this feedback.


lguychard:
✅

PR #2037, opened 2020-11-18
Sync coding exercise added to interviews.

chayim:


nicksnyder (commented):
This whole process is now 7.5 hours, which feels like a lot.

On 2020-10-15 in our 1-1 notes we started sketching out what we were looking for and then how to measure it. How does this process map onto that thought process?

nicksnyder:
(I was somewhat assuming that this exercise would replace the vuln assessment)

chayim:
> (I was somewhat assuming that this exercise would replace the vuln assessment)

I view this as additive. Candidates contacted via an outbound reach out, have no measure of written, async communication. Security candidates in particular tend to be either technical or research oriented, but rarely both. I think this signal is important, to ensure our candidates have both skill sets.

> This whole process is now 7.5 hours, which feels like a lot.

I agree. Aligning on the coding exercise, means that we could merge the _Security concepts_ and _Technical experience_ interviews. At this point they should be the same, and he should have increased signal given the coding exercise. As you read this, those items will be merged.


> On 2020-10-15 in our 1-1 notes we started sketching out what we were looking for and then how to measure it. How does this process map onto that thought process?

I think this maps directly to what we were thinking. We now try to gather security specific signals measuring understanding and communication. Additionally, we're now qualifying a candidate's coding skills. The new process helps give us a 360 degree view of our candidate, from technical through soft skills.

ElizabethStirling (approved):

PR #2038, opened 2020-11-18
Temp fixing an issue with emphasis and underscores in links 

Joelkw:
It seems that if trying to emphasize text, the underscores in any url paths break our docsite handling of the underscore string (VSCode preview and Github's own comment fields seem to handle this, so it doesn't seem like a limitation of markdown in general). 

The temporary solution is just to wrap the emphasized strings in a way that the url underscores aren't enclosed. 

In other words, before this change, this line looked like: 
  1. **\_Maintaining and expanding [code host integrations](https://docs.sourcegraph.com/dev/background-information/web/code_host_integrations)\_**


felixfbecker:
I think you could also use `***` instead of `**_`, `**` and `_` are equivalent in markdown

PR #2041, opened 2020-11-18
no company meeting between Dec 23 and Jan 1

sqs:
It wouldn't make sense to hold company meeting on, for example, December 28. Also, state that most team members will take that time off.


unknwon (approved):
❤️

PR #2043, opened 2020-11-19
first format take of an EM-PM partnership responsibility page

Joelkw:
This is one formatting attempt for the [EM/PM matrix](https://docs.google.com/document/d/1qtF3wfIG00I3MIiOpDO0hffJl4CJFoC_0IcAjJEGo6M/edit?ts=5fa52a34#) we want in the handbook.

I cut the "stakeholders" out because I felt like they were obvious in nearly all cases, they were just a few roles just listed repeatedly, and don't want to create a culture of "someone needs to be listed somewhere in the handbook for you to feel comfortable contacting them, if you think they are relevant." We hire smart people and they likely know when to inform the VP of Product or VP of Eng.

-----

_**Update 2020-11-19 3:41pm PT:** People prefer the alternate chart format for the "partnership" definition of this page, but there's a discussion about what the purpose of the page should be. You can [probably skip to the discussion](https://github.com/sourcegraph/about/pull/2043#issuecomment-730705178)._

------

### What I like about this format

- It makes it really easy to understand the responsibilities of the role (see [this conversation](https://docs.google.com/document/d/1qtF3wfIG00I3MIiOpDO0hffJl4CJFoC_0IcAjJEGo6M/edit?disco=AAAAKqBAdhM))
- It makes it clear there are "accountable" and "responsible, but someone else is accountable" sections
- The formatting is easy to read. 

### What I don't like about this format

I'm not sure it gets at the "partnership" angle enough, and it repeats some sections in both the PM and EM area. I have an alternate proposal for a format (currently commented-out) at the end of the `product_manager_engineering_manager_responsibilities` page, where we keep the EM/PM section-level split but rather than have the "EMs help PMs with" sections, we just put all that in an "EMs help" column of a table that goes in the PM section, and vice versa. For example, in the "PM accountable" section you would have: 

#### Product marketing 
| PM expectations | EM expectations | 
| --- | --- | 
| - Speak to customers about upcoming features <br/> - Promote new features internally in the company <br/> - Communicate with marketing + sales around upcoming or recent feature releases <br/> - Reach out to customers when heavily-requested features are released <br/> - Make sure the CE + Sales team understands and is excited to sell new features | - Look out for instances of gaps in product trust and understanding <br/> - Encourage engineers to provide demos of newly delivered features 

And then you would have no "Product marketing" section repeated on the page again in the EM section. 

I like that this alternative gets more at the partnership. I don't like that this alternative means the role descriptions are now split across both sections (some in the owned section, some in the "help the other manager with" column of the other manager's section). *But* my opinion is that perhaps the purpose of this page is to show how EM+PM partner on a specific task rather than give separate role descriptions. I also don't love that without adding a bunch of html+css (which I will do if we want to go this way) the table format here will be a bit uglier and less readable.  

### Other questions

We separately have PM and EM role descriptions (in the other two modified pages in this PR). This project started as a way for @jeanduplessis and I to figure out how to best work together. I'm worried it is morphing in to "here is a PM role description" and "here is an EM role description" and I wonder, is that useful in this page as well as the descriptions we already have in the handbook on the dedicated role pages? 

I justify this with "yes, this is additionally useful because it is about the partnership," in which case I return to my above concern that this format is best for "role description" rather than "partnership description." I think the core question here, if we agree this page should be partnership focused, is just what others find to be most readable/clear: the current format, or the suggested alternate format with 1/2 as many sections that have little table/columns in each section.

### DIFF
--- a/handbook/engineering/roles.md
+++ b/handbook/engineering/roles.md
@@ -31,6 +31,7 @@ Build an exceptional team that achieves results.
 - Own the end-to-end hiring process for the team and grow the team according to the hiring plan.
 - Make [compensation](../people-ops/compensation.md) decisions for direct reports and candidate offers to ensure that everyone is appropriately compensated at all times.
 - Hold teammates accountable for fulfilling their responsibilities.
+- If your team has a product manager, uphold your end of the [PM – EM partnership responsibilities](../product/roles/product_manager_engineering_manager_responsibilities.md).
 
 ## VP Engineering
 
--- a/handbook/product/roles/product_manager.md
+++ b/handbook/product/roles/product_manager.md
@@ -18,6 +18,7 @@ You will **own an area of the product**\* and:
 - Maintain a roadmap, backlog, and curate and solicit feedback for the product area.
 - Communicate product ideas clearly by focusing on the problems that are being solved, the outcomes, and how we will measure success.
 - Be transparent in your internal and external communication. Sourcegraph is open source, so most discussions are public or in channels where you communicate directly with our customers.
+- Uphold your side of the [PM – EM partnership responsibilities](product_manager_engineering_manager_responsibilities.md).
 
 \* Product ownership areas (current teams): [Campaigns](../../engineering/campaigns/index.md), [Cloud](../../engineering/cloud/index.md), [Distribution](../../engineering/distribution/index.md), [Security](../../engineering/security/index.md)
 
--- a/handbook/product/roles/product_manager_engineering_manager_responsibilities.md
+++ b/handbook/product/roles/product_manager_engineering_manager_responsibilities.md
@@ -6,6 +6,7 @@ An effective partnership between the EM and PM is fundamental to the success of 
 **Accountable**: The accountability for the responsibility stops with this person. Being accountable doesn’t mean you have to do all the tasks yourself; however, you should ensure it gets done.
 
 **Expectations**: This outlines the expectation that the EM/PM agrees to *proactively* fulfill in relation to the responsibility in the team. The EM/PM is encouraged to delegate and/or share these tasks with other teammates for the following reasons:
+
 - Share the load
 - Encourage collective ownership of the success of the team
 - Reduce the bus factor risk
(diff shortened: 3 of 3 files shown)


macraig:
This is awesome @Joelkw thanks for putting it together! 

> But my opinion is that perhaps the purpose of this page is to show how EM+PM partner on a specific task rather than give separate role descriptions.

I also like the alternate table proposal better, it gives it a more collaborative angle than just listing sections in separate roles. Having a side-by-side view of what each person owns/helps with in a particular section feels more like a team effort, almost like saying "This is the task to tackle and this is how we pair up on it as a team".

felixfbecker:
I would definitely prefer a table even just for the visual/readability aspect.

Joelkw:
From @christinaforney 's [slack](https://sourcegraph.slack.com/archives/CMT39K56Z/p1605825659429200?thread_ts=1605285751.315400&cid=CMT39K56Z): "I would like to figure out how these expectations are part of our actual role/job descriptions and are what we use to evaluate candidates. Because of this, I think Option 1 is better, because they can be split into two different pages for each roles (and then we could integrate them somehow similar to how we've done teams, goals, and roadmaps).

I do really like the property of having the expectations side by side, so if anyone has a good solution to wanting these requirements to be most tightly coupled in the role descriptions while also having the aligned location, I'd love to hear it!"

My reply: I agree there's a tension in those needs, but I'm not sure those are the right "needs" for this page. 

I think there are things in the individual [PM role description](https://about.sourcegraph.com/handbook/product/roles/product_manager) that would not belong on a page about the EM/PM partnership. 

For example, "Significantly grow usage and engagement of that area and Sourcegraph as a whole" is a PM responsibility, but the EM responsibility to help achieve the goals is the same regardless of what we direct the PM goals to focus on. If we start to input things that *define how* an individual's task should be done instead of *list* a task (for example, this is "set goals that grow usage" rather than "PM is accountable: set goals"), this is when it crosses over from what we (@jeanduplessis and I) originally were working on – the partnership and task split – into the territory of "very detailed job description." 

Another way of framing this, maybe, is "**is the PM–EM collaboration a process, or the job?**" 

To me, it feels like a process (similar to CE to Engineering handoff page), even though it's one that comprises much, but not all, of the job. So I propose that this should just be linked to from a job description as something like "Work with the engineering manager" (and that's how I've got it in this PR at the moment). 

If we decide instead that what we want is very detailed job descriptions, detailed such that defining this collaboration here is unnecessary, then I would propose separating this further, not keeping it on one page, and just putting the "accountable" and "responsible" sections for each role on the role pages that already exist.

jeanduplessis:
@Joelkw @christinaforney @nicksnyder: Here's my take on the intended use and therefore, the most appropriate formatting of this information.

1. This information is not intended to be a detailed job specification for either role. It's meant to be an acknowledgement between an EM and PM that the success of their product team relies on both of them playing vital parts in the partnership to ensure success and to understand their respective roles, the expectation of the one on the other, and to be able to hold each other to account.
1. I see the intended audience of this information to be:
   1. The EM and PM
   1. The cross-functional teammates of the team (engineers, designers, etc.) to understand who plays what role in the team and how to support them.
   1. Peers in other departments who have a dependency on/request for the team in understanding who is best to reach out to
1. I don't think it's appropriate to add all this info to a job spec. Still, I do agree we can and probably should add a reference to it from the job specs to allow candidates and teammates to understand what the successful partnership between an EM & PM might entail.
1. I also wouldn't like to see this as some rulebook for how each EM & PM need to work together. Rather I would frame it as a guideline, and encourage each EM & PM pairing to discuss it and come up with an understanding of how they will work together best. They should add to it, or take away from it as might be relevant for their product team and to complement the strengths that each one brings to the partnership. 
1. Another reason I don't think this should be considered as job spec level information is that you might have teammates who are not the EM or PM who might want to fulfil some of the tasks identified. This could be because they have a passion for that or they might want to develop in a specific area, and we should encourage teams to work together in whatever way leads to the best outcome for them.

So is this useful to have at a higher non-team specific level? Yes. It is a guideline that new EM & PM pairings can use to start a discussion and come to an understanding of how their team will best operate.

Given all this, I think the table layout works best.

`</2cents>`

nicksnyder:
I defer to @jeanduplessis and @Joelkw on what should ultimately land in our handbook. Optimize for solving the problem you have.

If Christina and I have a problem that is unsolved by what you add, we can propose subsequent changes.

Joelkw:
Per the above discussion, went with a table format because it served the purpose of this page better. Didn't block on content review since this was heavily reviewed in google docs form before copying over.

PR #2048, opened 2020-11-20
release process tweaks

bobheadxi:
Move the `release:close` step to right after we merge PRs, and note that this step generates tags


bobheadxi:
With https://github.com/sourcegraph/sourcegraph/pull/16033 , this isn't really necessary anymore

PR #2050, opened 2020-11-20
Update web tech stack

felixfbecker:
Team members get to use some pretty exciting technologies on our team, but we are not really advertising it anywhere. This adds some of these in our tech stack section on the team page and removes Go.
This update is just intended to reflect reality, nothing aspirational.

### DIFF
--- a/handbook/engineering/web/index.md
+++ b/handbook/engineering/web/index.md
@@ -19,7 +19,15 @@ See [goals](goals.md)
 
 ## Tech stack
 
-TypeScript, React, RxJS, GraphQL, Go.
+We use a modern, flexible tech stack.
+Here are some of the technologies we use to deliver on our goals:
+
+- [TypeScript](https://www.typescriptlang.org/)
+- [React](https://reactjs.org/)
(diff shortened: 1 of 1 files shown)


jeanduplessis (approved):

tjkandala (approved):

nicksnyder (approved):
don't think it is worth mentioning GraphQL?

felixfbecker:
@nicksnyder good suggestion, thanks!

tjkandala:
If we mention libraries like RxJS or Bootstrap, could we also mention codeintellify? I think it's really cool, fundamental to our UI, and a library that we should advertise, but it is _ours_, so it's not something that people will see and think "I've always wanted to work with that!".

felixfbecker:
@tjkandala I think what should go in here are technologies we use (code with) on at least a weekly basis. I think that's not true for codeintellify, it's kind of there, but we rarely have to touch it, and it's only for a small part of the functionality we own. It should be technologies that are known to externals that could excite them when they read them (or let them skip us if they hate that technology). I think listing all of our open source code is not valuable on that front. As you said it wouldn't be something that makes people say "I've always wanted to work with that!".

PR #2053, opened 2020-11-20
3.22 released

aileenrose:
updated to both home and /get-started pages


aileenrose (commented):
Verified that I can copy script and install new version.

PR #2058, opened 2020-11-24
Added entry for Scott Campbell

SPC74:


aileenrose:
Scott's bio successfully added in https://github.com/sourcegraph/about/pull/2095

PR #2060, opened 2020-11-24
Add page for seeking and giving better feedback

quinnkeast:
We have a great open culture around seeking and giving feedback. But sometimes, the feedback cycles aren't as strong as they could be, because the environment isn't clearly structured. This sometimes leads to circular conversations, unclear decision ownership, and can make folks less likely to seek or give feedback.

I've created a simple starting point for how to seek and give better feedback, with the idea that it's a frameworkable skill we can all work at improving.

### DIFF
--- a/handbook/communication/seeking-and-giving-feedback.md
+++ b/handbook/communication/seeking-and-giving-feedback.md
@@ -1,0 +1,50 @@
+# Seeking and giving feedback
+
+Seeking and acting on feedback is crucial for making good decisions and delivering our best work. Good feedback is a product of an intentional environment with positive feedback loops.
+
+## Seeking feedback
+
+When seeking feedback at Sourcegraph, it's important to:
+
+- Specifically say what *kind* of feedback you're seeking. A thumbs-up, thumbs-down? A gut call on the better of two approaches? A detailed critique?
+- Ask *specific questions* that you want feedback about.
+- Provide *context*. Make sure the specific problem and goal is clear.
+- Seek feedback according to [project roles](#project-roles).
+- After receiving feedback, follow up. Recognize contributions and share how you will move forward.
+
+### Great examples of requesting feedback
+
+(To be added!)
+
+## Giving feedback
+
+When giving feedback, it's important to:
+
+- Make sure you understand what kind of feedback is sought. If it's not clear what kind of feedback is sought, clarify before giving feedback.
+- Focus feedback on the specific feedback sought. Other feedback may be valuable, but should be separated into a distinct conversation.
+- Provide feedback in terms of **why**, not **how** (unless specifically sought).
+- Clarify where the feedback is coming from: is it based on a heuristic or best practice? Is it from a previous decision? Is it rooted in personal opinion?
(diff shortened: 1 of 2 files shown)


macraig (approved):

handbook/communication/seeking-and-giving-feedback.md
```diff
@@ -0,0 +1,50 @@
...
+
+(To be added!)
+
+## Giving feedback
+
+When giving feedback, it's important to:
+
+- Make sure you understand what kind of feedback is sought. If it's not clear what kind of feedback is sought, clarify before giving feedback.
+- Focus feedback on the specific feedback sought. Other feedback may be valuable, but should be separated into a distinct conversation.
+- Provide feedback in terms of **why**, not **how** (unless specifically sought).
```
AlicjaSuska, line 25: I agree with this line in general but I'd like to make it clear here that it's ok to suggest a potential solution @quinnkeast. Maybe we can describe it:

'When providing feedback always start with **why**. You're welcomed to suggest a potential solution (the **how**) but you need to describe your reasoning first so that your proposition is easily understandable.'

handbook/communication/seeking-and-giving-feedback.md
```diff
@@ -0,0 +1,50 @@
...
+
+### Great examples of providing feedback
+
+(To be added!)
+
+## Project roles
+
+When a project has multiple stakeholders, it's useful to identify specific levels of participation through project roles. Each project should clearly state who is:
+
+1. **Responsible:** Those who do the work to get things done.
```
AlicjaSuska, line 39: Should we add those categories of stakeholders to the RFC templated @quinnkeast @christinaforney ?

AlicjaSuska (commented):
Great work @quinnkeast Thank you for adding this!

handbook/communication/seeking-and-giving-feedback.md
quinnkeast: Maybe something like:

```suggestion
- Always start with **why**, not **how**. Potential solutions can be useful to clarify your feedback, but aren't feedback in themselves.
```

handbook/communication/seeking-and-giving-feedback.md
quinnkeast: I _think_ we talked about this when RACI roles first came up, and we agreed that if people find them useful, they'll start being used. Personally, I think they're only useful when an RFC reaches a certain level of definition (e.g. defined project and team). Too early, and they get in the way of defining the problem and proposing a solution.

handbook/communication/seeking-and-giving-feedback.md
AlicjaSuska: Looks good, thank you @quinnkeast !

quinnkeast:
@Joelkw I'm merging this so it becomes available as a resource right away. I'd still love to hear any additional feedback you mentioned!

Joelkw:
Of course! Didn't have any blocking comments. I'll make a PR with one suggestion I had and tag you!

PR #2063, opened 2020-11-25
Updating Temporary "PM" for Search

CecilyBee:
@mercadon

### DIFF
--- a/handbook/engineering/search/index.md
+++ b/handbook/engineering/search/index.md
@@ -55,7 +55,7 @@ Iterations start **every other Monday**.
 
 ## Members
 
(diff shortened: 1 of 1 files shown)


CecilyBee:
@nicksnyder Please use latest commit w/ Joel (not María & Alicja) as temporary PM.

nicksnyder:
What is the context for this change?

lguychard (changes requested):
This change is purely hiring-related (who should be involved in interviewing PM candidates for search), but should not be reflected on the search team's handbook page as it could be misleading re:responsibilities on the team.

CecilyBee:
@nicksnyder @lguychard Loïc is correct in that I was looking to change the list for scheduling purposes. I will change the job description instead. Thanks both!

PR #2064, opened 2020-11-25
Update Temp "PM" for Cloud

CecilyBee:
@mercadon

### DIFF
--- a/handbook/engineering/cloud/index.md
+++ b/handbook/engineering/cloud/index.md
@@ -113,7 +113,7 @@ By default, members of the team will provide a brief update about the goals they
 
 ## Members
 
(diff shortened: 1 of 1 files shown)


tsenart (commented):
@christinaforney: Can you provide some context on this change please? :)

christinaforney:
@tsenart - I also don't have context.

@CecilyBee - can you clarify this change? This section is for who is operating as the PM for this team - this change is incorrect for that.

If this is supposed to be for who should be on the interview hiring panels, can you please add this to the job description instead?

christinaforney (changes requested):
This is the list of actual team members, not who should be on the interview panels.

PR #2070, opened 2020-11-27
Document Gerrit deployment

pecigonzalo:
We deployed a dogfood Gerrit instance in https://github.com/sourcegraph/infrastructure/pull/2264.

/cc @sourcegraph/web

### DIFF
--- a/handbook/engineering/distribution/internal_infrastructure.md
+++ b/handbook/engineering/distribution/internal_infrastructure.md
@@ -34,6 +34,13 @@ Git URL: gitolite.sgdev.org
 
 Credentials: [1Password](https://my.1password.com/vaults/dnrhbauihkhjs5ag6vszsme45a/allitems/i5bm6syw45w2c33cvfrrlt4fhu)
 
+### Gerrit
+
+URL: https://gerrit.sgdev.org
+Git URL: gerrit-ssh.sgdev.org
+
+Credentials: Google Auth
+
 ## Customer environment replicas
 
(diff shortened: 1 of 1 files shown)


handbook/engineering/distribution/internal_infrastructure.md
```diff
@@ -34,6 +34,13 @@ Git URL: gitolite.sgdev.org
 
 Credentials: [1Password](https://my.1password.com/vaults/dnrhbauihkhjs5ag6vszsme45a/allitems/i5bm6syw45w2c33cvfrrlt4fhu)
 
+### Gerrit
+
+URL: https://gerrit.sgdev.org
+Git URL: gerrit-ssh.sgdev.org
```
bobheadxi, line 40:
```suggestion
URL: https://gerrit.sgdev.org

Git URL: gerrit-ssh.sgdev.org
```

(in markdown, there needs to be space between lines to be rendered as new lines)

image

bobheadxi (approved):

handbook/engineering/distribution/internal_infrastructure.md
pecigonzalo: This depends on the markdown format. GitHub flavored markdown for example does go to a newline without the need to haven an empty newline.

felixfbecker: Only in comments and issues. When viewing Markdown files inside GitHub, they will also ignore single line breaks (while double-line break for a new _paragraph_ always works, but idk if this is semantically a new paragraph). It is possible to force a line break though by escaping it with `\` (not all renderers support this), adding two spaces at the end (prone to be deleted by editors that trim trailing whitespace) or using `<br/>` (probably the best option).

pecigonzalo: I prefer the empty newline or double space at the end than `<br/>`. I try to avoid mixing HTML blocks to Markdown as much as possible.

PR #2071, opened 2020-11-30
Update code intelligence goals

efritz:


handbook/engineering/code-intelligence/goals.md
```diff
@@ -14,37 +14,42 @@ Progress on adoption and usage is tracked in our [Looker dashboard](https://sour
...
+1. Sourcegraph Cloud has up-to-date precise code intelligence for C++ repositories that use Bazel as a build system.
+    - Support Bazel projects in lsif-clang
+    - Auto-detect common Bazel project configurations in the auto-indexer
+    - Auto-index 5 repositories on Sourcegraph Cloud that use Bazel as a build system
+
+1. Sourcegraph Cloud has up-to-date precise code intelligence for C++ repositories that exceed 1mil SLOC.
+    - Reduce lsif-clang memory usage
+    - Auto-index 5 repositories on Sourcegraph Cloud that exceed 1mil SLOC
+
+1. _Placeholder_ (some other C++ property)
```
nicksnyder, line 44: We can add new milestones when we think of them

```suggestion
```

nicksnyder (approved):

handbook/engineering/code-intelligence/goals.md
```diff
@@ -14,37 +14,42 @@ Progress on adoption and usage is tracked in our [Looker dashboard](https://sour
...
 
 1. Increase precise code intel operations by ([*N<sub>1</sub>*][N1]+[*N<sub>2</sub>*][N2]+[*N<sub>3</sub>*][N3]) per month.
```
scalabilitysolved, line 46: @nicksnyder @christinaforney What's does success look like here?  Is success that we're on-boarding N customers and thus as a by-product code intel operations go up?  Or is it that code intel operations go up due to some combination of cloud usage/paid customers? cc/ @macraig

scalabilitysolved (commented):
Left a couple of additional questions regarding our goals @sourcegraph/code-intel

handbook/engineering/code-intelligence/goals.md
```diff
@@ -14,37 +14,42 @@ Progress on adoption and usage is tracked in our [Looker dashboard](https://sour
...
 - C++ and Java indexers are adopted by at least 3 customers each.
 - Precise code intel reaches [*N<sub>0</sub>*][N0] precise code intel operations a month.
 - CE and Sales have a significant amount of open source repos to showcase and can easily demonstrate the value of setting up precise code intelligence.
 - For languages that are particularly subject to unusual setups and tooling, our indexed open source showcase helps us determine if setup issues are caused by an unusual customer dev environment.
 
 **Milestones:**
 
-1. 🔄 Sourcegraph Cloud has precise code intelligence for 100 C++ repositories.
+1. 🔄 Sourcegraph Cloud has up-to-date precise code intelligence for C++ repositories.
     - 🔄 Improve stability and observability in lsif-clang
```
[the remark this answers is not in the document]
efritz: We could probably make these two different roadmap items and track each effort independently. I think stability is less about things like memory consumption and more about choking on certain inputs (segfaults, etc).

handbook/engineering/code-intelligence/goals.md
efritz: Yeah this was meant to be here only during the draft period of the PR.

handbook/engineering/code-intelligence/goals.md
macraig: This is how I've been thinking about it (and any input from leadership is much appreciated if it's not accurate):
Our main problem is low adoption, so we defined our goal as "Expanding adoption" (maybe increasing would be a better wording there). Therefore I think of success as codel intel operations going up as the focus (due to cloud and customer usage) instead of the other way around. That is, operations going up being the goal to achieve, and not a consequence of onboarding customers.

nicksnyder: There are two company goals that are not mutually exclusive, and these two goals apply to code intel as well.
1. Grow revenue (win customers)
1. Grow adoption (more usage, both at customers and in Sourcegraph Cloud)

Given limited engineering resources, we may need to make strategic prioritization decisions between these two goals (meta goal is to grow the team so we don't have to make tough tradeoffs like this).

In the past we have decided to focus on #1. I think this has been less successful than we would have liked because there are challenging technical problems we need to solve (e.g. lsif-clang) that are hard to solve when we don't have direct access to a customer's environment. So what should we do?

My non-binding opinion (I think ultimately the code intel team needs to decide on a defensible strategy): I _still_ think revenue/customers is ultimately the goal, HOWEVER, the most effective _means_ to achieve this goal seems to be by focusing on adoption in open source so (1) we can work through technical issues that customers are likely to experience without getting blocked and (2) so we can create a showcase of precise code intel for each language that will be useful for motivating customers to adopt precise code intel and push through any problems with their specific setup. In practice, I expect this to be a pendulum that oscillates between focusing on adoption and focusing on customers. For each quantum of work we do on adoption, that will/should enable a new set of customers to be successful. When customers run into issues, we can either directly fix those issues (when possible) or go find open source repos that we can use as a proxy to fix the issue.

I think of code intel operations as a [KPI](https://en.wikipedia.org/wiki/Performance_indicator). If we are doing the right things (expanding language support, landing at customers, building the global graph of code), then this number will be going up. If this number isn't going up over time, then we probably aren't doing the right things.

macraig (approved):

handbook/engineering/code-intelligence/goals.md
christinaforney: Everything Nick said, plus at the higher and longer term view (likely not super relevant to these goals in particular, but hopefully helpful context for Owen):

Every new customer sets up their instance with precise code intel for a majority of their code from the start because precise code intel is used to power other functionality of the product in addition to go to def/find refs.

This ties back to Nick's points: grow revenue (no deals blocked or delayed due to lack of precise intel, unlock more customer use cases) and grow adoption (this is the best way to understand all your code though data provided by our precise solution that powers search, code insights, campaigns).

PR #2077, opened 2020-12-01
Add product learning page to handbook

christinaforney:
Based on slack comments the search performance section might need to change.

From: https://sourcegraph.slack.com/archives/CHEKCRWKV/p1606774600141000


ebrodymoore (approved):
Love this. Thanks for doing it!

Joelkw:
Awesome! I have an open question about format (by team vs date) and the subsections, but those are non-blocking and can be resolved when needed as we use this more / I'll plan to add some browser extensions learnings to this in the next two weeks and make proposals in a PR then. Already excited to add to it!

Joelkw (approved):

christinaforney:
@Joelkw - totally! I wasn't sure about the format either, I expect we might have some cross functional entries? Anyhow, we can adapt the format as we add more info :)

PR #2120, opened 2020-12-04
Updating team members

virginiaulrich:
Looks like this might be the best way to update for now, but let me know if it requires any changes! And let me know, please @juliasourceress if anyone else should be a reviewer.


juliasourceress (approved):
Perfect, thank you!

virginiaulrich:
@juliasourceress all of the checks are green, but I want to explain what I did with @scalabilitysolved (who kindly jumped in to guide me). The checks failed because when I removed the line about hiring, it was trying to reference https://about.sourcegraph.com/handbook/ce/roles/head_of_customer_support, which no longer existed. Sure enough, when I go to https://about.sourcegraph.com/handbook/ce/roles, I don't see it listed. So, I removed the page. But ... is that the right thing to do? Maybe the info on the head of customer support role page needs to live somewhere else?

juliasourceress:
@virginiaulrich that works, I don't think that page needs to live elsewhere now that the role has been filled, so perfect.

PR #2149, opened 2020-12-09
Changing brex card amounts

chayim:
I noticed that the Brex amount was incorrect, versus the dashboard.


dadlerj:
This is actually correct already! Some managers who joined before this page existed (or those who have requested higher limits) have different limits, but this is the new _default_ for new managers for the future. As noted on the page, managers can request higher limits if needed, so there will always be a variety of different amounts out there.

chayim:
Closing and rejecting my change then :)

PR #2140, opened 2020-12-08
Add Code Insights team page

felixfbecker:
The goal of this is to have a page to reference when telling candidates about code insights as one of the different teams that will come out of the web team, to get them excited about this team. I.e. the target audience of this page are candidates.
It would be awesome to attract candidates that are specifically excited about code insights, but that is difficult if we don't have any documentation about this team in the handbook.

Goals were copied from the web team page.

I did not yet add anything from the usual team page structure about contact, processes, members etc, as that doesn't make sense without an operating team yet (and will be "forked" from the web team).

We can add more here that could be attractive (e.g. principles for this feature) but I think this is a good start.

PR #2153, opened 2020-12-09
Propose more focus for web team

nicksnyder:
This PR is in response to two pieces of feedback.
1. [The team feels like code insights is buried too deep](https://github.com/sourcegraph/about/pull/2140/files#r538486398) to be effective for attracting candidates.
2. The web team currently feels like a "bucket/catch all" team that owns too much given its size. This creates a feeling of overwhelm on the team.

The PR proposes a few changes:
1. Explicitly removing web platform work as an ownership area of this team. While this team has done web platform work, it has been incidental and in service to other goals. Especially looking at the documented roadmap, I think we should just be explicit that this team is not functioning as a web platform team today. We need to create that team.
2. Condense "extensions and integrations" into "integrations". This is just a naming change to avoid awkward sentences like "extensions and integrations, and code insights". In this PR "integrations" has bidirectional implications (dev tools integrating with Sourcegraph via our extension API, and Sourcegraph integrating with code hosts via our browser extensions).
3. It lifts the exciting product areas that this team owns to the top of the page, which is the most relevant content to readers outside of the team, including candidates.

I also have an opinion that we should rebrand/rename the web team to "Integrations and code insights" (or, if we are already ready to declare code insights to be a separate team, then rebrand the remaining web team to "Integrations and extensions" or something similar). This change is not implemented in this PR because that would involve a lot of renames.

### DIFF
--- a/handbook/engineering/web/code_insights.md
+++ b/handbook/engineering/web/code_insights.md
@@ -1,0 +1,23 @@
+# Code insights
+
+Code insights is the first feature in Sourcegraph that can tell you things about your code at a **high level**. Here's [the original RFC](https://docs.google.com/document/d/1EHzor6I1GhVVIpl70mH-c10b1tNEl_p1xRMJ9qHQfoc/edit) and a [demo](https://www.youtube.com/watch?v=XqeRb6Mc4Co) of a code insights prototype.
+
+Sourcegraph is in the unique position to give these insights because we have universal code search: To know _anything_ about your code at a high level means you must know _everything_ about your code at a low level.
+
+Code insights connects a lot of features that Sourcegraph already has and builds on top of them.
+We go beyond single-step code intelligence and search to connect the full cycle of analyzing (code intelligence), monitoring (code insights), and actionably changing a codebase (campaigns).
+
+Code insights will also allow to combine these Sourcegraph-created metrics with other third-party data sources.
+
+Code insights is the first feature primarily built for non-search-based user personas.
+
+## Vision
+
+1. Expose the value of Sourcegraph's knowledge of your codebase to users at all levels of an organization.
+1. Expose generalizable metrics that let our users measure and track their own goals, whether those are migrations, code smells, test coverage,  security needs, cross-collaboration, or other information about code.
+1. Make these metrics scalable across thousands of repositories on a Sourcegraph instance.
+1. Expose features that let our users build their own insights.
+1. Expose features that let users drill down into insights by teams, repositories, or other sub-metrics.
+1. Connect code insights to other Sourcegraph features – make a campaign to take action on a metric or set up code monitoring to notify you about changes to an insight.
+1. Add plug-and-play integrations to Sourcegraph code insights that let users easily integrate third-party data into their insights.
+1. Discover new, validated code metrics that engineering teams should track (and let teams track them).
--- a/handbook/engineering/web/goals.md
+++ b/handbook/engineering/web/goals.md
@@ -4,9 +4,8 @@
 
 ### Long term
 
-**_The web team has four areas of focus:_**
+**_The web team has three areas of focus:_**
 
-  1. **_Creating and maintaining a highly usable and intentionally designed webapp interface_**
   1. **_Delivering the full, unique value of Sourcegraph [extensions](https://docs.sourcegraph.com/extensions)_**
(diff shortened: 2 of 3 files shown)


Joelkw:
I agree with both pieces of feedback this PR is premised upon. 

### 1. Removing web platform work 

I have concerns about fully removing web platform work from this team unless we define an interim plan for that work. It seems we're ~2 months away from having at least two new hires who can focus on web platform work. This wasn't planned as a team, but rather part of the web team (sharing an EM and process) for the next 9+ months. (Unless we also want to change the [org growth plan](https://docs.google.com/document/d/1lzlsu9SoGSou1PTK7gp75ueYLKYitRtB8t8TxBk1irY/edit#heading=h.88ol1ws4zqar).)  

Here are some of the things that I think fall under "web platform" work that otherwise wouldn't get done – things which prior to this PR, I would expect we made progress on in the next 3 months: major scroll / spacing / accessibility bugs, navigation redesign (critical as we add new features), settings and menus design and updates, and possibly browser history fixes or alert / component designs. 

Some of these things, like urgent customer issues (or [accessibility issues](https://github.com/sourcegraph/sourcegraph/issues/16602)) and [nav bar updates](https://github.com/sourcegraph/sourcegraph/issues/16567), are (I think) critical within a <3 months timeframe. With this change, who would take this work on? 

If the answer would be "[this] team, but only when critical" then I'm not sure if that's much different than how we've currently structured this page, with web app goals listed 4th (lowest priority, closer to "only when critical") in our medium term goals at the moment. (Leaving this question/answer unsaid was the original confusion and tension that led to us explicitly adding this as an area of focus.) 

Is there a way we can function or change the wording on this page to emphasize that and try to provide a little more focus? Is there a way we can plan iterations to help with that? cc @jeanduplessis since you mentioned you've dealt with teams that have distinct focus areas before. These are all stopgap measures – I fully agree this needs to be its own team ultimately. 

There might be some further questions this raises about design work, design team goals, and if they need to adjust those goals depending on if we're now saying that [this] team won't address any non-extensions/integrations/insights work for a bit, but I'll wait to see what thoughts are about critical web platform work first. 

### 2. Condensing "extensions and integrations" into "integrations" 

I think a better word for this (which I believe I'm stealing from an old org plan I saw in our handbook) might be "extensibility."

We already know there are problems with the usage of "extensions" ([RFC 275 WIP: Resolve confusion around "Sourcegraph extensions" vs "Browser extension(s)"](https://docs.google.com/document/d/1QP1VsCJMm3vhDfTKyLnirZORGTI3LT1qFfxvQhpAXRU/edit#)). We also know there is sometimes confusion about "integrations" (which we use as both a [top-level term](https://about.sourcegraph.com/handbook/communication/content_guidelines/terminology_guidelines#writing-about-ourselves), [2](https://docs.sourcegraph.com/integration), and specifically to refer to "[native integrations](https://docs.gitlab.com/ee/integration/sourcegraph.html)" across the code hosts we offer it for).  Given that it's already confusing to have "extension" refer to both browser and Sourcegraph extensions, adding "integration" to refer to _both_ those things _plus_ our native integration is only going to add to this confusion, in my opinion.

Even if we still keep the names of the features separate, saying "the integrations team works on integrations and extensions" would already suggest we should use a different top-level term. However that RFC is resolved, I am fairly certain it won't be by re-using a different term already in use. 

Maybe "code host integrations" would refer to all things that bring sourcegraph to your code host (browser extensions + native integrations), if that solves the "extensions and integrations" case for at least the browser extensions. I would be okay with that as well, but this doesn't include (nor do I think it should include) Sourcegraph extensions, so we still need a top-level term for both. 

### 3. Reorder page

(No comments.) 

On a potential team rename, I would prefer "extensibility and code insights" to "integrations and code insights" for the prior stated reasons.

nicksnyder:
Thanks for the extra context Joel. Jean gave me some feedback that this PR might have felt disruptive to the team given some planning/discussions that were happening behind the scene and that definitely wasn't my intent.

I was trying to write down some thoughts in an async way, and a PR seemed like the best way to communicate this (I had no intention of pre-empting plans/decisions the team was making).

Jean and I are on the same page now, so I am going to close this and let the team do what it thinks is best.

Joelkw:
@nicksnyder thanks for that update! No worries – appreciate the intent.

PR #2163, opened 2020-12-11
Update careers.md

mercadon:


company/careers.md
```diff
@@ -54,12 +54,6 @@ We're hiring! Check out our open positions:
 
 - [Executive Business Administrator](../handbook/ceo/executive_business_administrator_role.md)
 
-## Apply
```
christinaforney, line 57: We probably only want to delete line 61, not the direction to apply within the job descriptions?

dadlerj: So to be totally transparent, I had thought that every JD had the email address in it. The template I've been using for years does, e.g.: look at all of these search results: https://about.sourcegraph.com/search?q=hiring%40sourcegraph.com —looks like 9 out of 24 or so?

This page feels like the page where it _should_ live to me. I know that we discussed this yesterday, but now I'm worried that I was talking about something completely different (ie., the links to the email address in the JDs themselves) 😅 ...

I hate to suggest this, but what about just changing the wording here? Something like 

>If you have any questions about a position, please submit them in the open questions box along with your application. If you have any technical or logistical questions, please reach out to us at hiring@sourcegraph.com.

To make it clear that this email is only about administrative matters?

Looking down the list of recent emails we've received to this address, the breakdown is (https://sourcegraph.slack.com/archives/C168VP0FM/p1607524586130000):

* 6 are about roles that don't have links (5 for the EBA role), and we'll be fixing this issue per @withdavidli 's suggestion
* 3 are just pure spam
* 1 is _sort of_ spam ("I am looking for a new job")
* 1 is about a role that was closed in one place (on Lever) but not in another (the handbook)
* 4 are about roles that appear on the list at the top of this message, which mention the hiring@ email address IN the job posting (the ones about the bizops and onboarding success roles)

...and that's it!

I don't feel very strongly about keeping the email here, but it does seem like the right place, and most of the problem above can be connected to other issues. Let me know what you think, and sorry for thrashing on this.

PR #2167, opened 2020-12-11
Update job description for cloud team manager

nicksnyder:
I want to reframe this goal to be better aligned with current state and needs.


withdavidli (approved):

tsenart (commented):
LGTM

PR #2209, opened 2020-12-23
add new templates

davejrt:


handbook/engineering/releases/release_deploy_sourcegraph_docker_template.md
```diff
@@ -0,0 +1,16 @@
...
+
+Arguments:
+- $MAJOR
+- $MINOR
+- $ONE_WORKING_DAY_AFTER_RELEASE
+-->
+
+# $MAJOR.$MINOR
+
+## $ONE_WORKING_DAY_AFTER_RELEASE Release deploy-sourcegraph-docker
```
bobheadxi, line 13: Currently these are a to-do item already as part of the normal release steps (it is noted in the generated PR, e.g. https://github.com/sourcegraph/deploy-sourcegraph-docker/pull/196 "follow the release guide")

handbook/engineering/releases/upgrade_managed_issue_template.md
```diff
@@ -0,0 +1,17 @@
+<!--
+DO NOTE COPY THIS ISSUE TEMPLATE MANUALLY. Use `yarn run release tracking-issue:create` from the
```
bobheadxi, line 2: I'm guessing this will be `release:close` instead?

davejrt: I would argue it's better to create all the issues at the start of the release, even if we don't complete them before the other issue is closed

handbook/engineering/releases/release_deploy_sourcegraph_docker_template.md
davejrt: I think it's fine to have both. That checklist on the other PR is easy to skip over and forget about. If we have an explicit issue that's harder to forget about.

bobheadxi: > That checklist on the other PR is easy to skip over and forget about

:( :( I think in that case we should address the issue of emphasizing that the checklist TODOs are requirements for each PR and not optional unless stated, rather than create more issues

handbook/engineering/releases/upgrade_managed_issue_template.md
bobheadxi: Hm, in that case I wonder if we should just include this in the release process? e.g. make this a requirement to closing the release tracking issue

davejrt: sorry this was a bit **strong**. My rationale is that we open all tracking issues to start the release and close them as they're done. Each task of these issues are largely independent, as they are tied to the release being cut...not it being finished. 

I hope that makes more sense.

handbook/engineering/releases/release_deploy_sourcegraph_docker_template.md
davejrt: We also need to merge the other PR before we can start on the customer-replica, so having it as a checklist item on that PR serves as a reminder and not really an issue to complete in my opinion.

davejrt: eg: we create a new branch v{MAJOR}.{MINOR}-customer-replica from the v{MAJOR}.{MINOR} branch. I think it's more important to merge this PR as part of the release day steps, then this afterwards anyway as there is currently only one customer dependent on this.

bobheadxi: Ah I might have misunderstood the steps then, in that case we should add a note to the checklist item that the generated PR does _not_ include pure-docker: https://sourcegraph.com/github.com/sourcegraph/sourcegraph@8a1e342e195644722cc6792e5f6da47b9940a740/-/blob/dev/release/src/release.ts#L496-497

I'm not sure if this is important, but this variable also implies pureDocker releases are tied to docker-compose releases: https://sourcegraph.com/github.com/sourcegraph/sourcegraph@8a1e342e195644722cc6792e5f6da47b9940a740/-/blob/dev/release/src/release.ts#L437 , would be nice to detach the two otherwise we have to block docker-compose releases as well.

Alternatively, could we change the steps to cut the `v{MAJOR}.{MINOR}-customer-replica` from the publish branch instead, while the PR is still open?

handbook/engineering/releases/upgrade_managed_issue_template.md
bobheadxi: makes sense!

handbook/engineering/releases/release_deploy_sourcegraph_docker_template.md
bobheadxi: > so having it as a checklist item on that PR serves as a reminder and not really an issue to complete in my opinion.

The intent is to have these checklist items be things that must be completed to turn a PR from "draft" to "ready for merge", since there are many things that were not possible to automate. For example, on the changeset on the main repo, the checklist includes upgrade notes and ensuring everything is merged before merging that PR: https://github.com/sourcegraph/sourcegraph/pull/16937

davejrt: I've just completed the 3.23 release, and in practice what we do is create the new customer-release branch off of the existing branch, then merge in the release branch. So in theory, I believe we could continue to create the branch the same way, however we could merge the publish branch into customer release. In doing so, we can complete all those items before closing the issue.

PR #2214, opened 2020-12-28
Proposed feature deprecation process

Joelkw:
The interactive search mode deprecation + removal caused us to realize [we don't have a good deprecation process documented in the handbook](https://sourcegraph.slack.com/archives/C0W2E592M/p1606764758299400?thread_ts=1606757921.298600&cid=C0W2E592M). 

This is a draft; suggestions are welcome.

### DIFF
--- a/handbook/product/deprecation_process.md
+++ b/handbook/product/deprecation_process.md
@@ -14,10 +14,27 @@ If the answers to these questions suggest that deprecation is the best path forw
 
 ## Documenting a feature's deprecation 
 
-You should create a blog post to publish when we deprecate the feature. The blog post should contain:
+You should then create a blog post to publish when we deprecate the feature. The blog post should contain:
 
 - Why we're deprecating the feature
 - The deprecation timeline
 - Workarounds for the feature's users, if relevant
 
-We use a blog post because it's publicly accessible, timestamped, and easy to send directly to any customers or prospects asking about the feature. 
+We document this in a blog post because blog posts are publicly accessible, timestamped, and easy to send directly to any customers or prospects asking about the feature. 
+
+## Preparing internal teams for the deprecation
+
+Once you have the blog post, you should share it with the CE team by sharing the blog post in #CE, as well as share any other context they might need in order to support customers or do high touch outreach. 
+
+You should also share the blog post in #Product so the product and engineering teams are aware. 
+
+Answer any questions or concerns about the deprecation that teams may have. 
+
+## Deprecating the feature
+
+Only after sharing this deprecation internally should you do any engineering work needed to deprecate the feature.   
+
+## Publishing the blog post 
+
+The blog post should be published when the first version that includes the deprecation is released, as a separate blog post. The release's blog post should link to the deprecation post from the release changelog section.  
+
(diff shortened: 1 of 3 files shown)


handbook/product/deprecation_process.md
```diff
@@ -0,0 +1,40 @@
+# Feature deprecation process
+
+Sometimes we need to deprecate features that we've already released. We follow this process to minimize the impact of that deprecation on both our customers and our internal teams. 
+
+## Choosing to deprecate a feature
+
+You should first determine the potential impact of the feature deprecation: 
+
+- Are any users are using the feature? Who are they? Is there another feature that also meets their needs?
```
macraig, line 9:
```suggestion
- Are any users using the feature? Who are they? Is there another feature that also meets their needs? 
```

handbook/product/deprecation_process.md
```diff
@@ -0,0 +1,40 @@
...
+
+Answer any questions or concerns about the deprecation that teams may have. 
+
+## Deprecating the feature
+
+Only after sharing this deprecation internally should you do any engineering work needed to deprecate the feature.   
+
+## Publishing the blog post 
+
+The blog post should be published when the first version that includes the deprecation is released. The release blog post changelog section should link to the deprecation post.
```
macraig, line 39: Thinking about communication around deprecating features, I'm thinking if it would also be helpful to make sure we mention it in the Product updates/Progress slide we have in company meeting. Would it help to have it as part of the documented communication process or is it already implied as it's a relevant part of the changelog?

macraig (commented):
Thanks @Joelkw for documenting this!

handbook/product/deprecation_process.md
Joelkw: I think it's worth calling out at the company meeting, and worth adding a note here – who knows if the product slide will always come straight from the changelog. I'll add a quick sentence.

Joelkw: Actually, on second thought, I feel like it's most useful in the company meeting to gather "be aware of this + speak up if there's pertinent information you have about this feature being deprecated" and Slack is a much better place for that. If folks wonder if the deprecation has happened yet, the blog post published/not published should be a good signal. So do you think #dev-announce is the right channel to get this feedback and awareness @macraig or should it be posted in something like #general instead?

macraig: That's a good point! I think I'd go for #general to make sure folks from non-dev related teams are aware (not sure if everyone non-dev actively follows #dev-announce). Another option would be #product, but we'll likely have the same problem than with #dev-announce as Slack is getting crowded and people will tend to follow channels directly related to their teams.

Joelkw: Great, I'll make that change.

handbook/product/deprecation_process.md
```diff
@@ -0,0 +1,40 @@
...
+- The deprecation timeline.
+- Workarounds for the feature's users, if relevant.
+
+We document this in a blog post because blog posts are publicly accessible, timestamped, and easy to send directly to customers or prospects.
+
+## Preparing internal teams for the deprecation
+
+Once you have the blog post, you should share it with the CE team by sharing the blog post in #CE. This gives the CE team context so they can support customers or do high touch outreach. 
+
+You should also share the blog post in #dev-announce so the product and engineering teams are aware.
```
Joelkw, line 29:
```suggestion
You should also share the blog post in #general so all teams are aware of it.  
```

Joelkw:
(@christinaforney un-assigning you and merged since Maria reviewed and there weren't any open questions. Happy to make edits in another PR.)

PR #2215, opened 2020-12-28
Adding guidelines for eng team goals format

Joelkw:
We want teams to move towards a Problem/Milestones/Outcomes goal format, so this is a take on documenting it in our handbook. 

This information is taken from product team discussions and current teams that have put goals in this format, with some influence from _Inspired_.

### DIFF
--- a/company/goals/guidelines.md
+++ b/company/goals/guidelines.md
@@ -30,14 +30,9 @@ We believe:
 
 ## Choosing goals
 
-A goal's format is:
-
-> **Goal title** \
-> Description of the goal and how to evaluate whether we met the goal (with a link to an automatically updated metrics dashboard, if appropriate).
-
 1. Each goal has a single person (not multiple people) who's ultimately responsible for it.
    - Many people can be working toward a goal, but there must be one person who's ultimately responsible.
-   - It's OK for a team to own a goal. That means the manager is ultimately responsible for it. A team should not have more than 1 or 2 goals (to make sure it's not becoming a workgroup with too many disparate goals).
+   - It's OK for a team to own a goal. That means the manager is ultimately responsible for it. A team should not have more than 1 or 2 goals (to make sure it's not becoming a workgroup with too many disparate goals). Engineering teams should all have [engineering team goals](#engineering-team-goals).
 1. No person can have more than 5 goals at once. (This number is arbitrary, but it feels right.) Fewer is better.
 1. Pick goals where failure would be painful.
 1. Pick goals where you can influence the outcome. Avoid using a lagging indicator as a goal.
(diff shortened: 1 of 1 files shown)


company/goals/guidelines.md
```diff
@@ -49,6 +44,27 @@ A goal's format is:
...
+A goal's general format is:
+
+> **Goal title** \
+> Description of the goal and how to evaluate whether we met the goal (with a link to an automatically updated metrics dashboard, if appropriate).
+
+### Engineering team goals 
+
+Each engineering team should have its own goals page with explicit goals. Because engineering team goals are larger than a single person's goal, we use the following format: 
+
+> **Goal title:** _The title for a vision that is achievable within a 6-12 month time frame._
```
nicksnyder, line 58: Is a vision that is achievable in 3-4 months too small to list as a goal?

company/goals/guidelines.md
```diff
@@ -49,6 +44,27 @@ A goal's format is:
...
+
+Each engineering team should have its own goals page with explicit goals. Because engineering team goals are larger than a single person's goal, we use the following format: 
+
+> **Goal title:** _The title for a vision that is achievable within a 6-12 month time frame._ 
+> 
+> **Problem:** _A prose description of the problem that the goal is meant to address. The "why" of the goal._  
+>
+> **Milestones:** _Checkpoints along a linear prerequisite path that would demonstrate incremental progress towards this goal. Milestones are not specific single tasks and likely encompass 1-2 months of progress. It may make sense to list milestone-specific outcomes beneath each milestone._ 
+>
+> **Outcomes:** _The quantitative and qualitative metrics that demonstrate the goal has successfully solved the problem. Outcomes are not "tasks completed;" they are signals that the problem itself is solved._
```
nicksnyder, line 64: I have a sense that outcomes should be tied to milestones. Otherwise, I worry there is duplication in the best case and a disconnect in the worst cast, between the actual milestones and the outcomes we are trying to achieve.

nicksnyder (approved):
This is an improvement, thanks for writing this up. I added some non-blocking questions/comments.

Also, I would love to add guidance on what makes a good goal. Christina has described it live in eng manager sync, and probably in PM syncs. Here are my notes from the former:

- Goals should be static, 6-8 month timeline to create a vision and drive focus
- Goals should help us say ‘no’ to things
- Each goal should have a top-level title explaining it, and should be finite.
- Write down what the problem is that we want to solve
- How do we know we have met the goal? Quantitative
- How do we know we’ve made forward momentum? Clear milestones, clear end-states, clear roadmap
- Cloud team is a great example of Goals with Milestones broken down
- You know goals are good when you have a shared vision within the team, when you can confidently say “no” because it’s not The Thing Right Now

company/goals/guidelines.md
Joelkw: I don't know that it is, this was just taken from [a description](https://docs.google.com/document/d/1xszELbG-5urpXf-FHPc1tLDMBb_PM3MTFjFSOFRHPTo/edit#bookmark=id.kkwv4akl15r4) @christinaforney gave in a product sync one day. 

I wonder if there are any 3-4 month goals that don't roll up into larger 6-12 month goals, however, and perhaps the implication was that 3-4 month goals are better as milestones in that larger goal?

company/goals/guidelines.md
Joelkw: I agree. I actually prefer a case where they are more tightly coupled – I'll update the description/format so milestones have outcomes after I play with it a bit on my teams' pages to make sure the template formatting looks good.

Joelkw: I've inlined outcomes into milestones in this template now. I'm not sure if "multiple outcomes" is better than just having a longer single outcome (example – "code insights drives 5 new sales AND 3 existing customers pay more to use code insights" vs listing those things as separate outcomes under a milestone) so perhaps we want to change that allowance. 

I also wonder if it might now make sense to give a format template and now move descriptions out of the sections or if it's still clear enough (visually) as to what a goals page should look like as is; I have no preference and see advantages to both. 

For an example of this, you can see [the PR for the code insights team page](https://github.com/sourcegraph/about/pull/2140/files#diff-851a820fbd3b799a3881f56e2de520c2e7b0e508387185d0312230e32dd315fbR28) that I've changed to now inline outcomes beneath milestones.

nicksnyder: I like the multiple outcomes per milestone. I think the current doc is clear and we don't need a separate template doc. Updates to code insights goals/milestones look great!

nicksnyder (approved):

company/goals/guidelines.md
nicksnyder: This is fine as is. Upon second reading, I think the "within" existing languages allows it anyway.

Joelkw:
@christinaforney I'm not going to block on you before merging this because our current format is entirely undocumented except by example, but I fully expect that you may have feedback, still think it would be valuable for you to review this page sometime before or when new PMs/teams might use it to set goals, and am happy to make those updates once you have.

PR #2236, opened 2021-01-04
Update internal_security.md

justdueck:
Initial draft of internal security best practices and anti-phishing measures.


dadlerj (approved):
Awesome, thanks!

Can you add a screenshot of what these look like as an example?

image

dadlerj:
We also might need to link to this page from another one for this to pass CI. Happy to help with these fixes if useful.

PR #2255, opened 2021-01-05
Latest news updates, plus cleaned up json mapping to be more efficient

aileenrose:


lorihey (commented):
Aileen, this article is behind a paywall, not sure if we should include the link to it here, what do you think? Do we have other news items that are behind paywalls?

aileenrose:
> Aileen, this article is behind a paywall, not sure if we should include the link to it here, what do you think? Do we have other news items that are behind paywalls?

Yes, there are other articles behind paywall i.e. WSJ.

lorihey:
ok

lorihey (approved):

PR #2270, opened 2021-01-06
Add Dhall to distrib tech stack

daxmc99:
We should include Dhall on our tech stack

### DIFF
--- a/handbook/engineering/distribution/tech_stack.md
+++ b/handbook/engineering/distribution/tech_stack.md
@@ -1,6 +1,6 @@
 # Distribution tech stack
 
-This document lists the Distribution team's technologies and best practices for each within the Distribution team.
+This document lists the technologies used by the Distribution team, what they are used for, and best practices for each within Distribution.
 
 Also see [internal infrasturcture](./internal_infrastructure.md) and [tools](./tools/index.md) for specific infrastructure and tools.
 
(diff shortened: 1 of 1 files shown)


handbook/engineering/distribution/tech_stack.md
```diff
@@ -43,3 +43,7 @@ Also see [internal infrasturcture](./internal_infrastructure.md) and [tools](./t
 ## Kubernetes
 
 **Use cases:** Deploying services
+
+## Dhall
+
+**Use cases:** Providing configuration for our deployment supported deployment types.
```
bobheadxi, line 49:
```suggestion
**Use cases:** Providing configuration for our deployment-supported deployment types ([work in progress](https://github.com/orgs/sourcegraph/projects/71)).
```

handbook/engineering/distribution/tech_stack.md
```diff
@@ -1,12 +1,12 @@
 # Distribution tech stack
 
-This document lists the technologies used by the Distribution team, what they are used for, and best practices for each within Distribution.
+This document lists the Distribution team's technologies and best practices for each within the Distribution team.
```
bobheadxi, line 3: What was wrong with the original description? I think the new one is a bit confusing:

* these aren't really Distribution's technologies (we just use them)
* a goal of this document should be to describe _what_ each item is used for, so that we can better choose when to use what (especially since the document arose over discussions on when to use TS), which the new description removes

daxmc99: This change was driven by grammar mostly. The previous sentence was written in the passive voice and had a comma splice. I also consider "Best Practices" inclusive of _what_ something would be used for. 
Happy to revert though.

bobheadxi (approved):

PR #2283, opened 2021-01-08
move goals to separate file

nicksnyder:
I moved this to a separate file to be consistent with other teams. Also, I added a title for the goal. My proposal is "Code insights generates revenue in FY22" based on the content already there, but feel free to adjust wording.


Joelkw (approved):
Yep, now that code insights has a complete goal this makes sense, thanks for starting this! Added a heading + a link from the company/goals page so it pulls in and this otherwise looks good.

PR #2293, opened 2021-01-10
Add Azure DevOps instructions

sqs:


unknwon (approved):
How do we know who has the permissions to do so? (besides you)

sqs:
@unknwon I don’t know and I spent several unsuccessful minutes trying to find out. I’ll just do these 1 by 1 for now unless it becomes a greater pain.

sqs:
@unknwon I don’t know and I spent several unsuccessful minutes trying to find out. I’ll just do these 1 by 1 for now unless it becomes a greater pain.

PR #2301, opened 2021-01-11
Update RFC docs for private folder

dadlerj:


christinaforney:
There's not a way to have them in the same folder? I really liked that all RFCs were in the same place - makes it easier for numbering, discovery, and search.

dadlerj:
Yeah, totally fair concern, this sort of sucks. I didn't see any way to do this in a shared drive, but I'll keep a note to look into this in the future!

nicksnyder (approved):

PR #2317, opened 2021-01-12
Remove Campaigns from list and update open roles

macraig:
Added all areas in need of a PM split by business need to be consistent with the Product Manager page.


christinaforney (approved):
Thanks for catching this!

PR #2324, opened 2021-01-13
Include engineering incident response playbook

christinelovett:
Added the link to Engineering's Incident Response Playbook to work through common failure scenarios.


pecigonzalo:
FYI, those are targeting our internal infrastructure for the most part, but a lot of the content is relevant. 

We also added (and are refining) the docs in [dashboards and what they show](https://docs.sourcegraph.com/admin/observability/dashboards )[alerts and known solutions (if any)](https://docs.sourcegraph.com/admin/observability/alert_solutions) and [troubleshooting](https://docs.sourcegraph.com/admin/observability/troubleshooting).

I don't know if they are worth referencing here.

PR #2346, opened 2021-01-15
Adding PTO

emchap:
I searched "PTO" in the handbook because that's the term I'm familiar with and couldn't find anything relevant to my question ("how do I request time off"). Until the target page is folded in to the handbook, is adding this okay?


dadlerj (approved):
Thanks!

jeanduplessis:
+1 it took me a while to unlearn the PTO term when I joined as its what I was used to as well.

PR #2363, opened 2021-01-20
Clarify the web team's Slack channel usage

jeanduplessis:


handbook/engineering/web/index.md
```diff
@@ -97,6 +97,19 @@ Here are some of the technologies we use to deliver on our goals:
...
+### Communication
+
+We recognize that frequent, open communication is key to the success of our team, especially in an all-remote environment.
+We default to asynchronous communication in Slack and GitHub issues over other mediums (video calls, emails) as we are respectful of our teammates' time. 
+
+The team communicates in the following channels in Slack:
+
+[**#web**](https://sourcegraph.slack.com/archives/CMT39K56Z): All work related commmunication happens in this channel. This is also the primary channel for teammates outside of the team to reach out to us.
+
+[**#web-chat**](https://sourcegraph.slack.com/archives/C01EM5J1NF8): Our daily standups (facilited through Geekbot) go into this channel as well as all non-work related commmunication. Random conversations, banter, jokes etc. are all welcome here.
```
nicksnyder, line 109: Was it an explicit decision to not follow the convention that other teams have adopted?

(i.e. `web-team` for internal team stuff and `web-chat` for external)?

jeanduplessis: @nicksnyder no it was not, in fact at the time of naming the channels there were some confusion as to what the convention actually is. I'm happy to align it with the convention- it might actually be good if we document that convention for future teams spinning up.

nicksnyder: Yeah, I don't think this convention is documented anywhere. Other teams have found the convention very helpful. Want to take a pass at documenting this?

PR #2371, opened 2021-01-20
Web team - Add note about tracking issues

umpox:


umpox:
@5h1rU for visibility

5h1rU:
### Detail
We have been working on this idea to improve the visibility of task progress and have more context about the different workflows used across other teams here at Sourcegraph.

**Note:** _We encourage you to share your thoughts here and ask any questions you consider pertinent._

### How was this process?
@umpox and @5h1rU  decided to talk with managers on different teams to learn how they deal with big tasks and their progress. We then talked about each meeting's conclusion and wrote a document with the most significant points. Here we have our notes.

### Goals

- Have a starting point to discuss with the team.
- Learn from the current process.
- Look for the best options.
- Iterate.
- Improve.

### Highlights

- It is challenging to know the progress of the tasks, not only in the web team.
- We don't want to suggest moving the workflow out of GitHub.
- None of the options provided by GitHub is a silver bullet.
- The success of any of the strategies needs a full commitment from all the actors.

### Options
#### Github Milestones
##### Pros:  
 - A single issue can belong to many milestones.
- From an open-source perspective is the most popular way to tag issues.
- UI progress bar.
- The percentage of completion based on how many issues are closed and belongs to the milestone.
- UI Due date
- Description field
##### Cons:
- An issue only can belong to one milestone.
- Milestones are only available by repo.
#### Bigger Tracking Issues
##### Pros:
- It is the 'de facto' way to do it.
- The handbook has good documentation about it.
- A bot is in charge of update it.
- It is centralized.
##### Cons:
- It doesn't have a progress bar.
- It doesn't have a progress UI.
- Big tasks are hard to follow.
- If you have big tasks, you need to create a new tracking issue and assign a parent.
#### GitHub Projects
##### Pro: 
- Projects allow us to add tasks from all repositories
- The boards are pretty customizable.
- An issue can be assigned to multiple projects.
##### Cons:
- Splitting the work into different projects can be overwhelming.
- Many tasks are small and isolated. Including them in a specific project can be challenging.

### Thoughts
- Follow a process with only one strategy would derive to different problems.
- Each engineer needs to be in charge of the own track and update the progress properly.
- It is a large topic.

handbook/engineering/web/index.md
```diff
@@ -118,6 +118,9 @@ We use the backlogs for tracking bugs, small features, and unplanned work. We do
 
 To add an issue, tag it `team/web` to notify the web team PM and put it in the "to triage" column of the board. Unless you're directly asked, only web team members should move issues out of the "to triage" column on the board. A web team member will confirm the issue is web team-related, and then move it to the appropriate column.
 
+#### Tracking Issues
+The web team makes use of [tracking issues](../tracking_issues.md) for tracking progress on the implementation of new features. When starting work on a new and significant feature, the web team should ensure that a tracking issue is created appropriately.
```
Joelkw, line 122: @umpox @5h1rU thanks for the proposal! Could you add a little (1-2 sentences or phrases) more guidance on what a "new and significant" feature is so that new teammates or others could know when to expect a tracking issue?

Joelkw (commented):
Thank you for putting this together! It seems like you've gone with tracking issues based on the PR, but you list 3 options with pros/cons – what was the "deciding reason" that made you pick tracking issue?

handbook/engineering/web/index.md
umpox: Have updated!

> The web team should ensure that a tracking issue is created when starting work on features that are expected to take longer than a single iteration to deliver.

I think this best summarises where our potential problems could lie. Additionally, if a single tasks is expected to take longer than a single iteration, then this will encourage us to further break down that work into manageable chunks

umpox:
@Joelkw We think tracking issues show the highest effectiveness with the lowest disruption to the teams process. We're already writing these and many other teams are too. We think it's better to focus on fixing the immediate problems we're facing as a team by using something which we can be confident has proven effective before. There's definitely scope to investigate projects/milestones as an alternative to tracking issues, and possibly encourage on a broader scale, in the future.

jeanduplessis (approved):

PR #2372, opened 2021-01-20
Adding comprensive demo script

emchap:
As discussed in our team meeting last week, our demo flow right now does not have a super in-depth script, and the 10-step flow lives in a GDoc and isn't targeted to the CEs specifically. I wanted to take a stab at writing a step-by-step flow that we could point new hires at, and which mentions the new demo practice meeting.

I wanted to get everyone's thoughts on the script as outlined and this approach in general.


justdueck (approved):
This is great!! Thanks for putting this together!

christinelovett (approved):
Awesome. Thank you @emchap!

emchap:
Since Jonah and Christine gave the go-ahead and are most senior, I'm going to go ahead and merge this in, but am still 100% open to feedback and edits if anyone wants to make them/has any.

PR #2376, opened 2021-01-21
update CE team page

sqs:
- Update the short team description at the top
- Alphabetize team members by surname
- Add meetings
- Mention that we are both pre-sales and post-sales


handbook/ce/index.md
```diff
@@ -25,26 +35,34 @@ The Customer Engineering team partners with the [Sales team] (../sales/index.md)
...
+  - We will be hiring Customer Support Engineers soon. Our Customer Engineers will continue in their support rotation for now.
 - [Quinn Slack](../../company/team/index.md#quinn-slack) (CEO) is leading Customer Engineering for now.
   - We are hiring for VP Customer Engineering and will have the role posted very soon.
 
 ## Roles
 
 We're hiring! Join us. See [roles](./roles/index.md) page.
+
+## Meetings
+
```
mike-r-mclaughlin, line 59: Could we add a note about the Friday product backlog review meeting and how to join it if needed?

handbook/ce/index.md
```diff
@@ -25,26 +35,34 @@ The Customer Engineering team partners with the [Sales team] (../sales/index.md)
...
+  - [Nick McMillen](../../company/team/index.md#nick-mcmillen-he-him)
 - Customer Support
   - [Virginia Ulrich](../../company/team/index.md#virginia-ulrich-she-her) (Head of Customer Support)
-  - We will be hiring support engineers soon. Our Customer Engineers will continue in their support rotation for now.
-- Training
+  - We will be hiring Customer Support Engineers soon. Our Customer Engineers will continue in their support rotation for now.
 - [Quinn Slack](../../company/team/index.md#quinn-slack) (CEO) is leading Customer Engineering for now.
   - We are hiring for VP Customer Engineering and will have the role posted very soon.
 
 ## Roles
```
mike-r-mclaughlin, line 54: This section seems unnecessary.  Maybe we could combine it with "Members" and add a link to the role in the header along with a short description of the role with the member names below.  Alternatively, you did cover the roles in the main "Customer Engineering" section so maybe we move the definition of the roles down to this section explicitly.

handbook/ce/index.md
```diff
@@ -1,6 +1,16 @@
 # Customer Engineering
 
-The Customer Engineering team partners with the [Sales team] (../sales/index.md) from a technical and product expertise perspective to ensure that all of Sourcegraph's prospects, customers, and users get the most possible value from it. This includes showing product demos, learning and understanding prospects' internal development environment and processes, helping companies integrate Sourcegraph into their engineering stack and internal processes to maximize value to their engineering teams, helping prospects run trials to recognize Sourcegraph value firsthand, building relationships with our accounts' champions and users, communicating the value of Sourcegraph, growing engagement and owning product training and technical support, and giving as many Sourcegraph "Aha!" moments as possible.
+On the Customer Engineering team, we help developers use Sourcegraph to solve their organization's Big Code problems. 
+
+- [**Customer Engineers**](roles/customer_engineer.md) help our developer users and champions get more value from using Sourcegraph.
```
emchap, line 5: I don't feel super strongly about this but I did like the idea that the previous description highlighted about us really helping the customers realize the value. We help them use it but we also help them recognize how they could use it, if that makes sense?

virginiaulrich (approved):
I gave a 👍🏻 to some of Mike's comments that resonated with me and/or already addressed my observations. All in all, I quite like these changes and appreciate you making them, Quinn. Clarity always = ✨!

handbook/ce/index.md
sqs: I will remove the separate roles page and inline them here.

handbook/ce/index.md
sqs: To "help our developer users and champions get more value from using Sourcegraph", I'll try adding "...in more and better ways". Does that capture it?

emchap: I think so!

PR #2386, opened 2021-01-22
Fix spelling error in canonical link

aileenrose:


aileenrose:
Staged handbook locally to confirm the fix.

PR #2387, opened 2021-01-22
distribution: update create-releases goal, update milestones

bobheadxi:
- close out `Releases can be done in a single day` 🎉 
- update goal for the next milestone
  - remove "single action", due to a preliminary investigation indicating this is likely not possible (at least without a hefty amount of work that is IMO not really worth it since going from 2~3 steps -> 1 step feels like significantly diminishing returns). more here: https://github.com/sourcegraph/sourcegraph/issues/9252#issuecomment-761713632
  - initialize project for next milestone: https://github.com/orgs/sourcegraph/projects/131
  - push back estimate to Q3 due to priority shifts. might be possible that we don't even get around to it this year? there's also a bit of a consensus that we're in a pretty good spot right now for releases and this is not particularly urgent compared to other things coming up (dhall, metrics, pipeline UX)

### DIFF
--- a/handbook/engineering/distribution/goals.md
+++ b/handbook/engineering/distribution/goals.md
@@ -31,7 +31,7 @@ We will start by looking at our Kubernetes deployment and working on an easier u
   - Design customization workflow. _Estimated: FY21-Q1_
     - Potentially Kustomize would still be used for last-mile changes and non-standard derivations.
 
-### Any engineer at Sourcegraph can create a release for all of our supported deployment types by running a single action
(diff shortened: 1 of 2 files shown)


pecigonzalo:
@bobheadxi I think it might good to close the goal and mark the last milestone as "descoped" (with a brief explanation of why) and creating a new goal for the new one as we dont know when/if we will work on it and otherwise that goal will remain open for an unknown amount of time.

bobheadxi:
Updated the PR @pecigonzalo !

pecigonzalo (approved):

PR #2398, opened 2021-01-24
Design and interaction guidelines

rrhyne:


quinnkeast (approved):
Happy to get these into the handbook ASAP! Thanks for doing this.

Will think about further changes once we've got this up.

handbook/product/design/design-and-interaction-guidelines.md
```diff
@@ -0,0 +1,107 @@
...
+
+_Clearly recognizable links help orient users to common signup actions._
+
+<img src='./large-margin-example.png' />
+
+_The large margins around the Kubernetes logo convey to the user: “This page is focused on Kubernetes”._
+
+### Web application icons
+
+In the web application sections, icons should always have an accompanying label. See this article on [Icon Usability](https://www.nngroup.com/articles/icon-usability/) from the Nielson Norman group for more information. Note that this stance is contradictory to the requirements for icons in search application pages.
```
AlicjaSuska, line 61: @rrhyne I wouldn't say it's ALWAYS necessary to have labels - for ex. 'info' icons with the tooltip on hover do not need labels. How about we change the wording to 'In the web application sections, icons should **generally** have an accompanying label.'?

handbook/product/design/design-and-interaction-guidelines.md
```diff
@@ -0,0 +1,107 @@
...
+
+* [Search results](https://sourcegraph.com/search?q=repo:%5Egithub%5C.com/sourcegraph/sourcegraph%24+file:.*/go+auth&patternType=literal)
+* [File views](https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/package.json)
+
+<img src='./margins-example.png' />
+
+_Small margins and reduced padding help to minimize scanning and scrolling. This allows developers to see code faster… and fit more of it on the screen._
+
+
+### Icons and labels
```
AlicjaSuska, line 31: I would consider changing the order here:

1. Code page design
2. Web app page design
3. Icons and labels - both for code page and web page -> it will be easier to compare those approached without jumping between the sections


@rrhyne @quinnkeast wdyt?

AlicjaSuska (changes requested):

handbook/product/design/design-and-interaction-guidelines.md
rrhyne: Good point.

rrhyne, line 61:
```suggestion
In the web application sections, icons should generally be accompanied by a label. See this article on [Icon Usability](https://www.nngroup.com/articles/icon-usability/) from the Nielson Norman group for more information. Note that this stance is contradictory to the requirements for icons in search application pages.
```

handbook/product/design/design-and-interaction-guidelines.md
rrhyne: Fine with me, would you mind making that change?

handbook/product/design/design-and-interaction-guidelines.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Two design styles
+Sourcegraph utilizes two design and interaction styles which are tailored to the content they display:
+
+Search - An IDE like, full screen, high density, low contrast ui (for elements other than code) which centers on the display of code
+Web application - a more traditional web application style with larger margins and padding, increased contrast and standard web page design experience requirements
+
+## Code pages design
+
+In Sourcegraph’s search and file experiences, code is the most important element. The design of these pages reflects that priority. Contrast is balanced so that code stands out above all else. Panels clearly segment content and tools from the code. Pages are full width, and margins and padding are slim to increase the density of information which helps us display as much code on the screen as possible.
```
rrhyne, line 17:
```suggestion
In Sourcegraph’s search, file, diff and other views, code is the most important element. To collectively refer to these pages we use the phrase 'code pages'. The design of these pages should focus on code as the most important element. Contrast is balanced so that code stands out above all else. Panels clearly segment content and tools from the code. Pages are full width, and margins and padding are slim to increase the density of information which helps us display as much code on the screen as possible. 
```

handbook/product/design/design-and-interaction-guidelines.md
rrhyne: This should be resolved now, thanks for the suggestion.

handbook/product/design/design-and-interaction-guidelines.md
rrhyne: I went to address that and realized I didn't understand what you meant. The icons and labels are sub-sections of each main topic (code pages and web pages) as they are treated in different ways for each. 

So in that sense, I don't think re-ordering makes sense. If Quinn does, we can address this in a separate pull request.

AlicjaSuska: @quinnkeast wdyt?

PR #2401, opened 2021-01-25
Adding on-call Distribution info

emchap:
This was the page I found when trying to escalate an issue from Yelp, and it didn't seem to have a reference to the support engineer rotation on Distribution.

@pecigonzalo Does this describe the process you want us to follow?

### DIFF
--- a/handbook/ce/ce_to_eng_handover.md
+++ b/handbook/ce/ce_to_eng_handover.md
@@ -2,7 +2,7 @@
 
 When a prospect or customer files a new support ticket or raises a new issue, CE and Engineering should follow the steps below to ensure the issue is handled successfully.
 
-If something hasn't risen to the level of creating an issue (for example, you're trying to confirm if something reported by a customer in Slack is intented or not), you can also contact the [Distribution on-call engineer](../engineering/distribution#contacting-the-support-engineer).
+If something hasn't risen to the level of creating an issue (for example, you're trying to confirm if something reported by a customer in Slack is intented or not), you can also contact the [Distribution on-call engineer](../engineering/distribution#contacting-the-support-engineer) if it is a Distribution-related issue. Escalation processes vary per-team; check their Handbook documentation for more info.
 
 ## Principles
 
--- a/handbook/ce/ce_to_eng_handover.md
+++ b/handbook/ce/ce_to_eng_handover.md
@@ -2,6 +2,8 @@
 
 When a prospect or customer files a new support ticket or raises a new issue, CE and Engineering should follow the steps below to ensure the issue is handled successfully.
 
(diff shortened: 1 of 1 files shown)


pecigonzalo:
I would not merge this as it only applies to Distribution related issues, Im not sure how other teams route their requests at the moment.

pecigonzalo:
I believe this could still be misleading as it only links to our contact information. I tried flipping it around to something like 

```
Escalation processes vary per-team; check their Handbook documentation for more info. 

Teams with a rotation:
- Distribution issues [contact](../engineering/distribution#contacting-the-support-engineer)
```

Or similar.

There is also a page that outlines how to route each support request, maybe its better to add this information there?

emchap:
@justdueck Hey, wanted to get your thoughts here since you have the deepest awareness of our internal documentation, I think. Gonza was pointing out that Distribution has an on-call rotation; I was trying to get it added to the documentation that I found when I was trying to ask them a question. Does it make more sense to link this to the area ownership page and highlight the Distribution on-call schedule there, or keep it here? (Or some other option.) I'm not 100% on what purpose this page is supposed to serve versus that one.

justdueck:
@emchap to Gonza's point, I can see the potential for confusion here. I think as a general rule, we should link back to the main routing issues page (https://about.sourcegraph.com/handbook/ce/routing_questions) side note: we should add rotation info there as well. 

Using Gonza's proposal, something like this may work: 

```
Escalation processes vary per-team; use the [customer issue routing glossary](https://about.sourcegraph.com/handbook/ce/routing_questions) or check their Handbook documentation for more info. 

Teams with a rotation:
- Distribution issues [contact](../engineering/distribution#contacting-the-support-engineer)
```

emchap:
@justdueck Okay! To mine, though, it feels weird that "ce-to-engineering handover" doesn't cover this process as well. As a new hire trying to figure out "how do I ask about this in Slack", I found this document before the routing one, which seemed much more about which _team_ to give an issue to. 

What's the intended point of this document/context for it being separate? Can we merge them or change the name on this one? If this is just the process for how to create and tag a GH issue, let's brainstorm on a name that reflects that.

justdueck:
@emchap good call. I think we should limit this page to just explaining how to create GH issues. For actually reaching out to eng teams or sending issues to them, we can use the routing issues glossary and link that glossary to this page with an explanation like "this guide explains how to create GH issues. once created, use this routing glossary to notify the appropriate team". I think we should also re-name this page to something like "creating GitHub Issues" to minimize confusion. 

then we could add the distribution teams on-call info to the routing issues glossary instead of on this page. Does that align with what you're thinking?

emchap:
@justdueck Yeah, that works for me and resolves my initial confusion! I'm going to close this and just make that newly-oriented set of changes in a different PR.

PR #2402, opened 2021-01-25
remove bizops analyst role

ebrodymoore:


dadlerj (approved):

dadlerj:
Thank you! Not yet, if you don't mind

dadlerj:
Actually, yes, if you don't mind, please do remove it as well. Thank you!

PR #2405, opened 2021-01-26
Add link to design workflow from eng onboarding

Joelkw:
@AlicjaSuska and I were chatting and thought we should add this to eng onboarding so all engineers are aware of how they can request or track design work. 

I think it applies to more than just frontend engineers since we encourage all of our engineers to take active roles in surfacing design issues, and because someone may be working on the backend that's dependent on a frontend design or is requested to comment on a design and they need to know what "stage" the design is in. I think that this section is a good place to link from if we agree that we want to include this at the onboarding level.


nicksnyder (approved):

Joelkw:
(@AlicjaSuska just noting this exists; merging/doesn't need your explicit review so feel free to tweak the wording if it's not what you envisioned.)

AlicjaSuska:
Thank you @Joelkw

PR #2406, opened 2021-01-26
Adds Slack channel naming convention for teams

jeanduplessis:
As per the discussion here, https://github.com/sourcegraph/about/pull/2363#discussion_r561197863, we are documenting a Slack channel name convention for engineering teams.

### DIFF
--- a/handbook/engineering/index.md
+++ b/handbook/engineering/index.md
@@ -70,6 +70,18 @@ Our engineering organization is divided into mission based teams that contain th
 
 Start a discussion with your manager if you are interested to switch teams. Your manager will chat with you to understand what you want and then propose next steps. Team transfers need to be approved by the [VP of Engineering](roles.md#vp-engineering).
 
+### Communication
+
+#### Slack channel convention
+
+Engineering teams follow the following convention when creating Slack channels for both internal (own team) and external (all teammates) communication.
+
+1. ${TEAMNAME}-team for internal team stuff. E.g. `web-team`
+1. ${TEAMNAME}-chat for external team communication. E.g. `web-chat`
+
+Teams are encouraged to keep their channels public and to ensure all work related conversation happens in the open in their relevant `-chat` channel.  
+If teams require additional channels they should follow the convension of prefixing the channel name with their team name. E.g. `web-onboarding`
+
 ## Open positions
 
 [Open positions](hiring/index.md#open-positions)
--- a/handbook/engineering/index.md
+++ b/handbook/engineering/index.md
@@ -74,13 +74,12 @@ Start a discussion with your manager if you are interested to switch teams. Your
 
 #### Slack channel convention
 
-Engineering teams follow the following convention when creating Slack channels for internal (own team) and external (all teammates) communication.
+Our goal is to facilitate open and transparent communication both within teams as well as between teams. Since these are two different audiences, we recommend that each team maintains two separate channels in Slack following these naming conventions:
 
-1. ${TEAMNAME}-team for internal team communication like standup notes, social chats, etc that does not relate to the team's goals or work they are doing. e.g. `web-team`
-1. ${TEAMNAME}-chat for general work related team communication like discussing goals, requesting help on an issue, asking general questions relating to the team's area of responsibility. e.g. `web-chat`
+1. `${TEAMNAME}-chat` is the default channel for discussing work that the team is doing and fielding questions from outside of the team. If you have a question to ask another team, ask it in their `-chat` channel.
+1. `{$TEAMNAME}-team` is a secondary channel that the team can use to communicate about topics that are not related to the work of the team and clearly only relevant to members of the team itself (for example: social chit chat, notifying team you will be late to a team meeting, scheduling a team meeting or offsite). This channel is *not* for having internal team discussions about work the team is doing; those discussions are potentially relevant to people outside of the team and should be done in `-chat`. The team should treat this as their own channel, even if non-teammates have joined. We encourage teams to keep this channel public for transparency.
 
-Teams are encouraged to keep their channels public and to ensure all work related conversation happens in the open in their relevant `-chat` channel.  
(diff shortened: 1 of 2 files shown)


handbook/engineering/index.md
```diff
@@ -70,6 +70,18 @@ Our engineering organization is divided into mission based teams that contain th
 
 Start a discussion with your manager if you are interested to switch teams. Your manager will chat with you to understand what you want and then propose next steps. Team transfers need to be approved by the [VP of Engineering](roles.md#vp-engineering).
 
+### Communication
+
+#### Slack channel convention
+
+Engineering teams follow the following convention when creating Slack channels for both internal (own team) and external (all teammates) communication.
+
+1. ${TEAMNAME}-team for internal team stuff. E.g. `web-team`
```
chayim, line 79: The manual has uses of e.g. rather than E.g.

handbook/engineering/index.md
```diff
@@ -70,6 +70,18 @@ Our engineering organization is divided into mission based teams that contain th
...
+1. ${TEAMNAME}-chat for external team communication. E.g. `web-chat`
```
chayim, line 80: Same

handbook/engineering/index.md
felixfbecker, line 79: Could we define "internal team stuff" and "external team communication" more clearly? I am worried that this leaves a lot of room to interpretation, and communication will drift to be mostly in the `-team` channels because they are argued to all be "internal team stuff". E.g. is a discussion around a decision towards the team's goals or an implementation discussion "internal team stuff", or "external team communication"? They could be argued to be internal, but I think as a company we profit from keeping those discussions in the external channels where people can subscribe if they are interested and chime in if they have something valuable to contribute. If we don't call this out here explicitly (in accordance with our openness value) I'm worried new teammates (coming from other companies with different cultures) will assume that the external channel is only for  _announcements_ to the wider company and everything else becomes effectively _internal by default_, when it should be the other way around (external/open by default).

The sentence below about "all work related conversation happens in the open in their `-chat` channel" does say this, but it could be missed easily and the two bullet points here don't intuitively align with that statement.

handbook/engineering/index.md
pecigonzalo:
```suggestion
1. ${TEAMNAME}-chat for general team communication. E.g. `web-chat`
```

handbook/engineering/index.md
jeanduplessis: @felixfbecker I updated the copy slightly to convey this message. Let me know if you think it still needs further tweaking.

handbook/engineering/index.md
```diff
@@ -70,6 +70,18 @@ Our engineering organization is divided into mission based teams that contain th
...
 Start a discussion with your manager if you are interested to switch teams. Your manager will chat with you to understand what you want and then propose next steps. Team transfers need to be approved by the [VP of Engineering](roles.md#vp-engineering).
 
+### Communication
+
+#### Slack channel convention
+
+Engineering teams follow the following convention when creating Slack channels for internal (own team) and external (all teammates) communication.
+
+1. ${TEAMNAME}-team for internal team communication like standup notes, social chats, etc that does not relate to the team's goals or work they are doing. e.g. `web-team`
+1. ${TEAMNAME}-chat for general work related team communication like discussing goals, requesting help on an issue, asking general questions relating to the team's area of responsibility. e.g. `web-chat`
```
chrispine, line 80:
```suggestion
1. ${TEAMNAME}-chat for general work-related team communication like discussing goals, requesting help on an issue, asking general questions relating to the team's area of responsibility. e.g. `web-chat`
```

chrispine (approved):
LGTM

handbook/engineering/index.md
```diff
@@ -70,6 +70,18 @@ Our engineering organization is divided into mission based teams that contain th
...
+
+Teams are encouraged to keep their channels public and to ensure all work related conversation happens in the open in their relevant `-chat` channel.  
+If teams require additional channels they should follow the convension of prefixing the channel name with their team name. e.g. `web-onboarding`
```
nicksnyder, line 83: I am a bit worried that this explanation over indexes on examples and doesn't describe intent enough (to make it easier to figure out what to do with things that don't fit into the given examples).

What do you think about this?

```suggestion
Our goal is to facilitate open and transparent communication both within teams as well as between teams. Since these are two different audiences, we recommend that each team maintains two separate channels in Slack following these naming conventions:

1. `${TEAMNAME}-chat` is the default channel for discussing work that the team is doing and fielding questions from outside of the team. If you have a question to ask another team, ask it in their `-chat` channel.
1. `{$TEAMNAME}-team` is a secondary channel that the team can use to communicate about topics that are not related to the work of the team and clearly only relevant to members of the team itself (for example: social chit chat, notifying team you will be late to a team meeting, scheduling a team meeting or offsite). This channel is *not* for having internal team discussions about work the team is doing; those discussions are potentially relevant to people outside of the team and should be done in `-chat`. The team should treat this as their own channel, even if non-teammates have joined. We encourage teams to keep this channel public for transparency.

If teams require additional channels they should follow the convention of prefixing the channel name with their team name. For example: `web-onboarding`
```

chrispine: I really like this suggestion 👍🏻

jeanduplessis: Thanks for the rewording @nicksnyder! 100% agree with your rephrasing to show intent.

handbook/engineering/index.md
```diff
@@ -70,6 +70,18 @@ Our engineering organization is divided into mission based teams that contain th
 
 Start a discussion with your manager if you are interested to switch teams. Your manager will chat with you to understand what you want and then propose next steps. Team transfers need to be approved by the [VP of Engineering](roles.md#vp-engineering).
 
+### Communication
```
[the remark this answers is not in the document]
jeanduplessis: I agree and will move it.

PR #2410, opened 2021-01-26
Combining customer_issues.md content with handover

emchap:
https://github.com/sourcegraph/about/edit/main/handbook/ce/customer_issues.md seems duplicative of this document. Taking the content from that article and moving it here.

### DIFF
--- a/handbook/ce/ce_to_eng_handover.md
+++ b/handbook/ce/ce_to_eng_handover.md
@@ -1,6 +1,6 @@
 # CE-to-Engineering issue handover process
 
-When a prospect or customer files a new support ticket or raises a new issue, CE and Engineering should follow the steps below to ensure the issue is handled successfully.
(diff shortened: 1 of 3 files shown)


emchap:
@virginiaulrich These are the changes we'd discussed. Do these seem okay to you? They match what my understanding of the process is/should be, but the docs seemed to be pretty out of date so I wanted to make sure it was correct for how we want to handle it going forward. Basically as a new hire I was confused about where to ask questions in Slack that didn't warrant an Issue (as far as I could tell from peer behavior), so wanted to clarify that, add the requested info about Distribution's on-call process, and consolidate the two duplicative pages we discussed.

emchap:
cc @justdueck since you were also on the last one, if it's of interest.

virginiaulrich (approved):
Perfect! Thank you so much for updating it to reflect current reality <3

PR #2411, opened 2021-01-26
Remove inline link in 3.24 blog description

Joelkw:
Having an inline link (which I copied from our changelog entry and forgot to check over; my bad) breaks our render and leaves an empty section. 

This should fix that. 

(I think this specific issue might be worth calling out in whatever handbook documentation we make on release blog posts after product+marketing has had their release blog discussion. If that discussion is not going to happen before the next release, please let me know and I'll add some note to what we have instead around release blogposts now. Specifically, to catch this locally you must run `yarn serve` instead of `yarn run`.)


Joelkw:
(merged since it's a small change, a live/visible bug, and I tested locally)

PR #2422, opened 2021-01-27
ce: route alerts and panels based on owner

bobheadxi:
All _Observables_ (alerts and panels) have been attached to an owner for a while (since https://github.com/sourcegraph/sourcegraph/issues/12010), this information is now consistently available across both alert solutions and the dashboards reference. Questions about specific alerts and panels should now to directly go the owning team first

image


christinelovett (approved):
Thank you for adding @bobheadxi looks great 👍

davejrt (approved):

PR #2443, opened 2021-01-29
Read sourcegraphAnonymousUid cookie value into HubSpot forms

attfarhan:
Closes https://github.com/sourcegraph/sourcegraph/issues/11658. 

This PR populates the hidden anonymous user ID field in HubSpot forms with the cookie value in `sourcegraphAnonymousUid`. 

For any HubSpot form created using our `createHubSpotForm` function, when the form loads, we will read the `sourcegraphAnonymousUid` cookie value and populate the `anonymous_user_id` field in the form with the value. One thing to note is that this PR also changes the polyfill we currently have for jQuery. 

The polyfill is from this article, which outlines how to provide custom callbacks to embedded HubSpot forms without importing jQuery: https://www.unstack.com/blog/hubspot-on-form-submit-callbacks-without-jquery. The old polyfill was taken from https://community.hubspot.com/t5/APIs-Integrations/Form-callback-throws-unrelated-jquery-error/m-p/232127/highlight/true#M11023 AFAICT, and replacing it with the new code does not seem to have any side effects. This is generally frustrating as the HubSpot embedded forms have a hard dependency on jQuery, but this feels like a reasonable workaround.


website/gatsby-browser.js
```diff
@@ -13,11 +13,12 @@ exports.onInitialClientRender = function () {
...
-      window.jQuery ||
-      (() => ({
-        change: () => {},
-        trigger: () => {},
-      }))
+      window.jQuery || function(nodeOrSelector) {
+        if (typeof(nodeOrSelector) == 'string') {
+            return document.querySelector(s);
+        }
+        return nodeOrSelector;
```
felixfbecker, line 20: It would be good to document the hack here in a comment.

Also, do we not run Prettier and ESLint on this repo? This looks formatted differently and uses old function style 🤔

website/src/components/HubSpot.tsx
```diff
@@ -1,21 +1,43 @@
...
-        ; (window as any).hbspt.forms.create({
+        ;(window as any).hbspt.forms.create({
             portalId,
             formId,
             target: `#${targetId}`,
+            onFormSubmit,
+            onFormReady: (form: HTMLElement) => {
+                // The `form` parameter is a jQuery form object. We have a polyfill
+                // in gatsby-browser.js which allows us to access it as a DOM element
+                // instead of a jQuery object.
```
felixfbecker, line 27:
```suggestion
                // The `form` parameter is normally a jQuery wrapper around a form element.
                // We polyfill jQuery in gatsby-browser.js to return the form element directly.
```

website/src/components/HubSpot.tsx
```diff
@@ -1,21 +1,43 @@
...
     hubspot!.appendChild(script)
+    const anonymousId = cookies.get(SOURCEGRAPH_ANONYMOUS_ID_KEY)
     script.addEventListener('load', () => {
-        ; (window as any).hbspt.forms.create({
+        ;(window as any).hbspt.forms.create({
             portalId,
             formId,
             target: `#${targetId}`,
+            onFormSubmit,
+            onFormReady: (form: HTMLElement) => {
```
felixfbecker, line 24:
```suggestion
            onFormReady: (form: HTMLFormElement) => {
```

website/src/components/HubSpot.tsx
```diff
@@ -1,21 +1,43 @@
...
+                // The `form` parameter is a jQuery form object. We have a polyfill
+                // in gatsby-browser.js which allows us to access it as a DOM element
+                // instead of a jQuery object.
+
+                if (form) {
+                    // We want to populate hidden fields in the form with values stored in cookies when the form loads.
+                    const anonId: HTMLInputElement | null = form.querySelector('input[name="anonymous_user_id"]')
```
felixfbecker, line 31:
```suggestion
                    const anonId = form.querySelector<HTMLInputElement>('input[name="anonymous_user_id"]')
```

felixfbecker: There are two variables in scope here called `anonId` and `anonymousId`, which is a bit confusing. I would suggest renaming this one to `anonymousIdInput`.

felixfbecker (approved):

website/gatsby-browser.js
aileenrose: @felixfbecker it looks like we are 3.0.3 sourcegraph/prettierrc
 is this the latest? 
 should we switch over to gatsby's plugin: https://www.gatsbyjs.com/plugins/gatsby-plugin-prettier-eslint/ ?

aileenrose:
I get an Access Denied when trying to view: https://deploy-preview-2443--sourcegraph.netlify.app/

attfarhan:
Weird, on the previews linked from Netlify it works. I'll try re-building.

website/gatsby-browser.js
attfarhan: There are a lot of prettier updates, I'll do this in a separate PR.

attfarhan:
@aileenrose this is super odd -- I opened a PR with the same commits on a different branch name and the preview works fine: https://github.com/sourcegraph/about/pull/2450. It sounds like this has inexplicably happened to others in the past with no clear resolution: https://community.netlify.com/t/getting-access-denied-error/27753/21. I'm quite confident that this will work fine when deployed.

website/src/components/HubSpot.tsx
```diff
@@ -1,21 +1,41 @@
...
+            onFormSubmit,
+            onFormReady: (form: HTMLFormElement) => {
+                // The `form` parameter is normally a jQuery wrapper around a form element.
+                // We polyfill jQuery in gatsby-browser.js to return the form element directly.
+                if (form) {
+                    // We want to populate hidden fields in the form with values stored in cookies when the form loads.
+                    const anonymousIdInput = form.querySelector<HTMLInputElement>('input[name="anonymous_user_id"]')
+                    if (anonymousIdInput) {
+                        // Populate the hidden anonymous_user_id form field with the value from the sourcegraphAnonymousUid cookie.
+                        anonymousIdInput.value = anonymousId || ''
```
aileenrose, line 32: Should we check that if anonymousIdInput.value is not null, don't overwrite?  The Hubspot form will prefill values if it is a known user.

attfarhan: Yeah good call, updated

aileenrose (approved):
I was able to verify that the cookie value is read into the hidden input field.

attfarhan:
@felixfbecker @aileenrose I'm going to merge from the other PR I created, which has the same commits (https://github.com/sourcegraph/about/pull/2450). Just to reduce any potential risk that the deploy preview issues here somehow make it to prod. I'm not sure if/how Netlify rebuilds before deploying to prod, and the preview works on that PR, so it feels lower risk to do it from there.

PR #2488, opened 2021-02-03
Update eng org plan

nicksnyder:
I needed a unified view of the following things:
- Our current and future planned team organization
- Headcount and hiring plans

This change creates a single page for these things so it is easier for anyone to understand the hierarchy of the eng team and plan for it moving forward.

Tradeoffs
- I deleted references from individual team pages because I need a single source of truth. Trying to update leaf pages and "generate" the intermediate structures would have been a lot more work, and I suspect we will be moving to a different tool for this in the future anyway.
- It is not as clear the cross-functional relationships between product/design/engineering. I am open to feedback on how to improve this.

cc @sourcegraph/engineering-leadership


jeanduplessis (approved):
I understand the reason for pulling team member names off the team pages, but I don't like it.
I have an alternative suggestion, which should NOT hold up the merging of this PR:
- transition teammate info to a data file (json/yaml)
- create helpers to read from the file and display the relevant teammate data
- update all pages to use the helpers to display the teammate info

Having an SSOT data file also opens up opportunities to use that data in other areas where automation can be beneficial.

As an example see GitLab's team.yml file: https://gitlab.com/gitlab-com/www-gitlab-com/-/blob/master/data/team.yml?expanded=true&viewer=simple

handbook/engineering/eng_org.md
```diff
@@ -0,0 +1,109 @@
...
+
+If you are already at Sourcegraph and see a current or future opportunity that you're interested in, please tell your manager. Your manager will chat with you to understand what you want and then propose next steps. Team transfers need to be approved by the [VP of Engineering](roles.md#vp-engineering).
+
+## Organization
+
+- [Beyang Liu](../../../company/team/index.md#beyang-liu), CTO
+- [Nick Snyder](../../../company/team/index.md#nick-snyder-he-him), [VP Engineering](roles.md#vp-engineering)
+    - Org: Search and replace (WIP name)
+        - FQ3 Engineering director, Nick acting manager until then.
+        - Team: Search platform and infrastructure (currently [Search](search/index.md)) {#search-platform-and-infrastructure}
```
nicksnyder, line 19:
```suggestion
        - Team: Search core backend (currently [Search](search/index.md)) {#search-core-backend}
```

lguychard (approved):

chrispine:
While I don't strongly object, I am with Jean: I don't love this. I like having a SSOT, but I'd rather we slurp that truth, rather than link to it. (If that means the team pages slurp from the org chart, so be it.)

On the other hand, if this is temporary, my opinions become a lot less strong.

nicksnyder:
Thanks for the feedback. I have prototyped a different way that keeps the source of truth for each team in their own page, but still allows a single unified view here: https://github.com/sourcegraph/about/pull/2507

Please add comments on the new approach there, I am going to close this.

PR #2490, opened 2021-02-03
Update values

mercadon:
Replace "be a good human" with "be welcoming and inclusive"


virginiaulrich (approved):

virginiaulrich:
Approved, though do wonder if it should be "be welcoming and inclusive ..." Not sure which is easier to read, so that doesn't change my review, but will comment in case anyone else has a stronger sense of which reads better

dadlerj (approved):

dadlerj:
Thanks for the thoughtful wording and structure here!

nicksnyder (approved):
Love it, and I think we can/should link directly to this from our job descriptions.

sqs (approved):

mercadon:
All credit goes to @virginiaulrich on the thoughtful update and rewording of this value! 🥇

jeanduplessis (approved):
Really happy to see this update ❤️

christinaforney (approved):

PR #2503, opened 2021-02-04
Document how we're using OKRs for FY22 goals

sqs:
We are setting goals for FY22 and Q1 using the OKR format. I updated the handbook page about OKRs with some information about OKRs, guidelines on how to write good OKRs, and some history of goal-setting.


virginiaulrich:
Just want to say as someone who has consulted dozens of companies on OKRs, all of this is a great start (well ahead of what I would have expected). Happy to talk OKRs with anyone who wants to :)

PR #2505, opened 2021-02-04
team: fix malformed GeoJSON

LawnGnome:


jeanduplessis:
Thanks for the fix @LawnGnome - didn't notice I broke it 😞

PR #2515, opened 2021-02-08
Update index.md

virginiaulrich:
@InesRoitman I imagine a lot of this will change in the future, but for now it seemed right to link to the page I am working on for our Customer Support Engineers


InesRoitman (approved):
This looks good. I agree there's going to be changes as we'll build an onboarding together. I'll show you what the general onboarding I send looks like and the general template I've created for team onboardings, that I think will help.
Thanks for this

PR #2522, opened 2021-02-10
Fix backend engineer JD link

lguychard:
See https://sourcegraph.slack.com/archives/C168VP0FM/p1612951232315800


felixfbecker:
I also just fixed this on master so I think this can be closed, thank you though!

PR #2542, opened 2021-02-15
Update desired outcome for Distribution goals

pecigonzalo:
We have decided to switch to a different scoring system instead of 1-10. We will use:
- Strong Negative
- Negative
- Positive
- Strong Positive

The reason for doing this is that its easier to set a desired target at the moment given we have no existing score so setting a target of 9+ could be impossible to achieve and setting a % based increases could be really misleading (If our current score is 1 and we move to a 2, that is a 100% increase but its still a really bad score).

### DIFF
--- a/handbook/engineering/distribution/goals.md
+++ b/handbook/engineering/distribution/goals.md
@@ -25,7 +25,7 @@ Our existing deployment pipelines to our Sourcegraph instances (such as Sourcegr
 
 - **Status**: Not started. _Estimated: FY22-Q1_
 - **Outcomes**:
-  - Our deployment pipeline earns a 9+ score from Sourcegraph engineering team.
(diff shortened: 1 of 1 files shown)


handbook/engineering/distribution/goals.md
```diff
@@ -25,7 +25,7 @@ Our existing deployment pipelines to our Sourcegraph instances (such as Sourcegr
 
 - **Status**: Not started. _Estimated: FY22-Q1_
 - **Outcomes**:
-  - Our deployment pipeline earns a 9+ score from Sourcegraph engineering team.
+  - Our deployment pipeline earns a possitive score from Sourcegraph engineering team.
```
uwedeportivo, line 28:
```suggestion
  - Our deployment pipeline earns a positive score from Sourcegraph engineering team.
```

uwedeportivo (approved):

PR #2546, opened 2021-02-15
Adding insiders build info

emchap:
So that it's not lost in Slack, adding info on how to move off the insiders build and how it winds up happening.


justdueck (approved):
LGTM. Thanks, @emchap!

PR #2561, opened 2021-02-17
Add 'aim to improve dev experience' to campaigns working agreements

mrnugget:
This is also an outcome of yesterday's retrospective in which we said (okay: I said a lot) that we should try to fight back against "this has already been bad"-inertia and try to actively improve our developer experience.

Now, I didn't know how to formulate this here, which is why it might sound a bit blurry. Suggestions welcome.

Alternative proposal:

> We try to fight back against "this has already been bad"-inertia and work to improve our developer experience


eseliger (approved):

malomarrec (approved):

LawnGnome (approved):
I like this framing! :+1:

chrispine (approved):
LGTM, thanks!

Issue #2573, opened 2021-02-19
Tooltips on Handbook pages are broken


aileenrose:
Just added Bootstrap and this page contains tooltips:
https://about.sourcegraph.com/workflow

marekweb:
This has been fixed.

PR #2582, opened 2021-02-19
Updated "Other Tools" table styling so text is visible

mike-r-mclaughlin:
Text was hidden due to a conflicting style named "tooltip" defined in bootstrap.min.css.  I don't see boostrap included when I run the handbook site locally.  So, it seems unlikely that this .tooltip style was intended to come from bootstrap.

### DIFF
--- a/workflow/index.md
+++ b/workflow/index.md
@@ -57,11 +57,11 @@ See product comparisons and integrations with:
     .gray {
         background-color: #222;
     }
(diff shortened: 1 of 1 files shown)


aileenrose (approved):

aileenrose:
also addressed this: https://github.com/sourcegraph/about/issues/2573

PR #2599, opened 2021-02-22
Clarify that releases on weekend are up to release captain 🚢 

daxmc99:
Give more power to the release captain to move the monthly release date to account for holidays / non-working days.

### DIFF
--- a/handbook/engineering/releases/index.md
+++ b/handbook/engineering/releases/index.md
@@ -127,3 +127,7 @@ This is because:
 - We haven't built the automated testing and update infrastructure to make continuous customer releases reliable.
 
 In the future, we may introduce continuous releases if these issues become surmountable.
+
+### Weekend releases
+
+If the 20th falls on a non-working day, the release captain has the right but not the obligation to change the day (+/- 3 days) the release images are published. 
--- a/handbook/engineering/releases/index.md
+++ b/handbook/engineering/releases/index.md
@@ -130,5 +130,5 @@ In the future, we may introduce continuous releases if these issues become surmo
 
 ### Weekend releases
 
-If the 20th falls on a non-working day, the release captain has the right but not the obligation to change the day (+/- 3 days) the release images are published.
+If the 20th falls on a non-working day, the release captain has the right but not the obligation to change the day (+/- 3 days) that the release images are published.
 If they decide to do this, they should move the google calendar events and announce the change in the #dev-announce channel.
--- a/handbook/engineering/releases/index.md
+++ b/handbook/engineering/releases/index.md
@@ -8,7 +8,7 @@ This document describes how we release Sourcegraph.
 
 **Sourcegraph releases are monthly.**
 A *release* refers to a minor version increase of Sourcegraph (e.g. `3.0.0` -> `3.1.0`).
-We create releases by 10am US Pacific Time on the 20th day of each month. ([why?](#why-the-20th))
+We aim to create releases by 10am US Pacific Time on the 20th day of each month. ([why?](#why-the-20th))
 
 These releases **may** require [manual migration steps](https://docs.sourcegraph.com/admin/updates).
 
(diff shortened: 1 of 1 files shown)


pecigonzalo:
Sounds great! I would share in #dev-chat and @ someone from CE and Product to review here.

emchap:
I don't see any issue with this, though I think long-term we might want to standardize on a date (like, always the first working day back on the 20th or something). From a process level seems like a totally reasonable work/life balance change.

Joelkw:
I am not super familiar with our site admin personas or relationship on how distribution works, but from product I'd frame this question around the user. Some quick example scenarios:

- if the people in charge of updating their instance wait for a release to drop explicitly / a slack ping, it probably doesn't matter when we release since we manually communicate it. 
- if those stakeholders like to schedule something regularly into their calendar to remind themselves when to update, it might even be more helpful for them if we explicitly set our release cadence to the 7-day week rather than a month-date? (Like, every 4th Tuesday or whatever; this way it's never on a weekend anyway) 
- if those stakeholders just generally update "after the 20th" then it might make most sense to *only* shift this day forward rather than forward *or* back, so the 20th is always the latest possible day and is easy to remember? 

Are any of those situations more/less applicable to this? 

I'm not worried about this change impacting internal product/eng teams' planning; no one depends on having exactly 1-month between releases.

handbook/engineering/releases/index.md
```diff
@@ -127,3 +127,8 @@ This is because:
 - We haven't built the automated testing and update infrastructure to make continuous customer releases reliable.
 
 In the future, we may introduce continuous releases if these issues become surmountable.
+
+### Weekend releases
+
+If the 20th falls on a non-working day, the release captain has the right but not the obligation to change the day (+/- 3 days) that the release images are published.
+If they decide to do this, they should move the google calendar events and announce the change in the #dev-announce channel.
```
Joelkw, line 134: Maybe add some minimum time this announcement must be made just for clarity? Like 2 weeks ahead?

emchap:
Of the paying customers that are on 3.25.0 now, we have: 

- Tradera at Paypal on Team
- Tinder (we manage them)
- Prezi
- Fiverr
- Qualtrics

So it feels like a relatively small customer impact. I like the idea of keeping it consistent on a day (fourth Thursday or whatever) because it allows us to always avoid Friday releases if we want to and can be scheduled into calendar software more easily.

virginiaulrich (approved):
I think this is a very good and healthy change!

handbook/engineering/releases/index.md
```diff
@@ -127,3 +127,8 @@ This is because:
 - We haven't built the automated testing and update infrastructure to make continuous customer releases reliable.
 
 In the future, we may introduce continuous releases if these issues become surmountable.
+
+### Weekend releases
+
+If the 20th falls on a non-working day, the release captain has the right but not the obligation to change the day (+/- 3 days) that the release images are published.
```
christinaforney, line 133: Why +/- 3 and not +/- 2? and/or why not the previous Friday or following Monday? Just curious to hear understand the thought process  :) (I do not feel strongly about this and support as written).

christinaforney (approved):
Agree with @emchap about standardizing this shift for more consistency, but don't feel like we need to do that at this time. +1 to work/life balance!

pecigonzalo (approved):

pecigonzalo:
How about moving to `-` days, and guarantee that the release is going to be there on or by the 20th? Similar to what @emchap suggested.

handbook/engineering/releases/index.md
pecigonzalo: We no longer announce releases in advance as part of the [single-day release](https://github.com/orgs/sourcegraph/projects/90) project.

If we guarantee that by the 20th the release is there, would there be a need for an announcement? Basically, if 20th falls on the weekend/holiday, we cut the last working day before that, so by the 20th, it's there and the blog can be sent the next working day after the 20th.

handbook/engineering/releases/index.md
daxmc99: My thought here was if the release was on a Monday holiday after the weekend we could release on the Thursday / Friday before.

daxmc99:
Kept the wording that we release by the 20th and allowed the release captain to move the day up to 3 days **before** the 20th to ensure we always release by then. 
Planning to merge but I'm happy to reopen if new concerns crop up.

slimsag (approved):
LGTM

PR #2600, opened 2021-02-22
Clarifying current vs. future state

emchap:
Per https://sourcegraph.slack.com/archives/CHPC7UX16/p1613994761006200?thread_ts=1613061743.251900&cid=CHPC7UX16

I'm clarifying this because it's caused confusion for me twice—once when asking an AE for an enterprise version for a key (confusing them) and once because a customer is on an old Starter license and what that meant was really unclear. Are we okay keeping it split like this until we launch the future state?


unknwon (approved):
Thank you!

PR #2619, opened 2021-02-23
Refresh campaigns goals

malomarrec:


malomarrec:
We need to add milestones leading to each outcome.
Since we are still exploring some the options leading to outcomes, I prefer making the outcomes clear, and set time later to define the milestones / features.

handbook/engineering/campaigns/goals.md
```diff
@@ -65,3 +56,11 @@ We maintain a [directionally correct roadmap](https://sourcegraph.productboard.c
...
 
 *Updated 2021-02-12*
+
+# How we built those goals and milestones
+We generally think about campaigns features in terms of 4 themes:
+
+- expand the realm where campaigns can run
+- create many changesets easily
+- manage and track many changesets easily
+- get changesets merged faster
```
chrispine, line 66: At least as we've talked about these in the past, we usually combined "manage" and "merge" as one item, and included as a fourth item "performance and robustness", so something like:

```suggestion
1. create many changesets easily
1. track, manage, and merge many changesets easily
1. expand the realm where campaigns can run
1. keep campaigns robust and performant
```

chrispine (approved):
Thanks for doing this! Even though we don't have the exact numbers yet (N1, N2, etc), I'd suggest merging what's here and updating with the precise numbers later.

PR #2641, opened 2021-02-26
Update support-team-rituals.md to add CS team calendar note

northyg:


virginiaulrich (approved):
Thank you for updating this! I had already forgotten about making the calendar and doing so

PR #2666, opened 2021-03-01
Fixing a couple of broken links and one typo

mike-r-mclaughlin:
Broken Links:
* "Observable":
    from `repo:^github\.com/sourcegraph/sourcegraph$ file:generator.go type Observable struct {:[def]} patternType:structural` (produces `No Results`)
    to     `repo:^github\.com/sourcegraph/sourcegraph$ type Observable struct {:[def]} patternType:structural` (produces 1 result in `monitoring.go`)
* "Prometheus alerting rules" => "#sourcegraph-prometheus" (from #alerting which does not exist on this page)

Typo:
* "(we did not have alertmanager at the time either.)  There" to "(we did not have alertmanager at the time either).  There"


slimsag (approved):

slimsag:
@pecigonzalo we should update CODEOWNERS so someone else on Distribution gets notified for PRs like this. (I'd still like to be notified)

PR #2672, opened 2021-03-02
Unlist code insights frontend role

felixfbecker:
With the new headcount plans I can't hire a second frontend engineer until later this year.


felixfbecker:
Done on main by @withdavidli

PR #2675, opened 2021-03-02
add Stompy to team page

airamare01:
adding Stompy to team page


virginiaulrich (approved):
It's official, official!

PR #2679, opened 2021-03-02
adding beatrix and stompy to ce page

airamare01:
Making edits to reflect new merges made on team page including beatrix and stompy in the ce team under cse(s)


virginiaulrich (changes requested):
Aaw, thank you for doing this! I already have this as part of the changes I mentioned here https://sourcegraph.slack.com/archives/C01JR51JR5J/p1614707461019200. Since you are done for the day, I'll push them via my change and you can cancel this PR tomorrow. If you were still online, I'd do that the other way around. Cheers to it being official official!

PR #2683, opened 2021-03-03
Updating info about demodotcom's state

emchap:
I believe we're at a point where we can do all demos on that instance unless we need code that's not loaded in. Putting to the team for review.


christinelovett (approved):
LGTM

PR #2693, opened 2021-03-04
handbook: make it clear that RFCs should have a single editor

pdubroy:
Per discussion on Slack: https://sourcegraph.slack.com/archives/C02FSM7DU/p1614856694370100

Also fixed up some formatting issues:

Screenshot 2021-03-04 at 20 48 06
Screenshot 2021-03-04 at 20 47 58
Screenshot 2021-03-04 at 20 47 50

### DIFF
--- a/handbook/communication/rfcs/index.md
+++ b/handbook/communication/rfcs/index.md
@@ -75,9 +75,13 @@ Effective RFCs contain the following information:
 - Title that includes the RFC number.
   - The title is inlined in the Google Doc so that it is more visible and will not disappear if exported to a different format.
 - Metadata about the state of the RFC. Including but not limited to:
-  - **Editor:** A single person responsible for iterating on the content of the RFC.
-     - Generally, this is the author of the Google Doc but it is possible for ownership to be transferred.
-     - If there are multiple authors, they can be listed separately (e.g. under _Authors_), but a single person should be chosen and listed as the editor.
+  - **Editor:** The single person who is directly responsible for the content in the RFC.
+     - The editor ensures that:
+        - The RFC is well written.
+        - The right people review the RFC.
+        - Comments and feedback are addressed.
+        - The RFC progresses toward a resolution.
+     - The editor frequently authors most of the document, but that is not a strict requirement. They may delegate responsibilities to other teammates.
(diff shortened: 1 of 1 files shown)


handbook/communication/rfcs/index.md
```diff
@@ -75,15 +75,16 @@ Effective RFCs contain the following information:
 - Title that includes the RFC number.
   - The title is inlined in the Google Doc so that it is more visible and will not disappear if exported to a different format.
 - Metadata about the state of the RFC. Including but not limited to:
-  - **Editor:** The person responsible for iterating on the content of the RFC.
-    - Generally, this is the author of the Google Doc but it is possible for ownership to be transferred.
+  - **Editor:** A single person responsible for iterating on the content of the RFC.
+     - Generally, this is the author of the Google Doc but it is possible for ownership to be transferred.
+     - If there are multiple authors, they can be listed separately (e.g. under _Authors_), but a single person should be chosen and listed as the editor.
```
nicksnyder, line 80: What problem does this solve that the Google Docs edit history doesn’t?

pdubroy: In certain cases it's useful to signal that more than one person should feel ownership over the contents. E.g., a discussion in a meeting moves into an RFC — the PM or EM might start the RFC (and be the editor) but the team really owns the content.

nicksnyder, line 80: My feedback on the current proposal is that "editor" and "author" sound like two different words for the same thing, so I worry it would result in confusion.

We could think of a different set of words to address that, or maybe clearer documentation here on the definition of what an editor is would be sufficient.

Below is my proposal, but I will leave it up to you to decide.

```suggestion
  - **Editor:** The single person who is directly responsible for the content in the RFC.
     - The editor ensures that:
        - The RFC is well written.
        - The right people review the RFC.
        - Comments and feedback are addressed.
        - The RFC progresses toward a resolution.
     - The editor frequently authors most of the document, but that is not a strict requirement. They may delegate responsibilities to other teammates.
```

nicksnyder (approved):
I made a suggestio

handbook/communication/rfcs/index.md
pdubroy: Sounds good to me.

PR #2710, opened 2021-03-06
Revert "Add Beatrix Woo's city location"

abeatrix:
Reverts sourcegraph/about#2703


abeatrix (commented):
Revert because of the failed build

PR #2729, opened 2021-03-09
updated blog posts for the rename from Campaigns to Batch Changes

chrispine:
This updates our blog posts for the campaigns/batch-changes rename.

**Notes:**

- We are intentionally _not_ updating changelogs and press releases.
- For the "find and replace Odyssey" post, since the content was out of date anyway, I just added a note to see the docs for the latest.
- I will come back and update the docs URLs when the docs have been updated.


content-hacker:
Thanks for making the changes!

PR #2732, opened 2021-03-09
updated CE and Marketing pages for the rename: Campaigns -> Batch Changes

chrispine:
This updates the CE and Marketing pages in our handbook to refer to Batch Changes instead of Campaigns.

_I am not confident that these changes are all correct, especially those relating to license keys._ So please double-check those.

(Finally, my editor trims trailing whitespace, so you might want to append `?w=1` to the url when viewing the diffs to ignore those trivial changes.)


virginiaulrich (approved):
I don't see any more, but we also do a monthly audit of our pages, so if there are any left, we will catch them soon. And thanks for fixing all of my white space mistakes ... I only learned recently that is not what you're supposed to do and hadn't yet gone back through to fix it all.

chrispine:
@justdueck Sadly, no, I had to consider them case-by-case, since "campaign" could also refer to a marketing campaign, email campaign, etc. 😄

PR #2745, opened 2021-03-10
Final[*] "campaign" -> "batch change" renaming for the About pages

chrispine:
This is the last[*] of the changes to the About pages for the "campaign" -> "batch change" rename, which only consists of our team pages.

[*] I will still need to do a final pass over all of the About pages to update docs links, once the docs changes have gone live.


chrispine:
People are noticing that these are in an incomplete state, so I'm going to merge these (and start on updating the links now that the docs are live). But still feel free to leave comments/feedback.

PR #2749, opened 2021-03-11
Add note about spending the home office budget later after joining

arussellsaw:
This was asked & answered in Slack, so just persisting the answer here https://sourcegraph.slack.com/archives/CQAGQKC4A/p1615390116094000


pdubroy (approved):
lgtm, thanks!

PR #2750, opened 2021-03-11
Update capitalization and plural/singular treatment

abschu:
Marketing is in the process of updating our writing guidelines. While we have this sorted for Batch Changes we will need to evaluate the implications for all other products and features. This update is to add clarity specifically for Batch Changes.

### DIFF
--- a/handbook/communication/content_guidelines/style_and_mechanics.md
+++ b/handbook/communication/content_guidelines/style_and_mechanics.md
@@ -722,7 +722,12 @@ Only use _we_ and _our_ (as in “our GitHub integration”) in informal documen
 
 ### Feature names
 
-Don’t capitalize descriptive product or feature names. Features are not proper nouns. Avoid phrasing that can imply official product names.
+When referencing the product or category, the following should be treated as proper, singular nouns:
+
+- Universal Code Search
+- Batch Changes
+
+Don’t capitalize descriptive product or feature names. Avoid phrasing that can imply official product names.
 
 <div class="usage">
 <div class="item yes">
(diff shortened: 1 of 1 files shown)


handbook/communication/content_guidelines/style_and_mechanics.md
```diff
@@ -722,7 +722,12 @@ Only use _we_ and _our_ (as in “our GitHub integration”) in informal documen
 
 ### Feature names
 
-Don’t capitalize descriptive product or feature names. Features are not proper nouns. Avoid phrasing that can imply official product names.
+When referencing the product or category, the following should be treated as proper, singular nouns:
+
+- Universal Code Search
+- Batch Changes
```
felixfbecker, line 728: I don't like the idea that we're making this different case-by-case. It now says that "descriptive" product or feature names should not be capitalized, but it's not clear to me how "batch changes" is less "descriptive" than "code insights", "code monitoring", "code intelligence", and our other features? I'd really prefer to know that I can either refer to both "Batch Changes" and "Code Insights" or to both "batch changes" and "code insights" the same way, respectively.

handbook/communication/content_guidelines/style_and_mechanics.md
```diff
@@ -743,19 +748,19 @@ Don’t capitalize descriptive product or feature names. Features are not proper
 </div>
 </div>
 
-Use the natural plural/singular form. If a feature name is a plural noun, treat it as a plural noun.
+Use the natural plural/singular form when describing the functionality.
```
felixfbecker, line 751: Do you have an example sentence in which "Batch Changes" would describe the _functionality_  and be plural?

handbook/communication/content_guidelines/style_and_mechanics.md
abschu: @felixfbecker - agreed. This was meant as a short term solution to support the Batch Changes launch. Marketing will eventually be updating our style guide for all products and features, but we need to get alignment with product and engineering first. At this point, we wanted to start with Batch Changes since there was a hard deadline.

handbook/communication/content_guidelines/style_and_mechanics.md
abschu: "The batch changes I created yesterday are already complete."

PR #2754, opened 2021-03-11
capitalize Batch Changes

chrispine:
This fixes capitalization inconsistencies where we were not using our current guidance.


chrispine:
Note that I did _not_ change "Batch changes" on the pricing page, since that's very customer-visible and is presented next to a bunch of other features which don't follow the same capitalization rules; I was concerned that it might look weird.

PR #2766, opened 2021-03-15
Removed weekly CE account strategy check-in line.

northyg:


virginiaulrich (commented):
Thank you for changing this! Would it be okay if you also updated the https://about.sourcegraph.com/handbook/ce/support-team-rituals doc as part of this update to remove the mention of this meeting there, too?

northyg:
@virginiaulrich Good call out! I missed that one. I will remove that reference as well :)

virginiaulrich:
Thanks, @northyg! I'll mark this as approved then. Happy to review again if you want me to, otherwise, looks like you're all set!

virginiaulrich (approved):

PR #2771, opened 2021-03-15
Add process docs for Customer Environment questions

daxmc99:
Part of https://github.com/sourcegraph/customer/issues/236

I felt like Bizops was the best place for this since BigQuery is the long-term data store for this.

### DIFF
--- a/handbook/ops/bizops/customer_environment_questions.md
+++ b/handbook/ops/bizops/customer_environment_questions.md
@@ -14,7 +14,8 @@ The [form](https://docs.google.com/forms/d/1GXDrbVTnKf9rbaY-0tYSy3BsufKXfQZJf7Gb
 is for internal use only. A Sourcegraph employee must fill it out. It is not
 required that the entire form be completed everytime we have a customer interaction
 but rather to provide a baseline knowledge of any considerations that need to be
-made when interacting with a customer.
+made when interacting with a customer. To make this form more discoverable, a 
+link to it is included in the [customer issue template](https://github.com/sourcegraph/customer/blob/master/.github/ISSUE_TEMPLATE/customer-issue.md).
 
 ### Data Pipeline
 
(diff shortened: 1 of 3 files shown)


virginiaulrich (commented):
For the support workflow, this covers what I was imagining for in terms of documentation. Thank you @daxmc99! As for what is ideal to have in Salesforce ... honestly: all the questions. But if for some reason we need to pick, from a support perspective, at least these:

- What are the customer's preferred tools and workflows to deploy off-the-shelf Kubernetes services like Sourcegraph?
- Does the customer utilize a custom registry?
- Does the customer currently modify the deploy-sourcegraph repo base configuration for their environment? 
- Do you run Sourcegraph as the only application in your cluster (single-tenant) or do you run Sourcegraph in a shared cluster with other applications (multi-tenant)?
- Is there anything we should know about the customer deployment environment? (Security requirements, CD pipeline configuration)
- Does the customer currently use a managed Kubernetes cluster? If so, what cloud provider do they use?

I'll let the CEs and/or Aimee/Tia speak to what the CE half of the team desire, since I know that we all use the information slightly differently.

handbook/ops/bizops/customer_environment_questions.md
```diff
@@ -0,0 +1,52 @@
...
+### Data Pipeline
+
+The pipeline for customer data is below
+
+![Image of customer data pipeline](https://sourcegraphstatic.com/customer_env_pipeline.png)
+
+
+###  Looker 
+
+We use Looker to view the aggregate results. This allows teams that use this data
```
ebrodymoore, line 29:
```suggestion
We use BigQuery to [transform](https://console.cloud.google.com/bigquery/scheduled-queries/locations/us/configs/615bc536-0000-2e57-bbd9-f40304364c86/runs?project=telligentsourcegraph) the results and Looker to [view them](https://sourcegraph.looker.com/dashboards-next/170). This allows teams that use this data. 
```

ebrodymoore (approved):
Good stuff!

tiawucher:
@daxmc99 This looks great! Let's give this a whirl and if we need do not see the engagement we expect we can iterate, but I think this a great start! Thank you for putting this together!

PR #2774, opened 2021-03-16
Add Focus Time for Beatrix Woo

abeatrix:
Happy to change to another day/time if someone prefers Wednesday AM /  Thursday AM more.


virginiaulrich (approved):
Same note for you that I left with Warren -- yay! And also, very glad folks are picking times they want. If we run into any issues, we can discuss, but first and foremost, it's good to see what you want and we will try to make that work first!

PR #2775, opened 2021-03-16
Update onboarding checklist tool in CS onboarding

InesRoitman:


virginiaulrich (approved):
Thanks so much for updating this, Inés! We have a fair amount of updates to make next month after we retro the onboarding experience and I am pretty sure we would have missed this one :)

PR #2779, opened 2021-03-16
Clarify product positioning

malomarrec:
* Clarify use cases
* Add Ship breaking changes use case
* Break down Discovery guide per user category
* Add interview guide links


abschu (approved):
Looks great!

PR #2796, opened 2021-03-18
Clarified priority tags in engaging-other-teams.md

northyg:


virginiaulrich (approved):
Thank you so much again for catching this, raising it to the group, and the follow through to reflect reality in the handbook!

PR #2801, opened 2021-03-19
Add more detailed compensation documentation + philosophy

dadlerj:


handbook/people-ops/compensation/equity-faq.md
```diff
@@ -0,0 +1,69 @@
+# Sourcegraph options and equity compensation FAQ
+
+## Sourcegraph options
+
+Sourcegraph currently only provides equity to employees in the form of options.
```
nicksnyder, line 5: Since we differentiate between employees and contractors below, and since both receive options, I think it is clearer to use "teammates" here.

```suggestion
Sourcegraph currently only provides equity to teammates in the form of options.
```

handbook/people-ops/compensation/equity-faq.md
```diff
@@ -0,0 +1,69 @@
...
+
+### How are option grants calculated for promotions?
+
+As a company grows in size, the value of each share (and option) grows, which means that the _number of shares_ granted will decline over time.
+
+This means that simply joining the company earlier can result in someone having more options than the latest bands would suggest. And if they are promoted, they could already have more options than the new, theoretically higher band would reflect.
+
+In such cases, teammates are not intended to be punished for joining earlier. If necessary, they will generally receive an additional number of options that matches the _increase_ between their previous and new levels in the latest bands. Some variation can exist in certain situations.
+
+As an example of how this could work in practice, imagine a teammate already has 1,000 options. The company's bands have evolved since that iniital grant, and they are later promoted from their current level, which has a new target of 300 options, to a higher level that has a target of 425 options. For their promotion, this individual would receive +125 options (the difference between 425 and 300) to get them to 1,125 total.
```
Strum355, line 53:
```suggestion
As an example of how this could work in practice, imagine a teammate already has 1,000 options. The company's bands have evolved since that initial grant, and they are later promoted from their current level, which has a new target of 300 options, to a higher level that has a target of 425 options. For their promotion, this individual would receive +125 options (the difference between 425 and 300) to get them to 1,125 total.
```

nicksnyder (approved):
Great docs, thanks!

handbook/people-ops/compensation/equity-faq.md
dadlerj: Oops thank you for catching this! Unlike most pages, this one actually intentionally uses a mix of those terms, but this should be teammates :)

PR #2814, opened 2021-03-23
Document our accessibility standards

umpox:
Relevant RFC: https://docs.google.com/document/d/1vl6Bt6VukyGhJSrLon2IVG5flX35QpeqqKsJQXH_mGU/edit#

### DIFF
--- a/handbook/product/design/design-and-interaction-guidelines.md
+++ b/handbook/product/design/design-and-interaction-guidelines.md
@@ -70,7 +70,7 @@ In these cases, we can utilize web application styles inside the search context 
 ## Accessibility standards
 People of all abilities should have first-class access to code and coding. In order to achieve this it is vital that our applications are designed and built to be accessible to all.
 
-Our criterion is to meet [WCAG Level AA Conformance](https://www.w3.org/WAI/WCAG2AA-Conformance). Use [this reference](https://www.w3.org/WAI/WCAG21/quickref/?currentsidebar=%23col_customize&levels=aaa&technologies=smil%2Cflash%2Csl) to get a better understanding of what this involves.
+We aim for [WCAG Level AA Conformance](https://www.w3.org/WAI/WCAG2AA-Conformance) in all areas. Use [this reference](https://www.w3.org/WAI/WCAG21/quickref/?currentsidebar=%23col_customize&levels=aaa&technologies=smil%2Cflash%2Csl) to get a better understanding of what this involves.
 
 ### Todo:
 
--- a/handbook/product/design/design-and-interaction-guidelines.md
+++ b/handbook/product/design/design-and-interaction-guidelines.md
@@ -68,7 +68,7 @@ Drawing the line between what is a code page and what is a web application page 
 In these cases, we can utilize web application styles inside the search context to give links greater importance through color and provide additional margin and padding to aid readability.
 
 ## Accessibility standards
-Our mission is to make it so everyone can code. In order to achieve this it is vital that our applications are designed and built to be accessible to all.
+People of all abilities should have first-class access to code and coding. In order to achieve this it is vital that our applications are designed and built to be accessible to all.
 
 We aim for [WCAG Level AA Conformance](https://www.w3.org/WAI/WCAG2AA-Conformance), use [this reference](https://www.w3.org/WAI/WCAG21/quickref/?currentsidebar=%23col_customize&levels=aaa&technologies=smil%2Cflash%2Csl) to get a better understanding of what this involves.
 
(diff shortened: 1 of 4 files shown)


handbook/product/design/design-and-interaction-guidelines.md
```diff
@@ -67,19 +67,18 @@ Drawing the line between what is a code page and what is a web application page
...
+
 ### Todo:
 
 * Displaying state in buttons
   * Default to the current state for the icon so that the user understands what is happening now
   * The button text should describe the action that will occur on click
 * Developer friendly as a first principle
   * Keyboard navigation is a priority
-* Accessibility
-  * People of all abilities should have first-class access to code and coding. We cannot fulfill that mission if our application is not accessible
```
felixfbecker, line 78: I liked this sentence "People of all abilities should have first-class access to code and coding" more than "Our mission is to make it so everyone can code" because it is more closely related to what our product enables today (access to code), as opposed to only what our product might enable in the future (access to coding).

I think the high-level goal and design principle is even more important than a WCAG standard, because [WCAG standards can be flawed at times](https://github.com/w3c/wcag/issues/695) too or sometimes it's easy to "nominally" pass the standard without providing the best possible user experience, and for those case it's important to think from first principles.

umpox: Great point, will preserve the previous text! I agree about the standard, it's not perfect in areas but it's a very good level to aim for 👍

handbook/product/design/design-and-interaction-guidelines.md
```diff
@@ -67,19 +67,18 @@ Drawing the line between what is a code page and what is a web application page
 
 In these cases, we can utilize web application styles inside the search context to give links greater importance through color and provide additional margin and padding to aid readability.
 
+## Accessibility standards
+People of all abilities should have first-class access to code and coding. In order to achieve this it is vital that our applications are designed and built to be accessible to all.
+
+We aim for [WCAG Level AA Conformance](https://www.w3.org/WAI/WCAG2AA-Conformance), use [this reference](https://www.w3.org/WAI/WCAG21/quickref/?currentsidebar=%23col_customize&levels=aaa&technologies=smil%2Cflash%2Csl) to get a better understanding of what this involves.
```
quinnkeast, line 73: My only comment here is that "aim for" allows us to not reach this level of conformance. Maybe something like:

```suggestion
Our criteria is to meet [WCAG Level AA Conformance](https://www.w3.org/WAI/WCAG2AA-Conformance). Use [this reference](https://www.w3.org/WAI/WCAG21/quickref/?currentsidebar=%23col_customize&levels=aaa&technologies=smil%2Cflash%2Csl) to get a better understanding of what this involves.
```

quinnkeast (approved):

handbook/product/design/design-and-interaction-guidelines.md
felixfbecker: @quinnkeast curious if in any case WCAG AA conformance ever contradicted what you believed to be the best user experience for a user with accessibility needs (e.g. see  https://github.com/w3c/wcag/issues/695), what would you want to do?

"aim" allows us to do what we believe is best, which may be good unless we have a _mandate_ to meet this standard to e.g. sell our product to a government agency (afaik we don't yet).

umpox: It's weighing this up with some areas where the standard isn't always perfect as Felix mentioned. I find your version much more significant though and I think for the vast majority of cases that's exactly what we want so will update!

umpox: @felixfbecker I agree but I'd be curious to understand if we have any examples where we actually haven't been able to adhere to these standards (e.g. where your example applies). It seems like it is best (at least here) to be as clear as possible.

umpox: OK, after better understanding the implications of that kind of issue I'm flip-flopping on this. 

How about:
> We aim for WCAG Level AA Conformance in all areas. Use this reference to get a better understanding of what this involves.

Worded slightly differently for significance. I think ultimately we understand what we need to achieve the actual wording here should have minimal difference unless we are legally required to declare this

quinnkeast: I'm also flip-flopping, because of @felixfbecker's example, which I strongly agree is a big problem with the guidelines using the guidelines a rote reference. The missing link for me is in how we convey "accessibility is not optional and isn't scope that can be cut in future work."

umpox:
@quinnkeast In the interest of moving forward, I'm going to merge this but happy to keep the discussion going! 

For me, `it is vital that our applications are designed and built to be accessible to all.` states the importance of this, and the standard will be used as a reference to achieve this.

PR #2826, opened 2021-03-25
handbook: Add `Getting cross-team feedback on RFC` section

valerybugakov:
Initially, I thought it would be great to have a separate board for RFC tracking, as we discussed briefly with @pdubroy. 

On the other hand, we can leverage issue labels to show only RFCs, which will benefit from a cross-team review. It will mostly help other teams to find RFCs to review on our board (we can pin board-link with RFC filter in query params in a #frontend-platform-rfc-feed channel). Also, we won't need to move the RFC issue on two boards. 

Let me know what you think about it.

### DIFF
--- a/handbook/engineering/web/frontend-platform/index.md
+++ b/handbook/engineering/web/frontend-platform/index.md
@@ -87,7 +87,7 @@ The team makes use of [tracking issues](../../tracking_issues.md) for tracking p
 #### Getting cross-team feedback on RFC
 
 - Create a [tracking issue](../../tracking_issues.md) on our [Kanban board](https://github.com/orgs/sourcegraph/projects/144).
-- Create a separate issue for the RFC tracking.
+- Create a separate issue for the RFC tracking. [Lock conversation](https://docs.github.com/en/github/building-a-strong-community/locking-conversations): RFC should be discussed in the Google doc.
 - Add _rfc/tracking_ and _team/frontend-platform_ labels to the RFC issue.
 - Link the RFC issue to the tracking issue.
 - Once RFC is ready for review, move it to the _In review_ column and assign _sourcegraph/frontend-devs _ or individual developers to the issue.
--- a/handbook/engineering/web/frontend-platform/index.md
+++ b/handbook/engineering/web/frontend-platform/index.md
@@ -8,7 +8,7 @@ The Frontend platform team (part of the Web org) defines and maintains the stand
 - [Alicja Suska](../../../../company/team/index.md#alicja-suska-she-her) ([Product Designer](../../../product/roles/index.md#product-designer))
 - [Patrick Dubroy](../../../../company/team/index.md#patrick-dubroy-he-him) ([Engineering Manager](../../../engineering/roles/index.md#engineering-manager)) {#frontend-platform-eng}
   - [Tom Ross](../../../../company/team/index.md#tom-ross-he-him)
-  - [Felipe Janer](../../../../company/team/index.md#felipe-janer-he-him)  
+  - [Felipe Janer](../../../../company/team/index.md#felipe-janer-he-him)
   - [Valery Bugakov](../../../../company/team/index.md#valery-bugakov-he-him) moves in FQ2 to the [Code Insights Team](../code-insights/index.md)
 
 ## Mission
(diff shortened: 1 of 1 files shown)


handbook/engineering/web/frontend-platform/index.md
```diff
@@ -84,6 +84,15 @@ We plan and track our day-to-day work on our [Kanban board](https://github.com/o
 
 The team makes use of [tracking issues](../../tracking_issues.md) for tracking progress on the implementation of new features. The teammates should ensure that a tracking issue is created when starting work on features that are expected to take longer than a few days (or require multiple PRs) to deliver.
 
+#### Getting cross-team feedback on RFC
+
+- Create a [tracking issue](../../tracking_issues.md) on our [Kanban board](https://github.com/orgs/sourcegraph/projects/144).
```
umpox, line 89: Is the tracking issue still important if we already have an issue for the RFC? Can we not follow this process with one issue?

umpox (approved):
Good idea 👍

handbook/engineering/web/frontend-platform/index.md
valerybugakov: 1. The tracking issue will include links to other action items which should appear on RFC approval. It will allow tracking the RFC implementation progress whereas the RFC issue should be moved to `done` once it's approved.
2. Right. We mention this case in the [tracking issues section](https://about.sourcegraph.com/handbook/engineering/web/frontend-platform#tracking-issues)👇

> The teammates should ensure that a tracking issue is created when starting work on features that are expected to take longer than a few days (or require multiple PRs) to deliver

umpox: Hmm I've only just noticed that the tracking issues text has changed. I think previously it was like `expected to take longer than a single iteration`. I feel like a tracking issue should created after an RFC is approved in this case, as it'll likely be empty until then and there will be some cases where the RFC is rejected.

My main reason for this is just to make it as easy as possible for us to follow this process (so we actually do it 😁 )

handbook/engineering/web/frontend-platform/index.md
```diff
@@ -84,6 +84,15 @@ We plan and track our day-to-day work on our [Kanban board](https://github.com/o
...
+- Create a separate issue for the RFC tracking.
```
5h1rU, line 90: Is there a way to block comments on issues? 

Issues naturally are spaces for discussions and we are using it as a way to track an external source (Google Docs).

valerybugakov: Yep, it we can [lock the conversation](https://docs.github.com/en/github/building-a-strong-community/locking-conversations).

valerybugakov: Added note about that 👍

handbook/engineering/web/frontend-platform/index.md
pdubroy: I like the idea of keeping it as simple as possible so that we follow it.

But as an aside, let's remember that it's a request for _comments_ and not request for _approval_. :-) But, the points still stands that we might not actually pursue the work after getting the comments.

valerybugakov: Moved the tracking issue to the last step and made clear that it's optional for small features.

5h1rU (approved):

eseliger (approved):

handbook/engineering/web/frontend-platform/index.md
```diff
@@ -84,6 +84,15 @@ We plan and track our day-to-day work on our [Kanban board](https://github.com/o
 
 The team makes use of [tracking issues](../../tracking_issues.md) for tracking progress on the implementation of new features. The teammates should ensure that a tracking issue is created when starting work on features that are expected to take longer than a few days (or require multiple PRs) to deliver.
 
+#### Getting cross-team feedback on RFC
+
+- Create an issue for the RFC tracking on our [Kanban board](https://github.com/orgs/sourcegraph/projects/144).
+- [Lock conversation](https://docs.github.com/en/github/building-a-strong-community/locking-conversations): RFC should be discussed in the Google doc.
+- Add _rfc/tracking_ and _team/frontend-platform_ labels to the RFC issue.
+- Once RFC is ready for review, move it to the _In review_ column and assign _sourcegraph/frontend-devs_ or individual developers to the issue.
+- Github integration will notify _@web_ in #frontend-platform-rfc-feed that the RFC is ready for review.
```
felixfbecker, line 93: Why the whole web org specifically? Shouldn't it either be all frontend devs, or only frontend platform?

PR #2828, opened 2021-03-25
WIP Security Manager job req

bcreager:
First pass at security manager job posting.  Wanted to get it up to start getting some feedback,


nicksnyder:
Lever is the source of truth for Job descriptions now (it used to be the handbook), so I would recommend writing the JD in a google doc first for collaboration

PR #2829, opened 2021-03-26
add dev tool time landing page

vanesa:
Adding a landing page for our upcoming web series Dev Tool Time. This is a rough draft to test how it could look like and if we can add a Hubspot form.
Unfortunately I'm getting an issue with the Hubspot form breaking the page and I can't figure out why. Apparently when 
```
const hubspot = document.getElementById(targetId)
    hubspot!.appendChild(script)
```

runs in https://sourcegraph.com/github.com/sourcegraph/about/-/blob/website/src/components/HubSpot.tsx
hubspot is null.
I don't see any difference to other similar pages e.g. https://sourcegraph.com/github.com/sourcegraph/about/-/blob/website/src/pages/contact/request-batch-changes-demo.tsx So I'm not sure why my code leads to the break. 
Would appreciate any help, thanks!

### DIFF
--- a/website/src/components/content/WebSeriesPage.tsx
+++ b/website/src/components/content/WebSeriesPage.tsx
@@ -1,0 +1,83 @@
+import React from 'react'
+import { RequestDemoAction } from '../../css/components/actions/RequestDemoAction'
+import { COLORS } from '../Jumbotron'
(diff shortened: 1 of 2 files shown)


vanesa:
Using a Hubspot landing page instead.

PR #2833, opened 2021-03-28
Update support-workflow.md

adeola-ak:
changed the line which mentioned a slack chat which the name has been changed from customer-support to customer-support-chat


virginiaulrich (approved):
Thanks for updating this! <3

PR #2835, opened 2021-03-29
Update support-schedule.md

adeola-ak:
how does the focus block summary sound? Should I add anywhere else?

### DIFF
--- a/handbook/ce/support-schedule.md
+++ b/handbook/ce/support-schedule.md
@@ -6,6 +6,8 @@ This starts by everyone on the team identifying their ideal working hours. If we
 
 Our schedule also makes it clear when folks on the team have dedicated heads-down focus time.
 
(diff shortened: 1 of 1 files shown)


virginiaulrich (changes requested):
You have all the right stuff, and I think I'd like to add some more context + reorder the framing slightly so we lead with what matters most for the company. Could we change it to this?

> The support workflow can quickly turn into a ticket factory, which slows down individual and collective team growth in becoming more proficient in our work. In order to mitigate this, focus time blocks ensure that the support team has dedicated time to focus on specific personal projects or development outside of working on tickets. During focus hours, the team is aware you are unavailable to meet and take up new tickets (but please make yourself available if there is an urgent matter). This time is intended to increase productivity and build positive momentum that you can use for the rest of the day. Allow two half-day blocks weekly in your schedule for focus time. If you have a ticket that cannot be put aside during your block, raise it to the team in #customer-support-team and we will figure out how to honor your focus time and help the customer.

adeola-ak:
yes I love the change

virginiaulrich (approved):
Marking approved and will let you make the change before merging (per our comments!). Thanks!

PR #2840, opened 2021-03-30
Update support-workflow.md

virginiaulrich:
Adding NPS details -- our workflow remains the same, but the details of how these things get to us are now captured here. @airamare01 we will want to explain this in the triage doc, too

### DIFF
--- a/handbook/ce/support-workflow.md
+++ b/handbook/ce/support-workflow.md
@@ -75,11 +75,9 @@ As explained above, our customers are able to initiate conversations for help vi
 * **Details:** This is a primary collaboration space for us to work with folks across the company and them to work with us.
 
 ### NPS responses
-*Note: This will be outlined very soon*
-
-* **Purpose:**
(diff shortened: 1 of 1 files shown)


virginiaulrich:
Additional context for your review @adeola-ak @abeatrix @northyg @airamare01 @DaedalusG -- prior to today, we had everything going into Zendesk and when triaging, I had to remove it as no action. Talked with Christina, Aimee, and Tia and we have opted for this approach instead. For you all, it's a support issue, but still good to know 1) the nuance of how it gets to ya and 2) that these will come in via email. This review is an FYI review so you aware, though as always, if anything isn't clear, I welcome suggestions to improve this write-up before we publish!

abeatrix (approved):
noted!

airamare01 (approved):
Noticed that this is mentioned in our triaging draft doc. Will include more details. It looks good.

PR #2843, opened 2021-03-30
create customer support triaging page

airamare01:
Creating the customer support triaging page on the handbook


virginiaulrich (changes requested):
Looks like this PR needs to be closed since this is the correct PR https://github.com/sourcegraph/about/pull/2846/files > please do so so we can keep our repository tidy!

PR #2861, opened 2021-04-01
engineering onboarding: update Percy instructions

limitedmage:
Percy accounts are a mess, modifying the instructions to join our Percy org mentioning that engineers will need to create a legacy account. See https://docs.percy.io/docs/accepting-invites-to-a-classic-percy-organization


camdencheek (commented):
I'd also specify "create an account through github" since you can also create an email-based account from that page.

camdencheek (approved):

PR #2865, opened 2021-04-02
Update "How to work with BizOps"

kelsey-brown:
Including language about adding issues in the analytics repo to the project board


ebrodymoore:
Does "add it to Business Operations project" = add to the project board or is it a tag in the issue?

kelsey-brown:
You can do it in the issue itself (on the right side under project) Is that generally referred to as a tag?

ebrodymoore (approved):
You're right, I always forget we can do that!

PR #2868, opened 2021-04-03
Update transparency chart

tammy-zhu:
Link to RFC handbook page for ease of reference and add to examples of private information


dadlerj (approved):
Thanks

PR #2874, opened 2021-04-05
Adding Virginia as assignee for non-action tickets

airamare01:
For data integrity, we need to make sure that Virginia is the assignee for tickets that don't require CS action


virginiaulrich (approved):
This captures it! It might make it a bit easier to read if you make step 4 "select assignee" and then have each of these two scenarios as two sub-points under this step, like:

Step 4: Select assignee
- If the ticket requires CS action:
- If the ticket does not require CS action:

Either way is good with me, so please only make this change if it suits you and it's ready for merging from my perspective, either way!

airamare01:
Closing this PR and replacing with this [one](https://github.com/sourcegraph/about/pull/3853)

PR #2883, opened 2021-04-07
Update visa-sponsorship.md

mercadon:


dadlerj (approved):

dadlerj:
This makes sense, and I think it is the right call to set this explicitly instead of relying on "case by case" terminology. I assume this falls into the same bucket as other benefits, though, where we can't go outside of this standard for special circumstances without some risk related to treating different people differently?

mercadon:
Yes, having "case by case" wording also makes it difficult to for us to define what cases would get approved by us and which ones wouldn't. And since most of these conversations would need to happen in the early interview stages, it would be even harder to make a call at that point.

PR #2890, opened 2021-04-07
Add director-level finance approvals

dadlerj:
Any thoughts? These all need a refresh, but I think this structure makes sense for director level teammates—they need to have their manager's buy-in for most purchases, but the finance involvement can remain high (the same level as VP)


TommyODonnell (approved):
Hey Dan, these limits make sense to me for sure.

PR #2900, opened 2021-04-09
Add changes for search-core and search-product

DaedalusG:
Our engaging team's docs do not currently contain team-specific contact protocols and are meant more broadly as a filter for who to call on various types of issues. This is an example of what including team-specific contact info here might look like. Additional changes for keywords might be good to edit into this PR


virginiaulrich (approved):
Thanks for updating this part of it!

PR #2922, opened 2021-04-13
Update locations.geojson

jharwell27:


company/team/locations.geojson
```diff
@@ -754,6 +754,18 @@
...
+       "properties": {
+         "name": "John Harwell"
+      }, 
+      "geometry": {
+        "type": "Point",
+        "coordinates": [
+          -84.566860,
+          34.144132 
+        ]
+      }
```
LawnGnome, line 769: Looks like we're missing a brace here. This should fix it:

```suggestion
      }
    }
```

LawnGnome (changes requested):
Just one little change needed, which you should be able to commit directly from my suggestion, and then I think we're good!

LawnGnome (approved):
Looks good! Thanks, and feel free to merge whenever you're ready.

PR #2926, opened 2021-04-14
Add convention for shortening URLs

InesRoitman:


handbook/communication/index.md
```diff
@@ -84,6 +84,17 @@ See "[1-1 meetings](../leadership/1-1.md)".
 
 See "[Announcements](announcements.md)".
 
+## Shortening URLs
```
dadlerj, line 87: Just to check @sarah-reece (since you were involved in setting up these URLs, feel free to ask someone else), do we prefer one or another of these? If we want to always use srcgr.ph, for example, we could remove the other two listed here...

handbook/communication/index.md
```diff
@@ -84,6 +84,17 @@ See "[1-1 meetings](../leadership/1-1.md)".
...
+Using shorter URLs is much easier to fit into small spaces or images than the full link to our page or handbook.
```
dadlerj, line 88: I _think_ you need an extra line spacing between the header and the text, but this might not be 100% necessary.

```suggestion

Using shorter URLs is much easier to fit into small spaces or images than the full link to our page or handbook. 
```

dadlerj (approved):
Thanks! Left one suggestion (which you can commit, if that's fine for you), and a question for Marketing on the URLs. Otherwise, this looks great.

handbook/communication/index.md
sarah-reece: @dadlerj @InesRoitman We've been using srcgr.ph - we secured the other two proactivey.

sarah-reece (commented):
Approved

PR #2933, opened 2021-04-15
Update style_and_mechanics.md

abschu:
@content-hacker @tommypez - I made a few changes to the guidelines. Please review and let me know if you have any feedback.

### DIFF
--- a/handbook/communication/content_guidelines/style_and_mechanics.md
+++ b/handbook/communication/content_guidelines/style_and_mechanics.md
@@ -804,7 +804,7 @@ Refer to the natural noun of the product or feature directly.
 <div class="item no">
 <h5>No</h5>
 <ul>
-<li>The Extensions product is available.</li>
+<li>The extensions feature is available.</li>
 </ul>
 </div>
 </div>
--- a/handbook/communication/content_guidelines/style_and_mechanics.md
+++ b/handbook/communication/content_guidelines/style_and_mechanics.md
@@ -726,26 +726,26 @@ We capitalize product names. Qualify product names with `Sourcegraph $FEATURE` o
 
 List of product names: 
 
+- Code Search
 - Batch Changes
 - Code Insights
 - Code Intelligence
-- Universal Code Search
 
 <div class="usage">
 <div class="item yes">
 <h5>Yes</h5>
 <ul>
-<li>With Sourcegraph Universal Code Search you can...</li>
-<li>You can search across all your repositories with Universal Code Search.</li>
+<li>With Sourcegraph Code Search you can...</li>
+<li>You can use Code Search to search across all your repositories.</li>
 <li>Companies benefit from a universal code search tool because...</li>
-<li>Sourcegraph Batch Changes are... Batch Changes allow you to...</li>
+<li>Sourcegraph Batch Changes allows you to...</li>
 <li>To create a batch change...</li>
 </ul>
 </div>
 <div class="item no">
 <h5>No</h5>
 <ul>
-<li>Use Sourcegraph universal code search to...</li>
+<li>Use Sourcegraph code search to...</li>
 <li>With code insights you can...</li>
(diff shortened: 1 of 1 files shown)


handbook/communication/content_guidelines/style_and_mechanics.md
```diff
@@ -781,13 +781,13 @@ Use the natural plural/singular form. If a product or feature name is a plural n
 <div class="item yes">
 <h5>Yes</h5>
 <ul>
-<li>Batch Changes are available.</li>
+<li>Over 100 extensions are available.</li>
```
content-hacker, line 784: I think Batch Changes is a better example for `Use the natural plural/singular form. If a product or feature name is a plural noun, treat it like a plural noun` The change made is showing the difference between capitalizing or not, which is not what's covered in this section.

handbook/communication/content_guidelines/style_and_mechanics.md
```diff
@@ -798,13 +798,13 @@ Refer to the natural noun of the product or feature directly.
 <div class="item yes">
 <h5>Yes</h5>
 <ul>
-<li>Batch Changes are available.</li>
+<li>Extensions are available.</li>
```
content-hacker, line 801: it may be helpful to include both an example of a product and feature (so include Batch Changes and Extensions)

handbook/communication/content_guidelines/style_and_mechanics.md
```diff
@@ -798,13 +798,13 @@ Refer to the natural noun of the product or feature directly.
...
 </ul>
 </div>
 <div class="item no">
 <h5>No</h5>
 <ul>
-<li>The Batch Changes product is available.</li>
+<li>The Extensions product is available.</li>
```
content-hacker, line 807:
```suggestion
<li>The extensions feature is available.</li>
```

content-hacker (commented):
@abschu left a few comments. Thanks for updating this!

handbook/communication/content_guidelines/style_and_mechanics.md
abschu: @content-hacker - I thought we treated Batch Changes as singular when referencing the product.

Batch Changes is now available.

content-hacker: If that's the case then the section headline needs to be updated and we should treat all products this way (I don't think it's an issue with other ones). Is it confusing to say "we treat products as singular and features as plural?"

content-hacker: Code Insights is now available would be the other. It does feel strange to treat these as singular. It doesn't look like we ever say Batch Changes is/are in any of the documentation/landing page/blog post.

content-hacker: But you're right, "Code Insights" and "Batch Changes" are singular products. Isn't language fun 😆

tommypez (commented):
I think Erica mentioned Batch Changes as a product name would be singular: "Batch Changes is now available for enterprise accounts; Batch Changes allows you to" etc.

handbook/communication/content_guidelines/style_and_mechanics.md
abschu: On https://about.sourcegraph.com/batch-changes we use Batch Changes in the singular "Batch Changes gives you a declarative structure for finding and modifying code across all of your repositories."

So the main things we want to convey are:
1. L1 products as defined in the product hierarchy use Title Case (when referencing the product)
2.  L1 products as defined in the product hierarchy are singular (when referencing the product).
3. Use natural plural/singular and lowercase when describing the functionality "The code insights I just generated are super useful."

Is that it?

CC @content-hacker @tommypez

content-hacker: @abschu yes, this seems right to me.

PR #2938, opened 2021-04-16
Create ce.md

amenne:
Adding Customer Engineer team page, synonymous with the CSE [support.md](https://about.sourcegraph.com/handbook/ce/support) page to provide greater clarity on CE & CSE teams. Will link from main ce page which currently directs to job posting.


handbook/ce/ce.md
```diff
@@ -0,0 +1,41 @@
+# Customer Engineers
+
+## Our ethos
+
+Customer Engineers at Sourcegraph are a combination of pre-sales sales engineering and post-sales customer success. We own the relationship with our developer users and champions at our customers and prospects. We serve as the technical experts during the sales cycle and help enable our customers throughout our journey together.
```
virginiaulrich, line 5: I wonder if there is a different way to say "we own the relationship with our...." to convey the directly responsible individual (DRI) aspect, as well as the tone that our customers are always "our customers" and never "my customer..."

handbook/ce/ce.md
```diff
@@ -0,0 +1,41 @@
...
+
+We show up for our customers, open source users, and teammates by living up to our Sourcegraph values and handbook, as well as in the following ways:
+
+- We deeply know and understand our customers and their needs; we enable them and guide them
+- We teach our customers new and better ways to get value from Sourcegraph as thought leaders and evangelists
+- We advocate on behalf of our customers internally to product and engineering
```
virginiaulrich, line 11: Given our conversations in recent days, it feels like this maybe could be reworded, too ... to reflect about doing what is in the best interest of our customers. I think Quinn's thoughts in the Uber thread make it clearer what this means, but maybe also not clear enough. I could just see this creating friction if folks ask CE's questions to understand, maybe even challenge and that feeling like they are alone in advocating ... if that makes sense?

virginiaulrich (approved):
Left a couple of comments -- very nitpicking about words, mostly because I was a lit major and I refuse to stop believing words matter 😬. Really appreciate you pulling this together!

handbook/ce/ce.md
amenne: Thank you for this. 

How about from: We own the relationship with our developer users and champions at our customers and prospects.

To: We are first and foremost responsible for building and fostering relationships with developer users and champions at our prospects and customers.

virginiaulrich: I love that change!

handbook/ce/ce.md
tiawucher: @virginiaulrich could it be "We, in collaboration with AEs and CSEs advocate on behalf of our customers internally to product and engineering" something like that?

amenne: I was more so thinking here of advocation in terms of voice of the customer, feedback, feature requests, etc.

How about from: We advocate on behalf of our customers internally to product and engineering
To: We serve as the voice of the customer to internal teams, providing key insights and feedback, and raising feature requests.

Also open to other suggestions!

tiawucher (approved):
Dig the change! Thank you for the focus on this...I know it will help!

tiawucher (approved):

handbook/ce/ce.md
virginiaulrich: I like your v2, Aimee!

PR #2939, opened 2021-04-16
Add missing word

vanesa:
Original: let them describe what the campaign should like in a YAML file?
Change: let them describe what the campaign should `look` like in a YAML file?


chrispine (approved):
Thanks for catching this!

PR #2971, opened 2021-04-23
Fix docsite check

daxmc99:
Docsite check is failing with 
`handbook/careers.md: disconnected page (no inlinks from other pages)` 

This adds a small link under the **Talent** heading


nicksnyder:
Oops, thanks, you can just commit directly to main for stuff like this

daxmc99:
Ahh yup, I will do that next time!

PR #2996, opened 2021-04-28
indeed case study page

mgyorffy:


tommypez (approved):
Looks good, can I simply squash and merge these when they come up?

mgyorffy:
> Looks good, can I simply squash and merge these when they come up?

In general, it sounds good :)

1. An exception: if the PR i a DRAFT PR, please don't merge it, then means I'm till working on it. (eg. like the PR for the code-search-guide-page)
2. And later when we have more complex changes we should give a chance to Farhan/Marek to comment on the code changes,
otherwise it sounds good :)

PR #2999, opened 2021-04-29
Update to reflect that we're not using Vault

ElizabethStirling:


andreeleuterio (approved):
LGTM (chaotic good according to the talk today)

PR #3059, opened 2021-05-07
Updating redirects for support page moves

virginiaulrich:


virginiaulrich:
@jeanduplessis do you mind making sure I truly did this correctly? I noticed looking at the list that a lot of things didn't end in the `/`, so I left that off ... but perhaps that was wrong? Thank you for teaching me how to do this!

_resources/assets/redirects
```diff
@@ -41,4 +41,5 @@
 /handbook/product/personas /handbook/marketing/personas 308
 /handbook/values /company/values 308
 
-
+/handbook/ce/support /handbook/support 308
+/handbook/ce/support/#slas /handbook/support#our-service-level-agreements-slas 308
```
jeanduplessis, line 45:
```suggestion

```
This rule isn't needed/won't work due to the fragment (# symbol and stuff behind it) never reaching the backend, only the client (browser) knows about it.

jeanduplessis: You could add an additional anchor to the SLAs section on the support page to catch the old fragment.

virginiaulrich: Ah got it. Will update that page, then, too. Thank you!

PR #3074, opened 2021-05-10
Add Turn the Ship Around and Orbiting the Giant Hairball to eng onboarding

nicksnyder:
Many teammates have read these books and found them to be very valuable.

> Huge +:100: on Orbiting the Giant Hairball. Nick recommended it to me when I joined. That combined with what’s in the book dramatically shifted my outlook on my work and how I approach it. Highly recommended.

We already have engineering managers read this book as part of EM onboarding, and we should have engineers read these books too.

If you are already on the team and haven't read these books yet, please do! (you can expense them!)


slimsag:
As communicated to the team, this sounds like highly encouraged but still optional reading.

As communicated here, this sounds like mandatory homework with the expectation that you will report back on what you've learned.

I'm sure these books are great, and understand the value in highly encouraging reading of them - but not everyone learns in the same way (I certainly would struggle if I was onboarding and thought I needed to read these on some mandatory timeline)

All this to ask: should this be mandatory? If yes, should it not also be mandatory for people already working here? If no, is there a way we can rephrase the wording here to make it sound more highly-encouraged-take-your-time but not strictly mandatory?

jeanduplessis:
I am 💯 on board with making this part of engineering onboarding. I asked new teammates in Developer Insights to read Turn the Ship Around and it has been really helpful in helping frame conversations around expectation and high agency.

@slimsag my 2 cents is that I wouldn't make it mandatory for existing teammates to read, but would absolutely highly encourage it. And considering the value I've seen new teammates get out of reading it, I think making it part of the onboarding guideline is a good idea. 

In my view, teammates and managers should always treat onboarding items as guidelines and adjust them according to the needs and preferences of the individual.

PR #3076, opened 2021-05-10
Adding handbook-announce to editing hb page

InesRoitman:


jeanduplessis (approved):
@InesRoitman thanks for adding this. I just fixed the list numbering so that the sequence shows correct.

PR #3092, opened 2021-05-13
update terminology guidelines

abschu:
@beyang - any concerns with adding "open source code" vs "public code" to the terminology guidelines?


beyang:
Mmm, I think there is a distinction here. "Public" indicates code that is publicly available, regardless of license, whereas "open source" is more specific implying usage of an open-source software license. Open source implies public but the converse doesn't hold.

Re: "Sourcegraph for public code" vs. "Sourcegraph for open source", the former is more technically correct, since we don't restrict Sourcegraph to indexing just open-source licensed public code, but the latter is a better term to use for marketing as it connects better with our audience and code that's public but not open source is not widely used.

abschu:
@beyang - thanks for the clarification. It sounds like there are cases where we would use public. Since it's not a hard and fast rule, I don't need to add it here.

PR #3104, opened 2021-05-17
Developer insights: add guidance around PR size

pdubroy:
- Add guidance (discussed with @felixfbecker and @valerybugakov) about keeping PRs <400 changed lines.
- Clarify the meaning of "blocked" with regards to PR reviews

### DIFF
--- a/handbook/engineering/developer-insights/index.md
+++ b/handbook/engineering/developer-insights/index.md
@@ -62,7 +62,14 @@ We want teammates to do what is best for the org as a whole. Don't optimize for 
 
 ##### Prefer small PRs (<400 lines)
 
-We extend Sourcegraph's company-wide guidance (see [what makes an effective PR](https://docs.sourcegraph.com/dev/background-information/code_reviews#what-makes-an-effective-pull-request-pr)) with a specific guideline that _PRs should contain less than 400 changed lines_ (excluding tests). Note that this is a _guideline_ and not a hard limit: there are situations where it doesn't make sense (e.g. PRs that are mostly mechanical changes).
+We extend Sourcegraph's company-wide guidance (see [what makes an effective PR](https://docs.sourcegraph.com/dev/background-information/code_reviews#what-makes-an-effective-pull-request-pr)) with a specific guideline that _PRs should contain less than 400 changed lines_ (excluding tests). Note that this is a _guideline_ and not a hard limit; there are situations where it doesn't make sense (e.g. PRs that are mostly mechanical changes).
+
+There are several reasons to prefer small PRs:
+
+- Reviews happen more quickly ("I'll just review this quickly right now" instead of "Hmmm, better schedule time for this later").
+- It enables higher-quality reviews, because there's less context for the reviewer to hold in their head. It's also easier to suggest major changes when your teammate has spent only a few hours on a PR, rather than a day or more.
+- It encourages a tighter feedback loop.
+- Smaller, atomic changes are easier to roll back if required.
 
 ##### Keep reviewers to a minimum
 
--- a/handbook/engineering/developer-insights/index.md
+++ b/handbook/engineering/developer-insights/index.md
@@ -62,7 +62,7 @@ We want teammates to do what is best for the org as a whole. Don't optimize for 
 
 ##### Prefer small PRs (<400 lines)
 
-We extend Sourcegraph's company-wide guidance (see [what makes an effective PR]((https://docs.sourcegraph.com/dev/background-information/code_reviews#what-makes-an-effective-pull-request-pr)) with a specific guideline that _PRs should contain less than 400 changed lines_ (excluding tests). Note that this is a _guideline_ and not a hard limit: there are situations where it doesn't make sense (e.g. PRs that are mostly mechanical changes).
+We extend Sourcegraph's company-wide guidance (see [what makes an effective PR](https://docs.sourcegraph.com/dev/background-information/code_reviews#what-makes-an-effective-pull-request-pr)) with a specific guideline that _PRs should contain less than 400 changed lines_ (excluding tests). Note that this is a _guideline_ and not a hard limit: there are situations where it doesn't make sense (e.g. PRs that are mostly mechanical changes).
 
 ##### Keep reviewers to a minimum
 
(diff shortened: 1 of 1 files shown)


pdubroy:
cc @sourcegraph/frontend-platform

jeanduplessis (approved):

handbook/engineering/developer-insights/index.md
```diff
@@ -50,17 +50,32 @@ In addition to the [engineering principles and practices](../principles-and-prac
 
 ### Unblocking others is your highest priority
 
-If a teammate is blocked by you on a question, your approval, or a pull request review, your top priority is always to unblock them, either directly or through helping them find someone else who can, even if this takes time away from your own or your team's priorities.
+If a teammate is blocked by you on a question, your approval, or a pull request review, your top priority is always to unblock them, either directly or through helping them find someone else who can, even if this takes time away from your own or your team's priorities. If you're the one who is blocked, be sure to communicate that so that others can prioritize appropriately.
+
+Normally, waiting for a PR review does not mean you are blocked: it's expected that you can start working on something else (e.g. a new PR that depends on the first one). However, in some scenarios it's important to get a review ASAP — e.g., to fix a regression or a broken CI pipeline. In those cases, you should communicate the urgency and expect that your teammates will prioritize unblocking you.
```
felixfbecker, line 55: I'm a bit worried here about making it black-and-white blocked vs not blocked, when in reality a review delay is always at least an _inconvinience_ (because of uncertainty in  follow-up work, potential merge conflicts, needing to rebase stacked PRs, delay in dogfooding, ...). Maybe we can make it clear that this is not a wildcard to delay reviews for days or weeks after e.g. whatever project you're currently working on is done. Imo every dev should generally budget some time every day (e.g. in the morning) to do some PR reviews.
```suggestion
Normally, waiting for a PR review does not mean you are blocked: it's expected that you can start working on something else (e.g. a new PR that depends on the first one). However, in some scenarios it's important to get a review ASAP — e.g., to fix a regression or a broken CI pipeline. In those cases, you should communicate the urgency and expect that your teammates will prioritize unblocking you.
Even when a teammate is not _blocked_, but _inconvinienced_ (because of follow-up work), don't leave them hanging for extended periods of time. You should generally budget some amount of time every day for doing reviews.
```

handbook/engineering/developer-insights/index.md
```diff
@@ -50,17 +50,32 @@ In addition to the [engineering principles and practices](../principles-and-prac
...
 We want teammates to do what is best for the org as a whole. Don't optimize for the goals of your team when it negatively impacts the goals of other teams, our users, and/or the company. Those goals are also your problem and your job.
 
-### Pull request reviews
+### Pull requests
+
+#### For authors
+
+##### Prefer small PRs (<400 lines)
+
+We extend Sourcegraph's company-wide guidance (see [what makes an effective PR](https://docs.sourcegraph.com/dev/background-information/code_reviews#what-makes-an-effective-pull-request-pr)) with a specific guideline that _PRs should contain less than 400 changed lines_ (excluding tests). Note that this is a _guideline_ and not a hard limit; there are situations where it doesn't make sense (e.g. PRs that are mostly mechanical changes).
```
felixfbecker, line 65: Excluding tests will make it a bit harder to self-assess this as GitHub's diff indicator won't do that. Tests need to be reviewed too, maybe omit that rule?

felixfbecker (approved):
LGTM! But more important to hear from the team :) @sourcegraph/code-insights @sourcegraph/extensibility @sourcegraph/frontend-platform

handbook/engineering/developer-insights/index.md
pdubroy: We had this on my previous team, and the reasoning was: if tests are included, there will always be an incentive to skimp on tests in order to keep your PRs under the guideline.

But, I also agree with your point. I'm fine with going either way.

handbook/engineering/developer-insights/index.md
```diff
@@ -50,17 +50,33 @@ In addition to the [engineering principles and practices](../principles-and-prac
...
+If a teammate is blocked by you on a question, your approval, or a pull request review, your top priority is always to unblock them, either directly or through helping them find someone else who can, even if this takes time away from your own or your team's priorities. If you're the one who is blocked, be sure to communicate that so that others can prioritize appropriately.
+
+Normally, waiting for a PR review does not mean you are blocked: it's expected that you can start working on something else (e.g. a new PR that depends on the first one). However, in some scenarios it's important to get a review ASAP — e.g., to fix a regression or a broken CI pipeline. In those cases, you should communicate the urgency and expect that your teammates will prioritize unblocking you.
+Even when a teammate is not _blocked_, but _inconvinienced_ (because of follow-up work), don't leave them hanging for extended periods of time. You should generally budget some amount of time every day for doing reviews.
```
pdubroy, line 56:
```suggestion
Even when a teammate is not _blocked_, but _inconvenienced_ (because of follow-up work), don't leave them hanging for extended periods of time. You should generally budget some amount of time every day for doing reviews.
```

handbook/engineering/developer-insights/index.md
muratsu: Makes sense. Just curious, how do we enforce this rule? Is there a linter for this or do we just ask the PR to be broken down when it shows up?

pdubroy: @muratsu For now we should just ask for it to be broken up, but a CI check is a good idea!

pdubroy: BTW you're right that it's not trivial to assess the number of lines not including tests, but it's also not that hard.  First, it only matters if you are near the limit. Also, GitHub shows the lines changed per file, so you just need to scan through and mentally subtract any tests from the total. In my experience it's not a big deal. You might see a 600-line PR, then spend maybe 10-15 seconds checking how many lines are tests.

felixfbecker: > We had this on my previous team, and the reasoning was: if tests are included, there will always be an incentive to skimp on tests in order to keep your PRs under the guideline.
> 
> But, I also agree with your point. I'm fine with going either way.

@pdubroy If you had this situation in your previous team, I'd trust your experience.

> Just curious, how do we enforce this rule? Is there a linter for this or do we just ask the PR to be broken down when it shows up?

I think this rule will only work if engineers agree on it and give each other feedback on PRs when it feels out of line (using their judgement to identify exceptions, or when the PR should be even smaller). I _think_ a mechanical check would only make it more likely folks would try to make the metric the goal.

muratsu: Makes sense! 👍

pdubroy: Thanks for the feedback! Gonna merge now, and we can revisit this discussion in a few weeks if necessary.

PR #3114, opened 2021-05-18
add amplitude docs

ebrodymoore:


handbook/ops/bizops/amplitude.md
```diff
@@ -0,0 +1,29 @@
...
+
+| Type of analysis        | Tool      | Example |
+|-------------------------|-----------|---------|
+| Retention/engagement    | Amplitude | [Link](https://analytics.amplitude.com/sourcegraph/chart/7l5vdg4?source=workspace)    |
+| Adoption of public code | Looker    | [Link](https://sourcegraph.looker.com/dashboards-next/175)    |
+| Progress towards OKRs   | Looker    | [Link](https://sourcegraph.looker.com/dashboards-next/166)    |
+| On-prem instances/pings | Looker    | [Link](https://sourcegraph.looker.com/dashboards-next/174)    |
+| Signup funnel on Cloud  | Amplitude | [Link](https://analytics.amplitude.com/sourcegraph/chart/jumvevm?source=workspace)    |
+
+
```
kelsey-brown, line 24: Is this accurate? If so, I think we should state it. I think the first question people have when something new comes in is "how will this change my current workflow?" 
```suggestion
Any analysis that was conducted in Looker prior to the implementation of Amplitude can still be done in Looker, and existing Looker dashboards and visualizations will still be maintained. Amplitude will help us conduct new and different analysis regarding product analytics. 
```

handbook/ops/bizops/amplitude.md
```diff
@@ -0,0 +1,29 @@
...
+## Why Amplitude?
+Amplitude is a specialized product analytics tool that specializes in turning event data into actionable insights and dashboards. 
+
+### Why aren’t we using Looker for this?
+Looker is very flexible in that we have the ability to set it up for any purpose. But this also means that to provide a great self-service experience for what we want from a product analytics tool, this would take a LOT of work. We get these capabilities out of the box with Amplitude. 
+
+### What is in Looker vs. Amplitude?
+
+Anything not based directly on analyzing events is in Looker. This includes pings from on-prem instances, anything we get from the Cloud Postgres database, and any data from third-parties tools (such as Google Analytics and Salesforce).
```
kelsey-brown, line 14: I'm not really sure what you mean by "anything not based directly on analyzing events is in Looker." My interpretation of an "event" may be wrong, but don't we look at things like searches submitted and code intel hovers in Looker? I think of those as an analysis of events (for better or worse). The way I was interpreting Looker vs Amplitude is that (for now) anything related to the analysis of drivers of retention/adoption/engagement is best done in Amplitude, but other analysis is done in Looker. Is that right?

kelsey-brown (approved):
Looks good! A couple thoughts/potential additions.

handbook/ops/bizops/amplitude.md
ebrodymoore: We do look at events in Looker but after they're aggregated in pings - I'll make that more clear! 

And yeah Retention/engagement is best done in Amplitude because they handle the events out of the box in those specific ways, but over the past day I've started to look at it more like self-service for any product usage question, like this [breakdown between clicking def and ref](https://analytics.amplitude.com/sourcegraph/chart/3lcp9ju?source=workspace). So you are right, but I've also changed my thinking quite a bit in the last 24 hours 😄

PR #3122, opened 2021-05-19
Successful search sessions <> week 1 retention learnings

ebrodymoore:
You guys agree with this summary and conclusion(s)?

### DIFF
--- a/handbook/product/product_learning.md
+++ b/handbook/product/product_learning.md
@@ -2,6 +2,51 @@
 
 The conclusion of every project should include analysis of performance. Below is the beginnings of us keeping track of what we've learned based on usage data.
 
+- [Search homepages and tour (2020-11-30)](#2020-11-30)
+- [Value of browser extensions](#2020-12-06)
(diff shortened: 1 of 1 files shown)


ebrodymoore:
cc @macraig there's some interesting code intel takeaways here

malomarrec (commented):
Generally agree, but we should try and avoid getting into a set definition of search sucesss that does not actually represent success. 
If an assumption is that users need the search tour, it means that they need help learning the syntax to get the results they want, which means that they could get better results with more syntax, which means their previous sessions were unsuccessful. 

**Suggestion** Use wording such as "proxy for search success" or clarify that the current definition only captures a part of search success.

malomarrec (commented):
It'd be nice to always have
P(success | action) and P(success | not action)

malomarrec (approved):
It'd be nice to always have
P(success | action) and P(success | not action)

ebrodymoore:
Thanks Malo. Added some clarifications to reflect all this

PR #3130, opened 2021-05-20
Update index.md

amenne:
Flattened CE page to one (vs two from when CSE was included) and added further clarifications on CE <> CSE collaboration.


virginiaulrich (approved):
This looks 🎉 to me!

PR #3141, opened 2021-05-21
cloudflare logo and OG text fix

mgyorffy:


mgyorffy:
removed file: cloudflare-vector-logo.svg as it was a duplicate

PR #3163, opened 2021-05-25
Define temporary engineering management positions

jeanduplessis:


handbook/engineering/engineering-management.md
```diff
@@ -54,3 +54,21 @@ Here are some books that multiple people on the team have read and recommend. If
...
+
+### Acting Manager
+
+An Acting Manager occupies a role temporarily and will move back to their original position after a set amount of time or other conditions. An Acting Manager may be experimenting with the role as a part of determining their career development path or could be filling in for a vacant role while we hire.
+
+### Interim Manager
+
+Interim Manager positions are for teammates pursuing an Engineering Manager role long term but lack the necessary people management experience. The requirement is that, before moving into the role full time, they will make at least one successful hire. The official promotion will not occur before 30 days after that new hire's start date, so that we can assess whether the hire was truly successful. If the new hire's success is indeterminate at the 30-day mark, we will continue to review until a firm decision is made. If the new hire is not successful, that does not mean that the Interim Manager cannot eventually move into the full-time role.
```
nicksnyder, line 68: I am wondering if "promotion" is the right word to use here, or if we should use something more neutral.

(sometimes it could be a promotion and other times it could be a lateral move from staff eng to manager, right?)

```suggestion
Interim Manager positions are for teammates pursuing an Engineering Manager role long term but lack the necessary people management experience. The requirement is that, before moving into the role full time, they will make at least one successful hire. The official transition to Engineering Manager will not occur before 30 days after that new hire's start date, so that we can assess whether the hire was truly successful. If the new hire's success is indeterminate at the 30-day mark, we will continue to review until a firm decision is made. If the new hire is not successful, that does not mean that the Interim Manager cannot eventually move into the full-time role.
```

nicksnyder (approved):
Thanks for writing this up!

I think it would be helpful to link to this page from handbook/engineering/roles.md (since that page also speaks to switching roles on the engineering team)

handbook/engineering/engineering-management.md
jeanduplessis: I agree with your wording due to the lateral move posibility.

handbook/engineering/engineering-management.md
```diff
@@ -54,3 +54,21 @@ Here are some books that multiple people on the team have read and recommend. If
...
+
+## Temporary Management Positions
+
+We create temporary management positions when there is an organizational need. These may be filled by someone who is transitioning into the role (Interim Manager), experimenting with the role as they work on determining their career path (Acting Manager) or, someone who is just filling in and who is not interested in pursuing an Engineering Manager role long term (Acting Manager). This difference should be made explicit with the individual and team members before the temporary role is assigned.
+
+### Acting Manager
+
+An Acting Manager occupies a role temporarily and will move back to their original position after a set amount of time or other conditions. An Acting Manager may be experimenting with the role as a part of determining their career development path or could be filling in for a vacant role while we hire.
+
+### Interim Manager
```
bcreager, line 66: I'm iffy on this section.  I think we should be explicit on the Interim Manager position, that its a trial for x number of days.  Often I have seen someone stuck in "almost manager" position for years, with no end in sight.  If we have a need for a manager role, and we have someone that thinks they can level up into that role, and the appropriate manager AND manager's manager (director or VP depending on the case) agree, we should have a trial, with check-ins along the way


I think something like 6-months is acceptable, with check-ins on "almost manager" every month.  There is an escape shoot (one or both parties decide its not working) and there is a deadline for official manager promotion

handbook/engineering/engineering-management.md
```diff
@@ -54,3 +54,21 @@ Here are some books that multiple people on the team have read and recommend. If
   - Session 3 week of June 21
   - Session 4 week of July 5
   - Session 5 week of July 19
+
+## Temporary Management Positions
```
bcreager, line 58: I think it would be good to call out what this isn't here, something like:

```
Team lead vs Temp Manager - 
A temp manager's first priority is the team care and feeding.  This includes 1:1, various HR tasks (reviews, hiring screens, etc), outward communication, and process.  They may find they have some time to do technical work (coding, code reviews, etc), but that is secondary.

A team lead's first priority is making sure they help others get work done....etc etc
```
Needs better wordsmithing, but something like that I think will be useful

bcreager:
All my comments are non-blocking (NB - do we have a standard for this here).  This is far better than what we currently have, which is nothing

PR #3170, opened 2021-05-26
Fix table HTML

emchap:
Fixing table HTML for accessibility.

[_Created by Sourcegraph batch change `emily/Bulletpoints`._](https://demo.sourcegraph.com/users/emily/batch-changes/Bulletpoints)


emchap:
Gave up, trying in https://github.com/sourcegraph/about/pull/3283

PR #3228, opened 2021-06-04
Add documentation for the Review Buddy system

felixfbecker:
This is adding documentation to our handbook for the review buddy system which some folks in our org are already using (Vova and Valery), but I want to open up to everybody 🙂


tjkandala (approved):
❤️

valerybugakov (approved):

PR #3241, opened 2021-06-08
Document search product support rotation

lguychard:
Document:
- What our support rotation is
- Where it is defined (Opsgenie)
- What an engineer should do if they're unable to take on support


novoselrok (approved):

camdencheek:
Thanks Loïc!

PR #3249, opened 2021-06-08
Upgrade website dependencies

jeanduplessis:
Upgrades/removes dependencies in an attempt to get the website to build on M1 MacBook


jeanduplessis:
@attfarhan @marekweb would you mind having a look at this PR? 

Context: I was trying to help @rebeccadee be able to get the website to run locally on her M1 Macbook. To get it to work I had to upgrade a lot of dependencies, switch some out (node-sass => sass), and disable some. 

What I'm not sure of is if there might be any unintended consequences as a result of this. As far as I can see the website seems to work as it was before, but I would love some more eyes on this.

website/package.json
```diff
@@ -30,23 +29,24 @@
     "gatsby-source-filesystem": "^2.11.0",
     "gatsby-transformer-json": "^2.11.0",
     "gatsby-transformer-remark": "^2.16.0",
-    "gatsby-transformer-sharp": "^2.12.0",
+    "jquery": "^3.6.0",
```
attfarhan, line 32: Why we need to include jquery and popper.js now?

website/package.json
```diff
@@ -9,17 +9,16 @@
...
     "gatsby-plugin-feed": "^2.13.0",
     "gatsby-plugin-google-gtag": "^2.8.0",
     "gatsby-plugin-google-tagmanager": "^2.11.0",
     "gatsby-plugin-react-helmet": "^3.10.0",
     "gatsby-plugin-react-helmet-canonical-urls": "^1.4.0",
-    "gatsby-plugin-robots-txt": "^1.5.5",
+    "gatsby-plugin-robots-txt": "^1.6.2",
     "gatsby-plugin-sass": "^3.2.0",
-    "gatsby-plugin-sharp": "^2.14.1",
+    "gatsby-plugin-sharp": "^2.7.0",
```
attfarhan, line 21: I wonder if you can remove this too. It has been a while, but IIRC this is imported for use by gatsby-image. 

No harm leaving it in to be safe.

attfarhan (approved):
Thanks! This looks fine to me. Curious to hear if you know why these deps would break the build on an M1 Macbook?

website/package.json
jeanduplessis: @attfarhan they were coming up as a peer dependency that was missing.

website/package.json
jeanduplessis: @attfarhan it's actually required by the `gatsby-remark-responsive-image`.

jeanduplessis:
@marekweb I'm going to go ahead and merge this, but feel free to leave comments and I'll address them in a follow-up PR if needed.

marekweb:
Looks good to me although I'm surprised that we need to include jQuery and Popper now, I assume these are dependencies of `react-bootstrap` but I'm not sure.

In case this information is relevant in the future: if we upgrade to bootstrap v5, then the corresponding version of `react-bootstrap` is v2, and it's supposed to no longer require jQuery as a dependency.

jeanduplessis:
@marekweb yeah it was a peer dependency expectation of Bootstrap. 
Considering the website is going to be rebuilt soon I didn't want spend time thinking about optimizing anything rather just solving the problem of making sure it can run on M1 Macbooks without issue.

mgyorffy:
@marekweb @jeanduplessis we also need jquery for Clearbit that we are in the process of adding to our website

jeanduplessis:
Thanks for the additional context @mgyorffy

mgyorffy:
@jeanduplessis , the changes in styles.css are breaking the build on local, so I reverted them.

PR #3267, opened 2021-06-10
updating support schedule

airamare01:
Changing schedule for more EU timezone alignment


virginiaulrich (approved):
Thank you for updating this, Stompy!

PR #3283, opened 2021-06-11
Moving hyphens to HTML lists

emchap:
Sorry, this took forever—gave up on trying to do this with a campaign. Can you confirm I didn't screw up any formatting that I didn't catch?


felixfbecker:
You're a hero! 🦸🏻‍♀️

felixfbecker:
I did notice some unbalanced tags, and realized this would be a lot easier to read and edit by going all-in on HTML for the tables. I just pushed a commit with that change, let me know if this looks good to you!

emchap:
@felixfbecker Looks great; I _much_ prefer not handling them as markdown.

emchap:
Thank you!!

PR #3289, opened 2021-06-11
Document speed solution?

Joelkw:
If will continue to be relevant


slimsag:
no harm in documenting it, we can remove if it's useless later.

PR #3342, opened 2021-06-15
Moved MOPs under Demand Gen

jakevsen:
Cleaning up Demand Gen handbook and adjusting team layout.


rebecca-rissinger (approved):
First time approver here... hopefully we don't break anything!

PR #3350, opened 2021-06-15
Commissions.md

kmorris50:


TommyODonnell (commented):
Hey, Thanks for dropping this in. 
the only change here is for CE's, i(finance) will actually send them the detail for their invoice, for AE's it will be sales ops

PR #3368, opened 2021-06-16
Adding Process St link

InesRoitman:
What do you think about including the direct link to the home dashboard? everyone will see what they have access to


virginiaulrich (approved):
I did not know that we could do that and am so happy we can! That was my only heartache about moving the tasks is that they weren't visible in the handbook and 💥 of course here you are with the good news! Lovely!

PR #3382, opened 2021-06-17
Document blog release post script so it isn't lost

Joelkw:
Added the script to the release post

### DIFF
--- a/handbook/marketing/release_post_process.md
+++ b/handbook/marketing/release_post_process.md
@@ -13,7 +13,12 @@ Product marketing is responsible for creating the monthly release post in collab
 
 **Up until release ships**
 
-- VP Product runs the script to generate changelog and merges it into the release template. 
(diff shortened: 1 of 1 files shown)


handbook/marketing/release_post_process.md
```diff
@@ -13,7 +13,12 @@ Product marketing is responsible for creating the monthly release post in collab
 
 **Up until release ships**
 
-- VP Product runs the script to generate changelog and merges it into the release template. 
+- VP Product runs the script to generate changelog and merges it into the release template (note: make sure to update the *version number* in the script itself):
```
Joelkw, line 16:
```suggestion
- VP Product runs the script to generate changelog and merges it into the release template
   - Note: make sure to update the *version number* in the script itself. If the version number isn't added yet and you need to capture the "Unreleased" then passing it the literal string match for the "Unreleased" heading, usually `Unreleased`, in place of a version number works)
```

PR #3410, opened 2021-06-21
Highlight dates, fiscal years/quarters and add tooltips that show the date in the user's local date format

felixfbecker:
If you keep forgetting when different fiscal quarters start (like me), or if you're confused by the YYYY-MM-DD date format we use, or just would like to be able to know what day of the week a date referenced in the handbook is, this PR is for you.

It highlights all references to fiscal years, quarters, and dates/times with a dotted underline, and when hovered, will show the date (or start of the quarter) in the user's local date format complete with day of the week, timezone, etc.

Depends on https://github.com/sourcegraph/docsite/pull/73

### DIFF
--- a/_resources/assets/dateHighlighter.js
+++ b/_resources/assets/dateHighlighter.js
@@ -1,0 +1,41 @@
+import { parse } from 'https://cdn.skypack.dev/chrono-node'
+
+window.addEventListener('DOMContentLoaded', () => {
+    for (const element of document.querySelectorAll('time')) {
+        element.title = getDateTooltip(element)
+    }
+})
+
+/**
+ * @param {HTMLTimeElement} element
+ * @returns {string}
+ */
+function getDateTooltip(element) {
+    // See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time#valid_datetime_values for possible values of `element.dateTime`.
+
+    // Handle MM-DD reference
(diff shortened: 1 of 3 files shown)


_resources/assets/dateHighlighter.js
```diff
@@ -0,0 +1,49 @@
+import { parse } from 'https://cdn.skypack.dev/chrono-node'
```
pdubroy, line 1: Consider specifying a major version here and on the line below.

_resources/templates/root.html
```diff
@@ -11,6 +11,8 @@
 	<link rel="stylesheet" type="text/css" href="{{asset "document.css"}}" />
 	<link rel="stylesheet" type="text/css" href="{{asset "search.css"}}" />
 	<link rel="stylesheet" type="text/css" href="{{asset "content.css"}}" />
+	<link rel="stylesheet" type="text/css" href="https://cdn.skypack.dev/tippy.js/dist/tippy.css" />
```
pdubroy, line 14: If you do add a major version to the URLs in the import lines, don't forget to add it here too (and probably add a comment that the two need to be kept in sync).

pdubroy (approved):
Nice!

For things like "Q4", maybe it makes sense for the tooltip to be something like "Begins on October 1, 2022" or something? Otherwise it could be confusing for text like "End of Q4", like you have in your example GIF, if hovering _Q4_ shows "November 1, 2021".

PR #3423, opened 2021-06-22
Update broken links to team goals

fkling:
I noticed that most of the team goals are missing:

There are multiple ways to solve this but I didn't want to create changes to too many team pages so I thought updating the links here is more reasonable.

I haven't tested it because I edited this page directly in GH, but based on the instructions I believe this should work.


fkling:
I guess a review of this might not be necessary but I just want to make sure.

jeanduplessis (approved):

jeanduplessis:
> I guess a review of this might not be necessary but I just want to make sure.

Yeah not strictly necessary for fixes likes this. Thank you for doing these!

PR #3430, opened 2021-06-22
customer-first doc

virginiaulrich:
Created in collaboration with Aimee, Gregg, Nick, Christina, and Beyang ... this document reflects the details of what it means to be a customer-first company


amenne (commented):
Added a couple things but looks good

PR #3435, opened 2021-06-22
Clarifying CSE relationship to CE

virginiaulrich:
Thanks, Aimee and Jonah for pointing out how this old RFC language doesn't reflect agreed upon reality! Updated.


amenne:
Looks good to me!

amenne (commented):
Looks good to me!

justdueck:
Thanks for working on improvements to this @virginiaulrich! 

This is a great start to clarifying. Would it it be possible to add additional info in the last section around collaboration? Currently this would update it to: 

> Collaboration overview: We can think of CE and support as work best friends, working closely together every day, primarily communicating in Slack. CEs (or others -- including customers -- but primarily CEs) may engage support at any point during the pre-sales and post-sales process to share context to help support best serve our customers. 

My proposal is to add more info around *what* a CE is expected to be responsible for (i.e how-to questions, product questions, etc) vs what a CSE is responsible for (reactive technical support for issues), what the typically looks like when engaging with CSE, and maybe some common examples if possible. Also, this would imply that CEs typically only engage with Support for help on existing issues or for providing additional context, but what about when a technical question comes in that a CE does not know the answer to, but would not fall under the responsibilities of a CSE? Linking to or providing additional info on handling those types of issues (and who to go to for help if not Support), would be great as well.

virginiaulrich:
@justdueck  there is another doc in progress https://github.com/sourcegraph/about/pull/3430 that will be linked at the start of this section that captures all of this. So what you suggest is an update away and I'll keep this section to the value-add bit rather than duplicate.

PR #3441, opened 2021-06-23
Two quick product tweaks

Joelkw:
@virginiaulrich thanks for putting this whole page together! Just had two quick proposed updates in this PR: 

1. Link to the product feature request process so it's cross-linked and easier to find
2. Add "or product" to the escalation stage that CEs might loop CS back in. If this was an intentional omission, I'm happy to explain why I think product should be listed at that level, but since it's earlier "CS or product" I assume it was just a shortening and I wanted to make it explicit that the way to get it back to product is not AE->CE->CS->VP CE->VP Product->Product person, but just AE->CE->[CS (or) Product].


virginiaulrich (approved):
These are great edits, @Joelkw! Thank you for making them!

christinaforney (approved):
Thank you for the additions!

PR #3444, opened 2021-06-23
updated workflow to reflect account level docs

virginiaulrich:


abeatrix (approved):
Brilliant!

airamare01 (commented):
Liking the new changes more so the shared drive for account information and deployments

PR #3445, opened 2021-06-24
changes to customer-first.md

beyang:
@virginiaulrich most of this is just minor nits from reading through the doc, but there is one substantive change that I've called out in an inline comment.

### DIFF
--- a/handbook/company/customer-first.md
+++ b/handbook/company/customer-first.md
@@ -1,48 +1,48 @@
 # Being a customer-first company
 
-Some companies are engineering-first or sales-first, meaning they consider how that part of the organization needs to work to be successful and everything else is informed from there. Sourcegraph is a customer-first company. Our goal is to always do what is in the best interest of the customer while simultaneously setting up Sourcegraph and each team within it for long term success. This handbook page documents what this means and how it plays out day-to-day.
+Some companies are engineering-first or sales-first, meaning they consider how that part of the organization needs to work to be successful and everything else is informed from there. Sourcegraph is a customer-first company. Our goal is to always do what is in the best interest of the customer while simultaneously setting up Sourcegraph and each team within it for long-term success. This handbook page documents what this means and how it plays out day to day.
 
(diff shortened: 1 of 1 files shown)


handbook/company/customer-first.md
```diff
@@ -1,48 +1,49 @@
...
+* If CE doesn’t know the answer to a how-to question, they ask for help in the ask-product-eng Slack channel. This channel is monitored by product, engineering, and CS. It is a best effort approach (i.e., there is no SLA for response time for answers).
+* Any other request for help from engineering should go through CS (for troubleshooting) or product (for feature requests) and not to engineering directly. (For example, CEs are welcome to start troubleshooting an issue, but if they get stuck, the next step is to engage CS, and CS will take responsibility for resolution and customer communication from there)
 * CE routes all feature requests to product for consideration and prioritization
-* If a CE feels compelled to let a customer know CS will help them, they should do so without using loaded terms like “ASAP” or that sign-up CS for something they may not be able to honor
-	* *A member of our support team will help you with this. They typically are able to start digging in within a few hours at most, so you are in good hands! If anything has changed on your side to make this more urgent, let me know and I can be sure they are aware, too.*
```
beyang, line 45: @virginiaulrich lmk what you think of the revised wording and process. My intent is to change the message from "Hey, I'm passing you off to support and they'll get back to you in a few hours while you wait" to "Hey, I'm still with you, but I'm going to loop in support in the next few hours and spend my time bringing them up to speed so they can bring this to resolution as quickly as possible." Whatever the SLA is, I think it's better to help the customer understand that the wait time doesn't mean they've been "dropped" until the next person comes along and picks them up. WDYT?

virginiaulrich:
Agree with 99% of it ... some of what CS brings to Eng are help requests, so I added that. And I have qualms about the word "ownership" so I just tweaked one change slightly to avoid using that word

virginiaulrich (approved):

PR #3453, opened 2021-06-24
Fix security policy wording

andreeleuterio:
Following up with @tammy-zhu's comments that weren't merged previously.


andreeleuterio:
@tammy-zhu I reviewed all comments and these were the only discrepancies I found from the Google Doc and what's on sourcegraph.com/security. Lmk if I missed any!

tammy-zhu (approved):

PR #3463, opened 2021-06-25
Updating license key info.

emchap:
Based on Adam's excellent deep dive at https://sourcegraph.slack.com/archives/CMMTWQQ49/p1624634031301000


LawnGnome (approved):
This looks like it reflects what we're shipping to me. @malomarrec asked me a good question, which was the _first_ version we shipped the licensing behaviour we currently have, and the answer is 3.22.

Thanks for writing this up!

PR #3475, opened 2021-06-29
How to add events to Amplitude

ebrodymoore:
@attfarhan @malomarrec please let me know if these steps are clear on how to add events to Amplitude from the prod/eng side as well as BizOps


ebrodymoore:
@malomarrec this is just adding them to amplitude, not adding the eventLog to the code so he might not find it super useful (just an FYI)

PR #3478, opened 2021-06-29
Add test envs

abeatrix:


DaedalusG (approved):
LGTM, might do additional PRs later to set folks up to address SSH credentials for the was deploy

virginiaulrich (approved):
Made a tiny change and other than that; looks great! Thank you!

PR #3480, opened 2021-06-29
Adding product and engineering demo day documentation

christinaforney:


christinaforney:
Would it be helpful to add a FAQ section? I'm expecting to see some questions such as "how is this different than posting in #progress"?

nicksnyder:
I don't think this is clear from the docs that we want this to be inclusive of small bug fixes, and behind the scenes engineering work that may not be visible to customers, especially since all the examples are splashy new features.

In addition to adding some clarification somewhere, I am wondering if we should consider a different name like "show and tell"

nicksnyder (approved):

beyang (approved):

christinaforney:
@nicksnyder - I tweaked the wording to try and address your point above. I'm going to stick with the name for now and see how folks react/participate before seeing if we need a new name. I'm going to go ahead and merge, and am happy for more feedback and will continue to iterate!

nicksnyder:
updates lgtm, thanks!

PR #3485, opened 2021-06-30
Add Hubspot form for security updates

andreeleuterio:
This PR adds (back) the hubspot form used to subscribe to Sourcegraph security updates


rebecca-rissinger (commented):
LGTM

PR #3507, opened 2021-07-01
Add extensibility history

Joelkw:
Adding this as a point of interest and also useful info for A.M. (Product Director) onboarding. Tried to give some of the timeline/history for the legacy of the team and link out to places that described the reasoning for each decision in more depth, when available. 

Extra history welcome from the folks who might have more context here, like @felixfbecker .

### DIFF
--- a/handbook/engineering/developer-insights/extensibility/index.md
+++ b/handbook/engineering/developer-insights/extensibility/index.md
@@ -130,3 +130,17 @@ Before team syncs, teammates and stakeholders should write down under "Agenda" i
 ### Hack Time
 
 Every week, we spend an hour and a half working on experiments outside of our prioritized lists. Examples of the type of work include extensions we feel strongly about, market intelligence tools, automation scripts, etc.
+
+## Extensibility history 
+
+In March 2016, [Sourcegraph released its first browser extension](https://about.sourcegraph.com/blog/browse-review-code-on-github-like-in-an-ide-with-the-sourcegraph-chrome-extension/) so that you can take the value of Sourcegraph to your code host for code reviews. 
+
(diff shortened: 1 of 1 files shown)


felixfbecker:
Everything here looks correct. I could add a _lot_ more detail and nuance if it's wanted – how much detail do you want though?

Joelkw:
@felixfbecker anything you feel like is particularly interesting to a PM? I would love, for example, failed experiments/directions we chose not to take (and why) perhaps?

muratsu:
Seems like at one point we had a vscode-browser, which was an Electron app. Maybe it was a fork of vscode?

jeanduplessis (approved):

felixfbecker:
There is a lot of really good background information (motivation etc) from Quinn Slack in https://docs.sourcegraph.com/extensions/principles

The VS Code fork was chronologically before the extension API and unrelated (we shut it down again before we started thinking about an extension API).

On the browser extension, I know it was originally written as a growth hack during a hackathon, and grew to one of our products with the most active users (more than sourcegraph.com itself I believe).

The first version of the extension API was also added during a hackathon and is largely inspired by VS Code's. In addition to the motivation in the principles doc above, a major motivating factor was separating language servers out of our core deployment and supporting them through extensions that talk to independent deployments (this is no longer done because of LSIF).

Interesting might also be the RFC that questioned whether we should keep extensions and the explicit decisions (and arguments) that we should: https://docs.google.com/document/d/1KpZh94AdpCgTJ7vyzeVEhBRd9q59glRL695S1RXtyms/edit#heading=h.osgdbz72s2kg

There weren't really any "failed experiments/directions" that I can think of because we have generally not had the resources to invest in them and _try_ any experiments/directions (let alone scrap any and start over). There is lots as far as it relates to code intel (srclib -> LSP deployed natively in Sourcegraph, no extensions -> LSP over WebSockets from extensions, decoupled from Sourcegraph -> LSIF natively in Sourcegraph, queried through extensions), but not for extensions and browser extensions themselves. There are also technical directions/refactors extensions and browser extension went through, but those are less interesting to a Product Director.

PR #3535, opened 2021-07-02
Adding CE to laptop specs

InesRoitman:
@amenne what do you think if we add CE in the title for new teammates to know they are included in that category?


amenne (approved):
Love this!

PR #3564, opened 2021-07-02
Document how Product can get info and trends from Customer Support

Joelkw:
Very often ([1](https://sourcegraph.slack.com/archives/C01HW836940/p1622069711161900), [2](https://sourcegraph.slack.com/archives/C01SD4A49E1/p1624438068019400?thread_ts=1624415946.017900&cid=C01SD4A49E1), [3](https://sourcegraph.slack.com/archives/C01HW836940/p1625242441450500) ) a product team member has noted that support tickets are a valuable source of product feedback if we can have insight into them, and they are right. But instead of asking this in slack, we should document it here. My bad for not doing so sooner – and for not doing so after the first repeat. Had no idea it'd be such a popular question 🙂.

@virginiaulrich I imagine if the process goes well, when you document the quarterly handoff we should link it from here (I'll try to do that but sharing that thought now).  I've also tried to update a few other moments where it's CS instead of CE now but please do correct any mistakes.

### DIFF
--- a/handbook/product/user_feedback.md
+++ b/handbook/product/user_feedback.md
@@ -33,7 +33,7 @@ We have a few different email lists that are used to send us feedback.
 #### support@sourcegraph.com
 
 - **Purpose:** This email list is for customers to request help and connects to Zendesk. 
-- **Owner:** The Customer Support team is [responsible for all support issues](../support/support-workflow#support-workflow) and will pull in as needed engineering teams or product to help answer questions.
+- **Owner:** The Customer Support team is [responsible for all support issues](../support/support-workflow.md#support-workflow) and will pull in as needed engineering teams or product to help answer questions.
 
 ### GitHub issues
 
--- a/handbook/product/user_feedback.md
+++ b/handbook/product/user_feedback.md
@@ -32,8 +32,8 @@ We have a few different email lists that are used to send us feedback.
 
 #### support@sourcegraph.com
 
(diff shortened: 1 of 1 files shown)


virginiaulrich (approved):
You are a step ahead of me; thank you! Yes, I will be creating something for the handbook as part of creating the first summary in August and I will link it here when I do.

I have appreciated that this keeps coming up -- it always feels good to know a report you are going to prepare is desired!

handbook/product/user_feedback.md
```diff
@@ -32,8 +32,8 @@ We have a few different email lists that are used to send us feedback.
 
 #### support@sourcegraph.com
 
-- **Purpose:** This email list is for customers to request help and creates tickets in JIRA for our Customer Engineering team.
-- **Owner:** The Customer Engineering team is responsible for all support issues and will pull in as needed engineering teams or product to help answer questions.
+- **Purpose:** This email list is for customers to request help and connects to Zendesk. 
+- **Owner:** The Customer Support team is [responsible for all support issues](../support/support-workflow#support-workflow) and will pull in as needed engineering teams or product to help answer questions.
```
Joelkw, line 36:
```suggestion
- **Owner:** The Customer Support team is [responsible for all support issues](../support/support-workflow.md#support-workflow) and will pull in as needed engineering teams or product to help answer questions.
```

PR #3567, opened 2021-07-02
Add lists and intro-to-gcp

abeatrix:


virginiaulrich (approved):
So exciting to see so much being added of late!

PR #3568, opened 2021-07-02
Document how to notice and fix if the handbook fails to pass checks

Joelkw:
After hitting red main branches twice in one week, I felt like these small problems were popping up often enough that it's useful to document this part. (All of the "bad" examples in this doc are mine since I don't mind). 

Recent examples of these types of problems from just the past month or so, to validate that this is a problem: [1](https://sourcegraph.slack.com/archives/CNC4YAL1E/p1625103493403400), [2](https://sourcegraph.slack.com/archives/CQ44Y7F4G/p1625261872342200), [3](https://sourcegraph.slack.com/archives/CQ44Y7F4G/p1624442609295200?thread_ts=1624441667.294800&cid=CQ44Y7F4G), [4](https://sourcegraph.slack.com/archives/CQ44Y7F4G/p1615838623022400?thread_ts=1615838336.021000&cid=CQ44Y7F4G), [5](https://sourcegraph.slack.com/archives/CQ44Y7F4G/p1623442515242700?thread_ts=1623271319.235000&cid=CQ44Y7F4G), etc. 

@virginiaulrich I would appreciate your eyes or clarity edits on this, because I've noticed you have great empathy and ability to phrase things for the less-github-familiar folks and I might have missed something important that's familiar to me. 

Someday we might even want to monitor how often we hit failing checks – possibly a responsibility of the future [Handbook PM ](https://boards.greenhouse.io/sourcegraph91/jobs/4027392004)? cc @dadlerj

### DIFF
--- a/handbook/editing.md
+++ b/handbook/editing.md
@@ -20,6 +20,7 @@ Here's the process for getting a change published to the handbook. For detailed 
 
 1. Propose the edits you want to make by creating a [pull request](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-requests) on the Git repository at [github.com/sourcegraph/about](https://github.com/sourcegraph/about).
    - Because the [handbook is public](usage.md#why-make-this-handbook-public), anyone in the world can see your proposed edits.
+1. Make sure the [handbook checks pass](#handbook-checks).
 1. You can request a review from specific teammates (who will get a notification) on the GitHub pull request page.
    - We don't have any rules around who needs to review what changes. Use your judgment (e.g., if your change affects the entire engineering team, request review from the VP Engineering).
    - For minor edits, you can skip review.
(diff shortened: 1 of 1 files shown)


pdubroy:
Have we considered making the checks actually block the merging of PRs?

IMO having optional checks rarely works out well. Inevitably they start failing and there's little incentive to fix the problem, and people get conditioned that it's normal for the checks to fail, so they assume it probably isn't related to their changes.

virginiaulrich (approved):
I think this is written very kindly and helpfully. I am on the fence about the exact best steps for folks to take when something fails. For example, I usually follow-up directly with folks when a check on my PR fails due to their page not being linked (vs fixing it for them) .... so I like that you guide folks to just post in the channel and try to get it resolved before merging. Maybe it's too much nuance to talk about why it's okay if you need to squash and merge to do so ... but that is the only other thought I have reading these great revisions!

Joelkw:
@virginiaulrich Thanks for the feedback! From conversations [I saw in #handbook ](https://sourcegraph.slack.com/archives/C021KFCA5MF/p1625557384058400) it looks like we're moving away from it being okay to squash and merge without fixing, otherwise I'd add guidance there as well.

PR #3688, opened 2021-07-08
a/b testing process updates

ebrodymoore:
@malomarrec @rrhyne proposed some raw thoughts on process improvements. feel free to edit and add

### DIFF
--- a/handbook/ops/bizops/ab-testing.md
+++ b/handbook/ops/bizops/ab-testing.md
@@ -42,6 +42,8 @@ Running A/B tests on a number of users that is too small is inefficient (waste t
 - Evaluating versions of button placement, CTAs, landing pages
 - Evaluating different messaging and positioning statements
 
+The higher the initial conversion percentage, the easier it will be to acheive a sample size quickly, but the tougher it could be to see a meaninful increase. 
+
 #### Bad A/B test candidates
 
 - Launching a new key feature backed by product and UX research: success can be measured directly with an adoption metric (there is an easier way)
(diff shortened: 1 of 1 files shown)


handbook/ops/bizops/ab-testing.md
```diff
@@ -52,19 +54,26 @@ Running A/B tests on a number of users that is too small is inefficient (waste t
...
+### Communicating an A/B test candidate
+
+Communicate good candidates with any relevant parties. This could include other product/engineering teams that the A/B test could impact, such as but not limited to PMs, designers, BizOps. Posting the proposal in a public channel (#product, #product-led, #analytics) could be a good route to do this. 
+
 ### Define the A/B test
 
-- Define the target metric
-- Define the A, B (and more) versions
-- Define the length of the test, depending on the number of users you need, and check for statistical significance with a calculator ([example](https://www.optimizely.com/sample-size-calculator/)). Let BizOps know if you need to understand how much existing traffic there is to determine the expected length of the test
+- Fill out all the following information on the [A/B testing tracker](https://docs.google.com/spreadsheets/d/1BSLrcvnhq-7X9XrsM81MQePYBVjozxB2GWgNmOUdyeI/edit):
```
malomarrec, line 63: There are two sources of truth: the tracker and the issues. This could result in none of them being up to date/confusion. 
What could we do to simplify this?

malomarrec (approved):
Left one comment about duplicate source of truth, would love to resolve that, otherwise LGTM

handbook/ops/bizops/ab-testing.md
ebrodymoore: A lot of information that should be in the source of truth (the numbers like sample size, expected increase, etc, and links to the query) weren't in the issues so I added them to the tracker but now I see the issue template has those. 

In the scenario we get rid of the tracker I would see two issues: 1) the marketing team is a lot more inclined to use Google Sheets than GitHub Issues 2) there's a lot more noise in issues to see an A/B testing schedule. 

For #1 we can handbook around that, but not sure a good solution to #2 other than a kanban board or something. Thoughts?

malomarrec: I'd love to see a board, that would resonate a lot with developers and PMSI think,but I understand your concerns about ease of use. Your call :)

ebrodymoore: You won me over

PR #3695, opened 2021-07-08
Adding info about deleting pages on redirects.

emchap:
@jeanduplessis My read of your description of what happened is that my not deleting the page cause it. I didn't see a clear indication that I ought to delete it in the instructions I was referred to—does this look okay/is it a correct understanding?


jeanduplessis (approved):

jeanduplessis:
@emchap I think this is a good addition. I didn't actually even know this was needed and only figured it out today as well 🤷‍♂️ .

emchap:
@jeanduplessis That comforts me. 😅 Sorry for the hubub! Thanks for sorting it out.

PR #3701, opened 2021-07-09
Dev Tool Time: Leah Culver recap

rebeccadee:
https://app.asana.com/0/1199959077348208/1200063735031308/f

### DIFF
--- a/blogposts/2021/dev-tool-time-leah-culver.md
+++ b/blogposts/2021/dev-tool-time-leah-culver.md
@@ -1,0 +1,91 @@
+---
+title: "'If you're still using Google Docs, what are you doing?!' Dev Tool Time with Leah Culver"
+description: "Leah Culver, iOS Developer at Twitter, joined us to share some developer life hacks, her minimalist, no-config philosophy for tools, and her secret for reaching inbox zero."
+author: Vanesa Ortiz
+authorUrl: https://twitter.com/vanesacodes
+publishDate: 2021-07-09T10:00-07:00
+tags: [blog]
+slug: dev-tool-time-leah-culver
+heroImage: https://sourcegraphstatic.com/blog/dev-tool-time/dev-tool-time-leah-culver.jpg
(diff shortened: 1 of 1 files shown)


rebeccadee:
@content-hacker @jennybergen I've taken a crack at this based on a draft Vanesa had started. I will ask for her eyes on it but she's out today! A few things I've tried here:

1. Based on feedback from [Slack](https://sourcegraph.slack.com/archives/C01NBPU6VFX/p1625737456061600), it sounds like teaser videos would be a great way to drive traffic to the video or recap, so I created one [here](https://drive.google.com/file/d/1O5eFU9P1FDlTgcdtlxmXlwt97odf3ui4/view?usp=sharing) which I'll share with the relevant parties in https://app.asana.com/0/1199959077348208/1200570738956554/f. 
2. Made the title a quote – hoping this is more attention grabbing than the formulaic one
3. I added in some pull quotes to give it more personality, so you can hear Leah's voice throughout

It sounds like there's a lot more we can do with DTT content (see https://app.asana.com/0/0/1200570738956575/f) but I'm trying to move quickly and do what I can here that doesn't require help from other teams in the interim. Would appreciate another set of eyes on this 💜

rebeccadee:
Hey @vanesa! Could you please take a look at this, which should be ready to publish 😊 You can see some of my thinking behind some changes I made [in my comment above](https://github.com/sourcegraph/about/pull/3701#issuecomment-877154253), let me know what you think!

(Also just noting for visibility I got a 👍 from Erica on this in our 1:1 earlier)

blogposts/2021/dev-tool-time-leah-culver.md
```diff
@@ -0,0 +1,91 @@
+---
+title: "'If you're still using Google Docs, what are you doing?!' Dev Tool Time with Leah Culver"
```
vanesa, line 2: While I thought this comment was funny, from a dev perspective it's better to not lead with google docs as it seems not as dev-focused. This comment also comes across as negative rather than positive. Could we pick something else?

vanesa (changes requested):
Other than the title, I think it looks great! Thank you!

PR #3703, opened 2021-07-09
Change significance calculator

malomarrec:
There were issues with Optimizely's cohort size calculator. We are standardizing on another calculator for sample size. We are also standardizing on one calculator for significance, so comparing results is easy.


rrhyne (approved):
LGTM!

PR #3706, opened 2021-07-09
Add explicit funnel metrics for Sourcegraph cloud

malomarrec:
Add metrics to have clarity into the acquisition/activation/retention stages on the user journey funnel on sourcegraph cloud, as proposed in RFC 407: Explicit standard metrics for Sourcegraph.com


handbook/ops/bizops/user_definitions.md
```diff
@@ -46,17 +44,43 @@ Additional Resources:
...
+| API docs                  | TBD                                | TBD                                                                                                 |
+
+
+## How are metrics calculated
+Our metrics infrastructure (Looker) gets user counts from our event_logs table (either directly or indirectly). We use the same method as on-premise for calculating user counts, pulling from the Sourcegraph Cloud `event_logs` table. We track unauthenticated users using their `anonymous_user_id`. This is a separate column which contains an anonymous ID, which is stored in a cookie for users that visit Sourcegraph.com. Therefore, for all charts that track Cloud active users, unauthenticated users are included in these counts. 
+
+There are shortcomings to this. For one, when a user converts into an authed user, their events conducted with their anonymous user ID are still in the DB, so we would count two different users being active rather than a single user.For analytics purposes in Amplitude, this is also not ideal because we are not able to connect the actions of a user before and after they've converted.
```
ebrodymoore, line 85:
```suggestion
There are shortcomings to this. For one, when a user converts into an authed user, their events conducted with their anonymous user ID are still in the DB, so we would count two different users being active rather than a single user. For analytics purposes in Amplitude, this is also not ideal because we are not able to connect the actions of a user before and after they've converted.
```

ebrodymoore (approved):

PR #3736, opened 2021-07-13
Update company spending types and travel phase II

cassiemelani:
Updating company spending guidelines, introducing subsections of budgets and best practices for spending, and moving into Phase II of Teammate Travel Reopening

### DIFF
--- a/handbook/people-ops/travel.md
+++ b/handbook/people-ops/travel.md
@@ -1,98 +1,166 @@
 # Company sponsored events (team and company-wide)
 
 
-## Phase I: Teammate Travel Meetup Guidelines
+# Phase II: Teammate Travel Meetup Guidelines
 
-As of June 16th 2021, we will resume reimbursing travel per our [team travel budgets](https://about.sourcegraph.com/handbook/people-ops/travel#team-travel-budgets) for travel and in-person events that are permitted by law. We want this to be gradual, and all events must be small and completely optional. 
+As of June 16th 2021, we have resumed reimbursing travel per our  [team travel budgets](https://about.sourcegraph.com/handbook/people-ops/travel#team-travel-budgets)  for travel and in-person events that are permitted by law. We want this to be gradual, and all events must be completely optional.
 
-As an all-remote company, we want everyone to feel included and connected, regardless of where they're located. Choosing to travel or meet is an incredibly personal decision, and we support team members who may not be interested in meeting up at this time.
+As an all-remote company, we want everyone to feel included and connected, regardless of where they’re located. Choosing to travel or meet is an incredibly personal decision, and we support team members who may not be interested in meeting up at this time.
 
+## Guidelines
 
-## **Guidelines:**
-
-- To avoid the perception that these events are mandatory, managers should not organize events for their teams. 
-   	- Managers can attend events and organize events that don't involve their direct reports.
-
--   Events should be outdoors rather than indoors.
+ -  All individual teammates can continue to plan and coordinate events.    
+ - Managers may now plan meetups
+	- Managers must actively gauge the level of comfort within their teams for both the activities you’ll be engaging in, and how you plan on meeting up. *Be inclusive of different interests in the group.*
+	- Managers must be aware that not everyone on your team may be able to join your meetup. If you will be discussing work related things that involve members of the team who are not physically present, you must find a way to include them remotely. *Accessibility to conversations is important, and all teammates must feel included, whether or not they are joining in person.*
+ 
+ - All meetups will continue to remain completely optional for all attendees.
     
--   Keep gatherings small. Use your best judgement and defer to local jurisdictions in determining the number of attendees.
+-   Use your best judgement and defer to local jurisdictions in determining the number of attendees.
     
 -   Follow local laws, customs, and safety regulations during every step of the meetup process.
-
-- Assume people don't want to shake hands, hug, high-five, etc. 
     
--   Don't ask about other people's private health choices or condition attendance on those factors (*such as vaccination status*).
+-   Assume people don’t want to shake hands, hug, high-five, etc.
     
-***Please notify People Ops (people-ops@sourcegraph.com) before you start planning an event.***
+-   Don’t ask about other people’s private health choices or condition attendance on those factors (_such as vaccination status_).
     
 
-## How we came to this decision:
+**_Please notify People Ops (people-ops@sourcegraph.com) before you start planning an event._**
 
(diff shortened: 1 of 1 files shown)


mercadon (approved):

kmorris50 (approved):

handbook/people-ops/travel.md
```diff
@@ -1,98 +1,166 @@
...
+Sourcegraph is an  [all-remote](https://about.sourcegraph.com/handbook/company/remote)  company. This means we all need to make an effort to travel to meet up in person and spend quality time with other team members all around the world.
 
-Team meetups are very important for building cohesiveness in remote teams and spreading Sourcegraph culture across the company.
+Team meetups are very important for building cohesiveness in remote teams and spreading Sourcegraph culture across the company. 
 
-We want and expect you to use your full travel budgets described here each year! For this reason, they aren't transferrable, and unused budget doesn't roll over into the next year.
+Be economical in choosing your accommodations, modes of transportation, and where you’ll be dining. We understand that there may be limits, and we urge you to find options that are both convenient for you while being reasonable reimbursements for the company. We trust you to use your best judgement while spending company money.
```
LawnGnome, line 42: This feels somewhat at odds with the more specific guidance further down around accommodation cost.

Practically speaking, the last place I worked with similar verbiage had problems with people interpreting this kind of advice _really_ differently: a fresh uni grad who's used to backpacking will probably have a very different idea of "economical" than a professional who's been travelling for work for decades, and that's without even getting into cross-cultural differences.

I understand the reasons why we don't just want to say "your budget is $400 per night; go" up front, but we might need more detail here around how to choose.

handbook/people-ops/travel.md
```diff
@@ -1,98 +1,166 @@
...
 
-Sourcegraph generally holds two company-wide events each year, each fully covered by the company:
+# Budget best practices and additional spending categories
 
-- Winter party (typically in San Francisco)
-- Spring/summer event
+## Hotel accomodations:
 
-## Team events
+We advise spending within reason for the location you are visiting, with a $400 USD per night maximum spend limit.
```
LawnGnome, line 107: At my previous employer, we ended up developing more fine grained policies based on the cost of living/accommodation in certain high cost cities. ($400 a night can be problematic in the Bay Area; meanwhile you've probably just rented an entire town in the Western Australian wheatbelt.) We may want to consider doing the same here eventually, as it helps manage expectations.

handbook/people-ops/travel.md
```diff
@@ -1,98 +1,166 @@
...
+> be the most economical option given the surrounding  hotels. Finance
+> will reply with an approval or request to book other  accommodations.*
 
-## Individual travel
+## Modes of transportation:
 
-- Each teammate has one fully-paid trip per year to meet up with a group of teammates in another location.
-  - This is *in addition* to the other benefits mentioned on this page.
-  - The purpose of defining this as one “trip” (instead of providing a set dollar budget) is to avoid punishing teammates that live far away from other teammates, and who would therefore incur higher airfare costs.
+ - Economy class is recommended for booking all travel. *(If you need any higher tier of booking, reach out to Finance/People Ops for approval.)*
```
LawnGnome, line 120: We might not be here yet in terms of our funding story, but one policy I came to appreciate at previous employers was being allowed to book the next highest class (usually premium economy, sometimes business) for longer trips: trans-continental travel is draining (I'm Australian; ask me how I know), trips are often cut down to the minimum possible time for both personal and business reasons, and few people do well with going straight into work meetings after a sleepless, overnight 12 hour flight in economy with a 10 hour time zone difference.

If we did this, I've seen different thresholds here. Six hours was the threshold at my previous employer, which mysteriously corresponded to the length of a specific flight our CEO tended to take often. I'd suggest basing it on time zones crossed instead, with a fallback for extremely long north-south flights.

handbook/people-ops/travel.md
```diff
@@ -1,98 +1,166 @@
...
+ - Ride-shares (Uber, Lyft, etc.) are encouraged whenever possible instead of renting a car.
+	 - If you’re in the US and would like to rent a car, we recommend checking with the following companies before booking your car rental elsewhere:
+		 - [Turo](https://turo.com/)  
+		 - [Getaround](https://www.getaround.com/)
+		 - [Zipcar](https://www.zipcar.com/)
 
+## Dining and your per diem:
 
-## Professional development and education
+   We ask you to use your best judgment when traveling and spending company money. Don’t go overboard with spending on food and drinks. Always ask for receipts and keep them organized.
```
LawnGnome, line 133: Could we consider a threshold below which we don't require receipts? Non-US teammates don't use Expensify, which makes receipt processing easier, and organising one's receipts for that €10 coffee and croissant at ~Charles de Gaulle~ a poorly organised airport because you haven't eaten for eight hours because ~Air France~ a poorly organised airline while you're simultaneously trying to make your connecting flight after being delayed at immigration isn't much fun.

handbook/people-ops/travel.md
```diff
@@ -1,98 +1,166 @@
...
     
--   Keep gatherings small. Use your best judgement and defer to local jurisdictions in determining the number of attendees.
+-   Use your best judgement and defer to local jurisdictions in determining the number of attendees.
     
 -   Follow local laws, customs, and safety regulations during every step of the meetup process.
-
-- Assume people don't want to shake hands, hug, high-five, etc. 
     
--   Don't ask about other people's private health choices or condition attendance on those factors (*such as vaccination status*).
+-   Assume people don’t want to shake hands, hug, high-five, etc.
```
LawnGnome, line 23: I really like this one. (I'm resisting the urge to suggest a Sourcegraph specific greeting dance.)

LawnGnome (commented):
I'm excited to see travel come back again!

As my passport will sadly attest, I've spent a fair bit of time travelling for work over the years: I have some suggestions for things that we could tweak that would help road warriors. Unfortunately, most of them involve extra expense, so I understand if we don't believe it's in Sourcegraph's best interests to do them right now, but I'd like to put them on the record, at least.

(Not approving because I don't consider myself an approver here, but I do :+1: the changes.)

handbook/people-ops/travel.md
cassiemelani: Great point @LawnGnome - we've adjusted to say receipts above $25 which is consistent with our policy in Expensify. Thanks for flagging this!

handbook/people-ops/travel.md
cassiemelani: Our Finance/People team connected and also agree we're not necessarily here yet in terms of our funding story :) Thank you for noting this, and we'll consider for future iterations of company spending revisions!

PR #3761, opened 2021-07-15
Update guidelines for single/double quote usage

quinnkeast:


handbook/communication/content_guidelines/style_and_mechanics.md
```diff
@@ -557,6 +557,25 @@ Use periods in complete sentences. Don’t use periods in headlines or labels. P
 
 ### Quotation marks
 
+When quoting content, use double quotation marks. Use single quotation marks to quote content within an existing quote.
```
felixfbecker, line 560: Might be worth giving this tip?
```suggestion
When quoting content, use double quotation marks. Use single quotation marks to quote content within an existing quote.<br>
In HTML, using `<q></q>` will automatically apply the right quotation marks for nested `<q>` tags.
```

handbook/communication/content_guidelines/style_and_mechanics.md
```diff
@@ -557,6 +557,25 @@ Use periods in complete sentences. Don’t use periods in headlines or labels. P
...
+<div class="item no">
+<h5>No</h5>
+<ul>
+<li>Add an insight to ‘Test Dashboard’</li>
+<li>“When quoting inside existing quotes, avoid “double quoting””</li>
+</ul>
+</div>
+</div>
+
 Watch out for dumb quotation marks (`'` and `"`). These are a relic of typewriters, and can be identified by how they’re straight rather than curly. Always use typographic quotation marks (`‘` and `’`, `“` and `”`).
```
felixfbecker, line 579:
```suggestion
Watch out for dumb quotation marks (`'` and `"`). These are a relic of typewriters, and can be identified by how they’re straight rather than curly. Always use typographic quotation marks (`‘` and `’`, `“` and `”`).<br>
In HTML, the easiest way to add the appropriate typographic quotation marks is to wrap the quote in `<q></q>` instead of manually quoting.
```

felixfbecker (approved):
Thank you!!!

handbook/communication/content_guidelines/style_and_mechanics.md
quinnkeast: Ooh, good call. Just made one extra tweak for clarity.

```suggestion
When quoting content, use double quotation marks. Use single quotation marks to quote content within an existing quote.
In HTML, using `<q></q>` will automatically apply the correct quotation marks for nested `<q>` tags.
```

PR #3779, opened 2021-07-16
Add interviewing option for contributing to blog

rebeccadee:
Clarifies different ways team members can contribute to the blog, including being interviewed if pressed for time or less confident with writing.


rebeccadee:
FYI @content-hacker @jennybergen I got some feedback that people didn't know this was an option so calling it out here 🤗

PR #3788, opened 2021-07-17
Document what "beta" and "experimental" actually mean, beyond the customer-facing support explanation doc

Joelkw:
[This captured](https://docs.sourcegraph.com/admin/beta_and_prototype_features) what customers could expect but doesn't convey what for and why we use words like "Beta." As code insights approaches beta, we've been getting more questions and confusion as to what that actually means and how it's different conceptually from a prototype. 

> **Clarifying note**: we decided to align on "experimental" instead of "prototype." The product team will be making the cascading updates as we can to align here – I'm aware the docs page I link to above says "prototype" as of now.


handbook/product/beta_and_experimental_feature_labels.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Why we use these labels:
+
+Labelling features correctly lets us move quickly and release early in order to validate our direction. An `Experimental` or `Beta` label and its [corresponding user-facing documentation](https://docs.sourcegraph.com/admin/beta_and_prototype_features): 
+  
+  - Set expectations so users are more forgiving of bugs
+  - Make users less likely to assume the overall quality of Sourcegraph is as brittle as an early feature might be
+  - Communicate that we are particularly interested in user feedback
+  - Communicate that a user's feedback is more likely to have meaningful impact due to the early stage
+  - Let us release paid-tier features initially for free, to usage feedback fast and build our sales and marketing knowledge
```
Joelkw, line 15:
```suggestion
  - Let us release paid-tier features initially for free, to get usage feedback fast or to build our sales and marketing knowledge
```

handbook/product/beta_and_experimental_feature_labels.md
```diff
@@ -0,0 +1,61 @@
+# Beta and experimental feature labels
+
+We use a hierarchy of feature labels to set expectations with our users and internally: `Experimental` << `Beta` << GA.
```
jeanduplessis, line 3: nit (non-blocking): The arrow directions really messed with my head.

```suggestion
We use a hierarchy of feature labels to set expectations with our users and internally: `Experimental` ➡️ `Beta` ➡️ GA. 
```

handbook/product/beta_and_experimental_feature_labels.md
```diff
@@ -0,0 +1,61 @@
+# Beta and experimental feature labels
+
+We use a hierarchy of feature labels to set expectations with our users and internally: `Experimental` << `Beta` << GA. 
+
+GA ("General Availability") has no in-product label and is the default user assumption for an unlabelled feature.
```
jeanduplessis, line 5: Do we have a better (internal maybe?) resource to link to what GA means?

```suggestion
[GA ("General Availability")](https://www.productplan.com/glossary/general-availability/) has no in-product label and is the default user assumption for an unlabelled feature. 
```

handbook/product/beta_and_experimental_feature_labels.md
```diff
@@ -0,0 +1,61 @@
...
+  - Communicate that a user's feedback is more likely to have meaningful impact due to the early stage
+  - Let us release paid-tier features initially for free, to get usage feedback fast or to build our sales and marketing knowledge
+
+## When to use each label: 
+
+There's no requirement that every feature begin as `Experimental` and pass through `Beta` before being in `GA`. This depends on the feature size, the development time, and the cost of getting a GA release wrong. 
+
+> While there is no strict required lifecycle, for the sake of our customers and our customer-facing teams, we **do** adhere to the minimum documentation, feature flag, and support requirements [defined in our docs](https://docs.sourcegraph.com/admin/beta_and_prototype_features). 
+
+> For example, you can't label a feature that's enabled by default `Experimental`, nor a feature with no documentation a `Beta`, but you can have the full necessary for `Beta` and still choose to use `Experimental`.
```
jeanduplessis, line 23: API Docs are enabled on sourcegraph.com by default and is labeled as Experimental which violates this example 🤔

handbook/product/beta_and_experimental_feature_labels.md
```diff
@@ -0,0 +1,61 @@
...
+- Is this feature at a quality you would be comfortable enabling it for all users by default? 
+- Does the feature communicate its own value without high-touch involvement? 
+- Are users more likely to give feedback on the "how I interact with this" or "I *also* want" rather than the "what is this" level? 
+- Are we ready to collect broad usage feedback in order to build our sales, pricing, or marketing strategy? 
+
+## When we use an entirely different label
+
+**Rarely**, sometimes we use a different label when a feature is in `Beta` in order to benefit from limiting initial rollout without invoking users' connotations with the word "Beta." 
+
+For example: a feature may be ready for GA in on a security and functional level, but we want to test the UX before entering GA. We don't want users to incorrectly assume the security isn't robust by attaching a `Beta` label, but we do need to communicate limited availability and possible UX polishing. Traditionally, we use the term `Early Access` for these cases.
```
jeanduplessis, line 61:
```suggestion
For example: a feature may be ready for GA on a security and functional level, but we want to test the UX before entering GA. We don't want users to incorrectly assume the security isn't robust by attaching a `Beta` label, but we do need to communicate limited availability and possible UX polishing. Traditionally, we use the term `Early Access` for these cases. 
```

jeanduplessis (approved):
Related thought: Should we consider linking our in-product labels to this new page in the handbook?

macraig (approved):
I think this captures where the line between beta and experimental is in an easy and straightforward way, thanks @Joelkw for documenting!

handbook/product/beta_and_experimental_feature_labels.md
macraig: +1 to this small change

Joelkw: Haha fair point.

handbook/product/beta_and_experimental_feature_labels.md
Joelkw: I don't think so, at least not yet. Will add to my backlog of possible docs to add and go with this for now.

handbook/product/beta_and_experimental_feature_labels.md
Joelkw: @jeanduplessis that'd actually be a bit "worse" than violating this example – it violates our live product docs https://docs.sourcegraph.com/admin/beta_and_prototype_features. 

But I'm not sure that's actually the case here. Are API Docs a cloud-only feature? If so, perhaps that docs page should except cloud features. If they're enabled by default on customer instances, then yes, I think we need to decide why that is (are they beta-level? If not, why are they enabled by default? Does that mean we need to reexamine the logic of this page?)

handbook/product/beta_and_experimental_feature_labels.md
ryphil: +++ to this change

ryphil:
All looks good to me!

ryphil (approved):

Joelkw:
> Related thought: Should we consider linking our in-product labels to this new page in the handbook?

@jeanduplessis Yes, but I think we should link to our customer-facing (and written for the customer audience) docs rather than our handbook page about this: https://docs.sourcegraph.com/admin/beta_and_prototype_features . I thought there was a longstanding ticket but not sure I see one – would that be a frontend platform area? [Here's a ticket for that.](https://github.com/sourcegraph/sourcegraph/issues/23001)

jeanduplessis:
Merging this PR as is because of `merge-on-any-approve` label. 

@Joelkw I think we can address the difference between Cloud and Customer instances with regards to enabled by default in a follow-up. If we are going to treat them differently we should call it out.

Frontend Platform can take care of linking the in-app labels.

PR #3819, opened 2021-07-20
Add link to the security policy to customer data policy

andreeleuterio:
Hey Eric, minor update to the doc you shared. Linking to the Security policy works best here IMO because we are much more likely to keep sourcegraph.com/security updated regularly. I added a `both at-rest and in-transit` bit that may be sufficient for most folks.


ryphil (approved):
Cool with me! Thanks for making this update, @andreeleuterio

PR #3888, opened 2021-07-26
Added Michael's readme.md file

topebali:
This is to add Michael's Bio to the readme page


virginiaulrich (changes requested):
@topebali you will need the file and the reference to the file in the same PR to get the tests to pass and be able to merge this. @rhiam (Mariam) also encountered this. She can walk you through how to get this into a good state!

PR #3892, opened 2021-07-27
distribution(managed): add more detail on preparing customer account

bobheadxi:
Messed up this process a little (https://github.com/sourcegraph/customer/issues/410) - email don't get sent because we don't have SMTP on managed instances set up by default (doh!). Also I originally took these handover steps to mean handing over the `sourcegraph-admin` account, which seems to not be the case.


daxmc99 (approved):
LGTM
It's been a second since I did this but do we still specify to use an email address of managed+$CUSTOMER@sourcegraph.com ? 
We should also make sure that is documented

bobheadxi:
@daxmc99 

> do we still specify to use an email address of managed+$CUSTOMER@sourcegraph.com ?

Yep! This is in two places now (I added one recently):

image

and 

image

emchap:
@bobheadxi Looks great, thank you!

PR #3896, opened 2021-07-27
Bring handbook design more in line with our Brand Guidelines

felixfbecker:
This makes our handbook styles follow the Brand Guidelines more closely. Concretely:

- Use our font (Source Sans/Code)
- Use our color palette
- Update all logos
- Tweak font sizes and line heights (I tried to stick to the guidelines, but had to make some exceptions for some headings because content didn't fit well)

Other changes:
- Remove dead CSS I noticed (if you see deleted CSS in the diff, it was never used, there is CSS later in the file overriding it)
- Use SVG favicon
- Make logo a bit larger
- Make `#` when hovering a heading scale to font size of heading
- There are also some autoformatting/consistency changes (I would really love to run Prettier over the whole repo sometime soon and set some automatic GitHub action up)

This is a bit of a prerequisite for my PR to add the career levels because I want those to look good, but extracted this part into its own PR.

Before:
image

After:
image


eseliger (approved):
Nice 🌟

PR #3909, opened 2021-07-27
Fix images, release-3.30.md

ltagliaferri:


ltagliaferri:
Discussed with Andy and shared deploy preview. Video can use additional work to make responsive (perhaps an `<iframe>` solution instead).

PR #3928, opened 2021-07-28
Update Outbound Workflow

kevin-quigley:
Added general principles and workflow videos


kevin-quigley:
This addition adds information on the sales outbound practices

PR #3980, opened 2021-08-02
Fix API Docs URL

abschu:


rebeccadee:
Thank you so much for fixing this @abschu! I think this was a case of the link working locally but not in production?! 🤔 

I don't want to be a blocker for quick fixes like this so feel free to just merge typo/link fixes in the future 😁

PR #4028, opened 2021-08-04
Adding source of truth

InesRoitman:
This might be a bit obvious but I was surprised not to find it there. Should we include it? @dadlerj


malomarrec:
Recursive source of truth definition. 🤯

PR #4039, opened 2021-08-04
add note to create projects in terraform

davejrt:
We should not be creating projects manually, and instead using terraform.


daxmc99 (approved):

bobheadxi:
I missed this PR, my bad - but maybe we should go with https://github.com/sourcegraph/about/pull/4041 instead? I added a link directly to the variable

davejrt:
Closing in favor of #4101

PR #4041, opened 2021-08-05
managed: create instance project in terraform

bobheadxi:
TIL https://github.com/sourcegraph/infrastructure/pull/2710, that explains why the instance I created recently wasn't in the right project 😅


daxmc99:
I think we should prefer https://github.com/sourcegraph/about/pull/4039/ for this

bobheadxi:
@daxmc99 ah sorry, missed that!

davejrt (approved):

PR #4058, opened 2021-08-05
Update people-ops.md

CecilyBee:
Added Officengine's Sourcegraph support request form for U.S. based teammates to use to ask questions regarding health insurance, dental insurance, vision insurance, life insurance, 401k, and payroll (including tax withholdings)


cassiemelani (commented):
question I have in reading through this -- it says "contact our back office directly" for xyz things, but then it also says to contact People Ops directly about the same things. is there some differentiation we want between when people reach out to office engine vs. us, or is the bottom part more for International teammates? thank you!!

CecilyBee:
@cassiemelani That was a mistake! It is supposed to list "PTO/leave, family planning support, professional development, travel, and moving"

PR #4084, opened 2021-08-09
update steps to create managed instances

davejrt:
Updates the steps to create a managed instance with the changes in https://github.com/sourcegraph/deploy-sourcegraph-managed/pull/84


bobheadxi (approved):
😍

PR #4096, opened 2021-08-10
Sponsorship policy edits

mercadon:


handbook/people-ops/how-we-engage-talent-outside-the-us/index.md
```diff
@@ -80,7 +80,7 @@ For candidates currently on a visa who prefer to keep their status in a particul
 
 For candidates in the US in need of sponsorship and who have an existing visa, we will accept a visa transfer from the former employer and sponsor the existing visa for the candidate so long as the visa is compatible with an all-remote work setting.
 
-For candidates in need of US sponsorship that require a new visa, we will agree to sponsor new visas for teammates so long as they have worked at Sourcegraph a minimum of 12 consecutive months prior to initiating a visa petition. 
+For candidates in need of US sponsorship, whether that requires initiating a new visa process or a permanent residency (green card) process, Sourcegraph will agree to initiate and sponsor a new visa or green card process for teammates so long as they have worked at Sourcegraph a minimum of 12 consecutive months prior to initiating a the visa or permanent residency petition.
```
beyang, line 83:
```suggestion
For candidates in need of US sponsorship, whether that requires initiating a new visa process or a permanent residency (green card) process, Sourcegraph will agree to initiate and sponsor a new visa or green card process for teammates so long as they have worked at Sourcegraph a minimum of 12 consecutive months prior to initiating the visa or permanent residency petition. 
```

beyang: What if they have an existing green card process in motion with their former employer?
* Will they have to wait 12 months before resuming the process at Sourcegraph?
* Will they need to restart that process upon switching employers? (Looks like there are some legal considerations there it might be worth flagging: https://www.nolo.com/legal-encyclopedia/can-change-employers-while-green-card-application-pending.html)

mercadon: Just added more language to clarify this a bit further, along with the differences for personal versus role required sponsorship and case-by-case decisions for hard-to-fill roles.

tammy-zhu (approved):

handbook/people-ops/how-we-engage-talent-outside-the-us/index.md
```diff
@@ -73,14 +73,21 @@ Hiring managers should consult the Talent and People Ops team early if an except
...
+### For personal reasons
+Since our teammates can generally work from anywhere in the world, we do not cover sponsorship and visa assistance costs for candidates and teammates who plan to move to or remain in a country where they require sponsorship or immigration assistance. However, if there is a role that is considered difficult to fill and a person posseses a unique skillset and/or particular experience level that is needed, the People Team, Talent Team, and Hiring Manager will make a decision on a case by case basis and may consider covering costs in some cases.  
 
-For candidates in need of US sponsorship that require a new visa, we will agree to sponsor new visas for teammates so long as they have worked at Sourcegraph a minimum of 12 consecutive months prior to initiating a visa petition. 
+For candidates or teammates that currently hold a visa and prefer to keep their status in a particular country for personal reasons, Sourcegraph may be able to sponsor the visa depending on whether we or our partner PEO have the ability to support it. The candidate or teammate would carry the cost of transferring and sponsoring the visa.
```
felixfbecker, line 84: > The candidate or teammate would carry the cost of transferring and sponsoring the visa.

Curious how would this work in practice?
So for the US for example, Sourcegraph would be the "sponsor" that initiates the case through Sourcegraph's immigration law firm with USCIS. Would Sourcegraph then forward every invoice from the lawyers (every billed hour, every filing fee, etc) to the new hire for them to pay? Or would perhaps the costs be deducted from their paychecks? Would it even be possible for the candidate to know how much it will cost before going into it, to not accidentally pile on a mountain of bills they can't afford?

felixfbecker (commented):
Thank you @mercadon, this is definitely clearer than before!

handbook/people-ops/how-we-engage-talent-outside-the-us/index.md
mercadon: @felixfbecker We're working with Wolfsdorf and Remote to try to get a ballpark range, but the short answer is that we likely won't know upfront since it will depend on the countries, visas, and personal circumstances involved with each case. It is something we'd have to discuss at the point in the recruiting process where candidates disclose their sponsorship needs and details. More to come on the payment process as well, we are still working with finance to determine the best way to do this across platforms and employment structures. 

For now I plan to merge these changes and will continue adding more information (to answer questions like the ones you've raised) as we get it!

PR #4136, opened 2021-08-12
Async communication page

InesRoitman:


handbook/company/asynchronous-communication.md
```diff
@@ -0,0 +1,152 @@
...
+- **Set timing expectations**. Add parameters like response time and due dates to ensure progress.
+- **Give yourself time**. Give yourself time to adjust to this new way of working - and to do this type of work. Thoughtfully crafting, reading, and responding to async communication takes longer (in the upfront) than the rapid back-and-forth of real-time conversation. 
+- **Synchronous meetings.** Consider synchronous meetings when needed. Synchronous meetings are not the enemy and can provide a ton of value. Make sure to be inclusive and respectful of other people's time when you schedule them.
+
+Each team should also have a section to explain how they do x, y and z asyncly. Define processes and communication channels. For example, how does the design team work? What is the decision making process? 
+
+
+## How to choose sync vs async?
+
+<img width="60%" height="60%" src="onepager-async.jpg">
```
felixfbecker, line 119: Is there a chance this could be exported as an SVG instead of JPG? It has a lot of text content that won't be readable for screen reader users (but would be if it was an SVG)
What design tool was it created in?

PR #4138, opened 2021-08-12
Debugging cheat sheet

airamare01:
Creating debugging cheat sheet doc in support handbook as an enablement item


adeola-ak:
do you think we should wait until we have some stuff on the page?

airamare01:
@adeola-ak I was thinking we could be updating it as we go? What do you think?

bobheadxi:
Would this be valuable to have as information in pages like e.g. https://docs.sourcegraph.com/admin/install/docker-compose/operations and https://docs.sourcegraph.com/admin/install/kubernetes/operations#list-pods-in-cluster instead? Then admins attempting to troubleshoot themselves can discover it too, and it's easier to find and share - we already have debugging/troubleshooting docs in a lot of different places 🤔

airamare01:
Hey @bobheadxi.
That's a great point you bring up. The intended content for this page is an A-Z step process for specific issues CSEs, admins can take to resolve a particular issue. We were looking at curating all the steps we took to solve specific problems referenced from tickets that we've handled in the past.

It's a great idea to have them in the links you've posted. Maybe @adeola-ak and @amberfurbush0317 have additional insights in case I missed anything on intended purposa

adeola-ak:
@bobheadxi hi! yes we will put it on those pages. I think we are using it in our tools pages right now as it is being worked on but upon completion its meant to be used internally and for our customers as well

PR #4156, opened 2021-08-13
Improve incident identification documentation

quinnkeast:
Our incident identification and resolution process is well-crafted and documented. I'm proposing some small changes to:

- Reduce friction around "pointing out smoke" by explicitly noting that false positives are a **good thing**, because they help us to improve (something lead to that false positive, and so we can probably make something better).
- Make the first action in triage and from incident leads to "thank the person who identified the incident." This is based on [Toyota's andon cord process](https://taylorpearson.me/interestingtimes/andon-cord/), where the team lead's first action upon reaching the person who reported an incident is to thank that person. This is a cultural reinforcement objective, aimed at reducing friction for reporting incidents. Folks should never feel hesitant to identify an incident, and we should always recognize and thank folks when they do. By doing this, we'll avoid cases where folks don't feel confident "pointing out smoke."


handbook/engineering/incidents/index.md
```diff
@@ -1,90 +1,104 @@
...
-The incident responder will need to select a Sourcegraph.com account to attach the subscription to (typically the account should belong to the customer, so they can access the license key directly from their user profile, but in an emergency, the incident responder can use their own account in lieu of asking the customer), and can then manually generate a license key. No license "tags" are necessary.
-
-If a customer's instance is reporting "license expired" already, note that [there is a 72hr grace period](https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/enterprise/internal/license/license.go#L43:15) before non-admin users are locked out.
-
-## Incident Lead
+### Incident Lead
 
 The indicent lead is responsible for resolving the incident as quickly and safely as possible. They are the DRI coordinating the incident, tasked with driving it to resolution and ensuring clear internal and external communication with stalkholders and customers.
 
+1. **Thank** the person who identified the indicident.
```
quinnkeast, line 69:
```suggestion
1. **Thank** the person who identified the incident.
```

handbook/engineering/incidents/index.md
```diff
@@ -1,90 +1,104 @@
...
 1. **Critical** - Issues causing very high impact to customers. Immediate response is required. Examples include a full outage, or a data breach.
 2. **Major** - Issues causing significant impact. Immediate response is usually required. We might have some workarounds that mitigate the impact on customers. Examples include an important sub-system failing.
 3. **Minor** - Issues with low impact, which can usually be handled within working hours. Most customers are unlikely to notice any problems. Examples include a slight drop in application performance.
 
-## Triage
+### Triage
 
 The goal of triage is to either quickly resolve the incident using basic procedures, or quickly identify the right owner.
 
+1. **Thank** the person who identified the indicident.
```
quinnkeast, line 48:
```suggestion
1. **Thank** the person who identified the incident.
```

unknwon (approved):
I don't know if `handbook/engineering/incidents/incident-flow-chart.png` should be included, but doesn't seem like so because we link the image to Google Drive.

quinnkeast:
> I don't know if `handbook/engineering/incidents/incident-flow-chart.png` should be included, but doesn't seem like so because we link the image to Google Drive.

I wasn't quite sure – the original was an `iframe` for a diagrams.net flow chart, which wouldn't show in at least Firefox. I figured turning it into an image linked to the diagram would be a good compromise... @bcreager, any thoughts here?

bcreager (approved):

PR #4163, opened 2021-08-13
change image format

InesRoitman:


InesRoitman:
I think I'm doing this right XD

felixfbecker:
Hmm this is an SVG, but the text inside it is not included as text but as vector graphics, meaning it is still not readable by screen readers. Mind sharing what tool was used to create this infographic? Maybe I can help.

PR #4230, opened 2021-08-18
fix page rendering

davejrt:
This stops these two handbook pages being rendered incorrectly. 

I am not sure if this is a recent change that the pages wouldn't grow and accommodate long lines. I'm not familiar enough with the handbook code but this unblocks people needing to use these pages right now. 

Perhaps someone with more experience developing the handbook knows why?


davejrt:
Example:
image

eseliger (approved):

PR #4249, opened 2021-08-19
Adding notes about access...

virginiaulrich:
...to sourcegraph.com as an admin and managed instances


jasonhawkharris (approved):
looks good to me!

virginiaulrich:
@alexAtSourcegraph went ahead and merged this, but would still welcome your thoughts when you have a moment. No hurry on that part -- we can edit however we need to!

alexAtSourcegraph (commented):
Crystal clear 👍

PR #4251, opened 2021-08-19
Remove explicitly set width and height on blog post

csbailey5t:
The explicitly set width and height, in combination with the embed's css, push the embeds off screen on mobile widths. Removing these properties provides an immediate fix, though doesn't address spacing between the bottom of the embeds and the next text paragraph.


rebeccadee:
@csbailey5t Thank you so much for doing this #highagency 😁

PR #4259, opened 2021-08-20
Add automatic Prettier GitHub action

felixfbecker:
Overall, the goal of this PR is to make the content in this repo a lot easier to edit and read without getting into anyone's way.

Currently, editing CSS or JS (and even Markdown in some cases) for the handbook or the website is a really frustrating experience. Indendation is inconsistent between 2 and 4, sometimes tabs and spaces are mixed within the same file. Selectors are sometimes formatted in a hard to read way. TSX needs to be indented _manually_.

There is trailing whitespace everywhere, especially in Markdown files. It is extremely difficult to not end up with huge diffs (and consequential merge conflicts) on every PR because editors auto-trimmed whitespace without having to jump through many hoops to reconfigure editors just when editing the handbook.

Markdown is formatted in inconsistent ways when Markdown supports multiple ways to do the same thing (lists, bolding, italics, ...),  making it harder to read and understand the plain text source, sometimes even leading to errors.

Markdown tables are effectively unreadable (and uneditable) in their source because the columns are completely misaligned (and who has time to format them manually?).

**All of this could be easily solved by Prettier with autoformatting.** What speaks **against** that is:
1. People would be _required_ to autoformat it (can't edit without proper editor setup or through the GitHub UI)
2. We'd need a CI check and could not push to `main` directly anymore, meaning every change needs to go through a PR
3. The initial diff would probably reformat many files and make `git blame` harder to trace

**I believe all of these points are solvable, and that this PR addresses each.**

1. It adds a Prettier GitHub action that will automatically reformat files on PRs and push a commit to the branch, meaning people don't need to worry about setting up a local editor and can even continue to use the GitHub UI for edits. On merge the commit is squashed so there will be no trace of it on `main`.

2. We've already enforced PRs and required status checks for a while now after many breakages had happened from pushing to `main` that would cause subsequent changes and the deploy to be blocked. I consider this a positive change for many reasons, as overall it also makes sure as our organization scales edits to the handbook are done in PRs that can be reviewed and commented on by anyone at the company.

3. For (3), Git supports the option `--ignore-revs-file` (also configurable globally with `git config`) that has been adopted in open source projects already (e.g. https://github.com/v8/v8/blob/master/.git-blame-ignore-revs). After merging this PR, I will add such a file with the commit hash of the format commit, which would still allow us to blame, ignoring the first large format commit. I will also say though that I think we should do this regardless because with frequent changes to the handbook eventually the commit will be long forgotten about anyway and we do this all the time in our main repo too, even without a `blame-ignore-revs` file, e.g. every time we upgrade ESLint. The benefits of editing experience outweigh the cost.


Joelkw:
As someone who often forgets to run prettier (because of infrequency of updates and/or having to push something between meetings) this would be awesome!

eseliger (approved):
neat 👍

courier-new (approved):
What a great idea!

> For (3), Git supports the option --ignore-revs-file (also configurable globally with git config) that has been adopted in open source projects already (e.g. https://github.com/v8/v8/blob/master/.git-blame-ignore-revs). After merging this PR, I will add such a file with the commit hash of the format commit, which would still allow us to blame, ignoring the first large format commit. I will also say though that I think we should do this regardless because with frequent changes to the handbook eventually the commit will be long forgotten about anyway and we do this all the time in our main repo too, even without a blame-ignore-revs file, e.g. every time we upgrade ESLint. The benefits of editing experience outweigh the cost.

😱 TIL. We should *absolutely* start doing this when we upgrade ESLint or something similar in the main repo, too.

.github/workflows/prettier.yml
```diff
@@ -0,0 +1,39 @@
...
+          cache: yarn
+        run: yarn
+
+      - name: Read package.json
+        id: read_package_json
+        run: |
+          JSON=$(cat package.json)
+          echo "::set-output name=packageJson::${JSON//'%'/'%25'}"
+
+      - name: Prettier
```
valerybugakov, line 31: Awesome automation! What do you think about running prettier only against changed files on CI?
It seems that such an option is available [out of the box](https://github.com/creyD/prettier_action#example-2-using-the-only_changed-or-same_commit-option-on-pr).

.github/workflows/prettier.yml
```diff
@@ -0,0 +1,39 @@
...
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          # Make sure the actual branch is checked out when running on pull requests
+          ref: ${{ github.head_ref }}
+
+      # Make sure @sourcegraph/prettierrc is installed
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '14'
```
valerybugakov, line 21: Curious, what do you think about using the latest [lts](https://github.com/actions/setup-node/issues/26#issuecomment-871413720) version here?

valerybugakov (approved):
Thanks for the detailed PR description! 🤗

.github/workflows/prettier.yml
felixfbecker: I honestly don't trust it enough for the benefit it would provide (performance). It is implemented with `git diff --name-only --diff-filter=d HEAD^..HEAD`, which already I am not sure if it will handle merge commits correctly, which appear frequently when people click "Update branch" in GitHub. Prettier is already pretty fast (haha), probably the fastest build step from the ones we have, so I don't think it's worth the risk/complexity over just formatting all files on every run.

.github/workflows/prettier.yml
felixfbecker: I updated it to read .nvmrc, then Renovate can upgrade it accordingly.

nicksnyder:
This broke nested list formatting in our handbook. What is the right way to fix this?

Screen Shot 2021-09-13 at 11 30 06 AM

felixfbecker:
@nicksnyder The right "fix" would be to migrate our Markdown renderer from `blackfriday` to e.g. `goldmark`. More context in https://github.com/sourcegraph/about/issues/4446 and https://github.com/russross/blackfriday/issues/329. People have been struggling with blackfriday's rendering even before Prettier because the 4 spaces indentation is not intuitive and inconsistent with how GitHub and any other CommonMark-compliant parser renders it. And not being CommonMark-compliant also brings other bugs with it unrelated to nested lists or Prettier that exist in our handbook (e.g. bolding inside links).

Introducing Prettier of course made this issue a lot more apparent and I'm sorry for that – I didn't think of this bug when adding it. I still think it's the right thing to have Prettier though considering the benefits and that this was a frequent (although slightly less frequent) problem before. I investigated several workarounds (including changing indentation in the Prettier config, which didn't work as expected) and I think the best option is to just use `<!-- prettier-ignore -->` for now for nested lists where this issue occurs.

Btw, this even a bug in our Markdown rendering in our product too, because we use `blackfriday` there too.

PR #4275, opened 2021-08-23
Add why Google dev tools post worked

rebeccadee:


rebeccadee:
@csbailey5t @jennybergen 👋 I think you discussed this post as part of Scott's interview process so I'd be keen to hear if there's anything else you'd add here? Thanks!

jennybergen:
@rebeccadee you did a great job at pulling out the main points of why this post works so well. +1 to all you wrote! I think those are the most important things to call out.

The only thing I'd add that I think "works" for this post, which gets a bit more detailed, is that Beyang did a good job at leading into the topic of the post and clearly stating what the reader can expect to learn. He broke up his text into sections with subheaders and broke up what would otherwise be large chunks of text with bullet points or numbered lists.

PR #4351, opened 2021-08-28
Clarify activation

malomarrec:
Clarify the definition of activation, make sure it cannot be confused with active.
Activation definition is still WIP, this is a placeholder until we determine what leads to success. Adding private code is a good candidate.


handbook/ops/bizops/user_definitions.md
```diff
@@ -60,22 +60,30 @@ In the site admin panel, we have a Usage stats page that displays number of MAUs
...
 
-##### Activation qualifying events
+### Key metrics
 
-Users activate if they perform one of these qualifying events:
+We track the following key metrics:
 
-| Product area              | Qualifying events                  | Description                                                                                         |
+- acquisition: number of visitors
+- activation: the percentage of users that had a successfull initial experience.
```
dadlerj, line 79:
```suggestion
- activation: the percentage of visitors that had a successful initial experience.
```

Just to check, is this correct? Or should it be "the percentage of active users that had a successful initial experience"? Curious how the calculation is performed/what the denominator is.

handbook/ops/bizops/user_definitions.md
```diff
@@ -60,22 +60,30 @@ In the site admin panel, we have a Usage stats page that displays number of MAUs
...
+- retention: the number of active users retained time period over time period
+
+#### Activation
```
dadlerj, line 82: Just to confirm, the list of events below is not the same as the list of events that is used as the definition of an active user in a period, right? That list has more included (such as viewing search results)?

dadlerj (approved):
Thank you!

handbook/ops/bizops/user_definitions.md
malomarrec: Indeed. The list of events is very large, and define in Amplitude. This is defined in the table above.

PR #4364, opened 2021-08-31
Remove inline style on image to fix scaling on mobile

csbailey5t:
Re: Hashimoto Dev Tool Time iPad workflow image

### DIFF
--- a/blogposts/2021/dev-tool-time-mitchell-hashimoto.md
+++ b/blogposts/2021/dev-tool-time-mitchell-hashimoto.md
@@ -35,7 +35,7 @@ Mitchell also has a stack of [Raspberry Pi](https://www.raspberrypi.org/) comput
 
 > "I just like this form-factor of carrying something in my hand."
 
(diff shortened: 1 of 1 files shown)


rebeccadee:
@csbailey5t Thank you for doing this! There may not be a way around this, but removing the sizing makes the inline images huge now: 

Screenshot 2021-08-31 at 11 48 47

csbailey5t:
We probably need to set a more explicit `max-width`. Right now, our CSS is calculating `max-width` and `max-height` based on viewport width and height. We *should* be fixing this in the underlying CSS so we aren't needing to explicitly set any styles in our markdown, but that's a future step.

csbailey5t:
Checking back over previous posts, this is a bug that affects any of the images in the blog that have the explicitly set smaller height and width.

csbailey5t:
@rebeccadee I'm going to go ahead and merge this, since the images in this post per this PR are consistent with the video embed at the bottom, and slightly large images on a large desktop are probably better than having distorted images on mobile. A real fix for this can be part of the larger blog redesign.

rebeccadee:
High agency in action, thanks @csbailey5t!

PR #4444, opened 2021-09-03
Clarify priority blog content

rebeccadee:
Aims to eliminate confusion around the role and execution of the priority buckets of content for the blog.


nickmyyz:
> Aims to eliminate confusion around the role and execution of the priority buckets of content for the blog.

This is much clearer, I think!

PR #4463, opened 2021-09-07
Create readmes subfolder

jyavorska:
The root of the product folder was collecting cruft with multiple readmes in it.


jyavorska:
Hey @dan-mckean what do you think? I added my readme, but I thought it might make more sense to have them in a subfolder together.

PR #4464, opened 2021-09-07
Delete how-to-deconstruct-the-monolith.md

rebeccadee:
Deletes post because `published: false` doesn't seem to be working


rebeccadee:
FYI @content-hacker @jennybergen https://sourcegraph.slack.com/archives/C07KZF47K/p1630856161192500

PR #4469, opened 2021-09-07
Use goals page as index

jyavorska:
This makes the goals page an index of links to pages with per-team goals.


jyavorska:
@anna-mmikhova @serinadawn @christinaforney what do you think? The content that was here previously was not updated in a year. I understand the individual goal pages are out of date too, but this can be a step towards getting those up to date.

The list of goals pages is copied from the onboarding materials. Should any be removed? Anything missing that should be added?

christinaforney (approved):
Great start! Feel free to merge as is and we can continue to iterate on an improved structure for planning docs 😄

PR #4477, opened 2021-09-07
Update code insights planning process

felixfbecker:
This is a proposal for how we can mature the planning process we used through FQ2, facilitated by the new GitHub project Beta, so it is easier to repeat and scales to our new team size and product maturity.

It aims to set clearer expectations for when and how in the iterations planning happens than we had previously documented. One of the goals is to be as asynchronous as possible, while still allowing synchronous discussion (in the weekly sync and ad-hoc).
As before, all teammates are given agency in planning (planning does not happen top-down).
It also includes the expectation for maximum workload and how to deal with overflowing issues to avoid overloading teammates or "crunching" in iterations (we want to work sustainably – that's the main reason we're planning at all).
It also adds some notes on how we deal with the mismatch between iterations<->release dates.

Additionally, this PR documents part of our process/philosophy we had been practicing but hadn't documented yet, namely how we approach longer-term (multi-iteration) projects: The process of @Joelkw's product RFCs, the following design/planning phase, and finally the implementation phase – which we parallelize to plan the next project ahead, but constrain to never have too many things going on at the same time.


handbook/engineering/developer-insights/code-insights/processes.md
```diff
@@ -16,28 +16,79 @@ Attendees are encouraged to add comments on talking points asynchronously before
...
-- When starting work, engineers pull cards from the _Planned_ column and move it to the _In Progress_ column.
-- There should never be more than a couple of cards in the _In Progress_ column at the same time. If we paused work on something (e.g. because priorities changed), it should be moved back to _Planned_ or _Icebox_ as appropriate.
-- Things we triaged from the _Inbox_, but don't plan to work on, go into the _Icebox_.
+Our current process is as follows:
 
-Note: we intentionally plan so we can avoid merging significant work less than two days before a release (if a release is on the 20th, our last day to merge is the 18th). Exceptions require explicit approval of both the PM and EM.
+- We work in **2-week iterations**. Every iteration as a GitHub milestone, which is created at least one iteration in advance.
 
-## Tracking Issues
+- Incoming tickets (from team members or from other teams) are added to the [GitHub project](https://github.com/orgs/sourcegraph/projects/200/views/1) and will be visible in the "All iterations" view under "No milestone". This represents our **backlog** of _unscheduled_ work.
```
Joelkw, line 26: Can we make "are added" less passive voice and/or just include what/who adds them? (Is this you/me? Is it a bot? If I was reading this as an "outsider/CE" it should ideally not leave me questioning "wait, do I need to add them?")

handbook/engineering/developer-insights/code-insights/processes.md
```diff
@@ -16,28 +16,79 @@ Attendees are encouraged to add comments on talking points asynchronously before
...
+
+  - Engineers, designer, PM and EM **_propose_ issues to be worked on** the week _before_ the mid-iteration [Monday sync](#weekly-sync) by adding them to the next iteration milestone and setting "Status" to "Proposed". This is done considering our [higher-level goals and roadmap](./goals.md).
+
+    - The issue should also have its **_Estimate_ column filled out**, so that it can be evaluated whether it fits into the iteration. If the proposer lacks the information to estimate the issue, they reply on the issue in GitHub or raise it in our Slack channel to get the missing information or get an estimate from the appropiate person. Teammates may also discuss this in ad-hoc synchronous meetings if beneficial. An assignee may also already volunteer or be proposed, but this may still be changed at the [Monday sync](#weekly-sync) to distribute workload.
+      - If **technical exploration** is needed to get more information, a _spike_ (a time-boxed investigation task meant to facilitate more granular planning) can be proposed for the next iteration instead to get that information.
```
Joelkw, line 33: Minor – what if we both need a spike and otherwise need to complete both spike + feature in the iteration? (Let's say the estimation uncertainty is "anywhere from 2-5 days"; can we just slate both spike and feature for that iteration?)

handbook/engineering/developer-insights/code-insights/processes.md
```diff
@@ -16,28 +16,79 @@ Attendees are encouraged to add comments on talking points asynchronously before
...
+
+  - As much as possible, the proposer **involves the necessary stakeholders _asynchronously_ to get agreement** on whether the issue should be worked on in the next iteration before the [Monday sync](#weekly-sync). For example, the PM or EM might ping engineers in GitHub or Slack on whether an issue seems feasible, or engineers might ping their EM and PM to get buy-in whether the issue fits into our goals.<br>
```
Joelkw, line 35: :+1:

handbook/engineering/developer-insights/code-insights/processes.md
```diff
@@ -16,28 +16,79 @@ Attendees are encouraged to add comments on talking points asynchronously before
...
+
+- During an iteration, teammates **work on their assigned issues for the iteration in the order they are listed** in the ["Current iteration" view](https://github.com/orgs/sourcegraph/projects/200/views/1) of the board. When starting work on a task, the teammate **updates its status column to "In Progress"** to communicate it to their team. This gives a good overview in the ["Current iteraton" view](https://github.com/orgs/sourcegraph/projects/200/views/1), which can also be viewed in [Kanban layout](https://github.com/orgs/sourcegraph/projects/200/views/1?layout=board), on how the iteration is tracking.
+
+- If one or more issues that were planned for an iteration are looking to **not get finished** in the [current iteration](https://github.com/orgs/sourcegraph/projects/200/views/1) (while maintaining sustainable work practices) the assignee **raises this as soon as possible asynchronously** to the team (including the PM and EM), e.g. on the GitHub issue or Slack. These issues then **become _proposed_ issues for the next iteration** (meaning nothing carries over automatically, but we also don't just drop and forget missed issues).
```
Joelkw, line 45:
```suggestion
- If one or more issues that were planned for an iteration will **not be finished** (which includes testing) in the [current iteration](https://github.com/orgs/sourcegraph/projects/200/views/1) while maintaining sustainable work practices, the assignee **raises this as soon as possible asynchronously** to the team (including the PM and EM), e.g. on the GitHub issue or Slack. These issues then **become _proposed_ issues for the next iteration** (meaning nothing carries over automatically, but we also don't just drop and forget missed issues).
```

handbook/engineering/developer-insights/code-insights/processes.md
```diff
@@ -16,28 +16,79 @@ Attendees are encouraged to add comments on talking points asynchronously before
...
+Despite following two-week iterations, our [releases are monthly on the 20th](../../releases.md#releases) and we may sometimes need to order tasks in a way that we can get important projects into the next release.
+
+We also intentionally plan so we can avoid merging significant work less than two days before a release (if a release is on the 20th, our last day to merge is the 18th). Exceptions require explicit approval of both the PM and EM.
+
+### Planning larger projects
+
+Larger (usually multi-iteration) projects coming from our roadmap go through three phases:
+
+1. **Product context and scope agreement** (usually 1-2 weeks)<br>
+   The PM creates a product-level RFC summarizing user desires from feedback and proposes a reasonable timeline to reach a scope agreement (mentioned in the RFC).
```
Joelkw, line 58:
```suggestion
   The driving individual creates a product-level RFC summarizing user desires from feedback and proposes a reasonable timeline to reach a scope agreement (mentioned in the RFC).
```
In some cases it could be Felix or Alicja, and in engineering-focused major projects I imagine anyone on the team might lead these efforts? Or do you want to specify "larger projects" = "larger product features"?

handbook/engineering/developer-insights/code-insights/processes.md
```diff
@@ -16,28 +16,79 @@ Attendees are encouraged to add comments on talking points asynchronously before
...
+
+### Releases
+
+Despite following two-week iterations, our [releases are monthly on the 20th](../../releases.md#releases) and we may sometimes need to order tasks in a way that we can get important projects into the next release.
+
+We also intentionally plan so we can avoid merging significant work less than two days before a release (if a release is on the 20th, our last day to merge is the 18th). Exceptions require explicit approval of both the PM and EM.
+
+### Planning larger projects
+
+Larger (usually multi-iteration) projects coming from our roadmap go through three phases:
```
Joelkw, line 55:
```suggestion
Larger (usually multi-iteration) projects coming from our roadmap go through three phases. (Note that these three phases are the team collaboration phases, but there are often additional individual phases like product/design research processes or technical research that may come before phase 1) 
```
Might be worth noting just so folks know we're doing research before the RFCs are created/don't expect research + RFC to land at the same time

handbook/engineering/developer-insights/code-insights/processes.md
```diff
@@ -16,28 +16,79 @@ Attendees are encouraged to add comments on talking points asynchronously before
...
+   After discussion, the agreed-upon scope is recorded in a section at the bottom of the RFC.
+
+2. **Design and implementation planning** (usually 1-2 weeks)<br>
+   For a UI project, our designer will create wireframes and designs which are reviewed and discussed by all teammates in Figma.
+   Similarly, engineers will make a proposal for the needed GraphQL APIs (e.g. in an RFC) and agree on the API shape, think the implementation through, and create concrete implementation sub-tasks with estimates, testing plans and a tracking issue collecting them.
+   The goal of this phase is to be "ready to go" for the implementation phase – having a plan (with reasonable confidence) that only needs to be executed.
+
+3. **Implementation and testing** (usually 1-4 weeks)<br>
+   Engineers execute on the implementation plan, putting a set of issues from the tracking issue into each iteration.
+   This also includes that each sub-implementation-task is sufficiently tested, meaning by the end of this phase the project is ready to ship to customers with confidence.
```
Joelkw, line 70: I would like us to specify where/how the testing goes. Do individual teammates add it to issues and does one of us review it? Do I/we include broader test plans earlier in this process? Either could work, curious what you were thinking.

handbook/engineering/developer-insights/code-insights/processes.md
```diff
@@ -16,28 +16,79 @@ Attendees are encouraged to add comments on talking points asynchronously before
...
+
+  - Engineers, designer, PM and EM **_propose_ issues to be worked on** the week _before_ the mid-iteration [Monday sync](#weekly-sync) by adding them to the next iteration milestone and setting "Status" to "Proposed". This is done considering our [higher-level goals and roadmap](./goals.md).
+
+    - The issue should also have its **_Estimate_ column filled out**, so that it can be evaluated whether it fits into the iteration. If the proposer lacks the information to estimate the issue, they reply on the issue in GitHub or raise it in our Slack channel to get the missing information or get an estimate from the appropiate person. Teammates may also discuss this in ad-hoc synchronous meetings if beneficial. An assignee may also already volunteer or be proposed, but this may still be changed at the [Monday sync](#weekly-sync) to distribute workload.
+      - If **technical exploration** is needed to get more information, a _spike_ (a time-boxed investigation task meant to facilitate more granular planning) can be proposed for the next iteration instead to get that information.
+
```
Joelkw, line 34:
```suggestion
         - Estimations **include the testing time necessary** to fully test an issue against test plans or test cases defined in the issue. 
         
```

and then is the assumption you/me add test cases to issues, or at least folks start by adding their own and we review them?

Joelkw (approved):
Thanks for drafting this! most of my "important" comments were trying to add places to define that done = also tested, and where/how we define test plans!

handbook/engineering/developer-insights/code-insights/processes.md
felixfbecker: I intentionally chose "are looking to" over "will" because I don't want anyone to wait until they _definitiely know_ it won't make it (because it's Friday morning at the end of an iteration...) but rather raise it earlier (I'd rather have a false positive sometimes).

I do like your testing suggestion though

handbook/engineering/developer-insights/code-insights/processes.md
felixfbecker: Haha ideally a GitHub action but in lieu of that we'll all have to do it manually (and correct if we catch a missing one). Working on automation

handbook/engineering/developer-insights/code-insights/processes.md
felixfbecker: I think that's possible. Do you think there's value in documenting that? I'm a little concerned that this process becomes very long to read through/overwhelming

handbook/engineering/developer-insights/code-insights/processes.md
felixfbecker: Hmm I was thinking back and forth on this. My primary motivation was to basically document what you were already doing for larger "product features": "What are these RFCs that Joel always does? What are they for? What am I expected to do with it? What comes after it?"

At the same time, there can absolutely be "larger" features that go through similar phases that are driven by engineering or design, my intend is by no means to prohibit or discourage that.
However, in those cases the expectations here would also not be fitting – I wouldn't expect an engineering RFC to necessarily come prepared with all the customer feedback like a Product RFC from our PM does. It might even skip this phase entirely, but still adapt the other phases.
But then if we reword this to make this more generic, the details and specificity on your RFC format are lost.

So honestly not sure which of your two options is more useful: To only document "Larger **product** features" and leave it implicit this can be a blueprint for other larger projects, or to document a generic process for all larger features with less specificity.

handbook/engineering/developer-insights/code-insights/processes.md
felixfbecker: Good question (I hadn't thought about this as part of this PR). I would say both: The higher-level expectations could be documented as part of the scope agreement in the product RFC. Concrete testing for each sub-task (issue) I think should be included by the issue author (engineers, usually) when the issues are filed in the planning phase.

handbook/engineering/developer-insights/code-insights/processes.md
felixfbecker: Engineers are responsible end-to-end to ship a _High Quality_ product with _High Agency_, so I think it should not be me/you generally. I'd generally say the issue author if possible, otherwise the assignee can add it (editing the issue), at some point before starting implementation.

CristinaBirkel (approved):
Just a couple of typos. I think it all makes sense to me, but I will probably have more of an opinion once I've gone through a couple of iterations! 😄

handbook/engineering/developer-insights/code-insights/processes.md
unclejustin: Agree with @felixfbecker here. I don't think we need to explicitly state that we may have exceptions. During the sync meetings if something comes up we should be able to just say it then and move on.

unclejustin (approved):

handbook/engineering/developer-insights/code-insights/processes.md
AlicjaSuska: I think that using 'driving individual' in all cases is better. It doesn't mean that the PM cannot participate in the RFC, it's only about driving the innitiative.

handbook/engineering/developer-insights/code-insights/processes.md
```diff
@@ -16,28 +16,80 @@ Attendees are encouraged to add comments on talking points asynchronously before
...
+Larger (usually multi-iteration) projects coming from our roadmap go through three phases:
+
+1. **Product context and scope agreement** (usually 1-2 weeks)<br>
+   The PM creates a product-level RFC summarizing user desires from feedback and proposes a reasonable timeline to reach a scope agreement (mentioned in the RFC).
+   Engineers, designer (as applicable), PM, and EM discuss the needs in the RFC and the feasibility/complexity of different options.
+   **The goal of this phase is to agree on a scope for the project**, which will generally be a _subset_ of the desires the PM outlined in the RFC based on the implementation complexity, how long it would roughly take to implement, and how much time we want to dedicate to the project.
+   After discussion, the agreed-upon scope is recorded in a section at the bottom of the RFC.
+
+2. **Design and implementation planning** (usually 1-2 weeks)<br>
```
AlicjaSuska, line 64: I'm not sure about specifying the timelines here. @felixfbecker what's the goal of it?

handbook/engineering/developer-insights/code-insights/processes.md
```diff
@@ -16,28 +16,80 @@ Attendees are encouraged to add comments on talking points asynchronously before
...
-  - Work should not be moved into the _Planned_ column until it is ready for development _and_ estimated and prioritized.
-  - Work that is ready for development but not yet estimated or prioritized sits in the _Ready for Dev_ column.
-  - Anything that needs design input gets the `needs-design` label and goes in the _Needs design_ column.
-- When starting work, engineers pull cards from the _Planned_ column and move it to the _In Progress_ column.
-- There should never be more than a couple of cards in the _In Progress_ column at the same time. If we paused work on something (e.g. because priorities changed), it should be moved back to _Planned_ or _Icebox_ as appropriate.
-- Things we triaged from the _Inbox_, but don't plan to work on, go into the _Icebox_.
+Our current process is as follows:
 
-Note: we intentionally plan so we can avoid merging significant work less than two days before a release (if a release is on the 20th, our last day to merge is the 18th). Exceptions require explicit approval of both the PM and EM.
+- We work in **2-week iterations**. Every iteration as a GitHub milestone, which is created at least one iteration in advance.
```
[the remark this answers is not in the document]
felixfbecker: Ah this was actually missing an "h" ("has a milestone"). Thanks!

handbook/engineering/developer-insights/code-insights/processes.md
felixfbecker: @AlicjaSuska are you saying you would prefer to document a _generic_ process here and remove the expectations around summarizing user desires/feedback? (see my comment above)

handbook/engineering/developer-insights/code-insights/processes.md
felixfbecker: To give a general idea of the timescale of these phases and the whole cycle overall. These can vary per project of course, that's why I wrote "usually".

I considered instead documenting that the timeline should be set explicitly at the start of each phase, but I think it helps to have a rough expectation of timelines before-hand. E.g. Joel shouldn't expect that he can put up a product RFC and then go through the review and design process in a matter of days (not saying he would!), and engineers should have a general understanding of if we're in week 2 of the design phase and the design is ready for review now, they're expected to not put the review at the end of my TODO list for 2 weeks because that would blow the project timeline out of proportion.

While it is good to have a conversation about timelines for every project, having a rough expectation means we don't need to re-negotiate every time whether we're talking about days or months.

It also gives a rough idea of which size project should be to use this process, and at which size they should be broken up into multiple projects. In a way that also helps when selecting the project scope in the first phase (pushing back on a scope that could not be designed or implemented in this order of magnitude of time).

Do you have a concern here? (with documenting a rough timeline at all, or with the specific numbers?)

handbook/engineering/developer-insights/code-insights/processes.md
```diff
@@ -16,28 +16,80 @@ Attendees are encouraged to add comments on talking points asynchronously before
...
+
+- If one or more issues that were planned for an iteration are looking to **not get finished** (which includes testing) in the [current iteration](https://github.com/orgs/sourcegraph/projects/200/views/1) (while maintaining sustainable work practices) the assignee **raises this as soon as possible asynchronously** to the team (including the PM and EM), e.g. on the GitHub issue or Slack. These issues then **become _proposed_ issues for the next iteration** (meaning nothing carries over automatically, but we also don't just drop and forget missed issues).
+
+### Releases
+
+Despite following two-week iterations, our [releases are monthly on the 20th](../../releases.md#releases) and we may sometimes need to order tasks in a way that we can get important projects into the next release.
+
+We also intentionally plan so we can avoid merging significant work less than two days before a release (if a release is on the 20th, our last day to merge is the 18th). Exceptions require explicit approval of both the PM and EM.
```
coury-clark, line 52: In general I don't like this code freeze. I think it hasn't been particularly effective at stopping issues, and I think it leaves a lot of unresolvable ambiguity around what is `significant`. I also think it's a relatively low agency constraint that anyone in any form of leadership role should need to be an approver to merge code, as well as the more practical problem that leaving branches for many days in a very active repo can make merging unnecessarily complicated and risky.

What were the original motivations behind this freeze?

coury-clark (commented):
I left some thoughts - just challenging some points to vet them.

I will say in general I think this process works fine for well defined features, and I think it works less well for ambiguous problem statements. For example, "build dashboards" versus "improve performance". In the past I've been on teams that try to fit work that clearly doesn't match the model process and it's very awkward (square peg round hole thing). I'd love to see some flexibility that if we take on a project that doesn't fit the mold very well, we are flexible enough to adapt.

handbook/engineering/developer-insights/code-insights/processes.md
vovakulikov: @coury-clark as I remember that that was introduced to avoid last-minute change before the release cut which we had pretty often (and I guess we still have it). I think this rule isn't bad but because we have been violating this rule for a long time already (as I remember the last 3 releases had these last-minute changes literally) we need to accept that this rule just doesn't work for us. 

But it is also important to note that we usually violate this rule not because we want to ship another big thing but cause we usually find some problems on staging right before the release cut. We probably need to reconsider this rule. Like to be a truly code-freezed before the release cut I think we should end our active development 2-3 days before the code-freeze and then start the testing process actively to check things that we haven't covered by unit/integrations tests but with that system, we reduce our time for developing something for the next release. So I don't have a good solution to remove/replace this rule with something else but just mix feeling about that.

handbook/engineering/developer-insights/code-insights/processes.md
Joelkw: Sounds good!

handbook/engineering/developer-insights/code-insights/processes.md
Joelkw: Cool, automation would be awesome, otherwise can we get semi-automation? We used to have a bot that could automatically put issues in a project kanban column – I'd be okay if we just made a "fake" project board (the old kind) and automated the inbox the same way, then you or I could use that to send things to the new project view.

handbook/engineering/developer-insights/code-insights/processes.md
Joelkw: I would be okay replacing this with an alternative proposal, but I would otherwise keep this as the "default in lieu of a better solution." As the team grows I would expect our process to naturally mature (and as we grow we should have fewer moments where we have to choose between building in testing, since we should have a capacity to do both). 

I consider "significant" to be roughly anything that modifies more than one directional flow or that is not just a copy change (copy changes often land close to release due to needing more stakeholders to review + catching things on dogfood, and I'm okay with that). You can also think of "significant" as "how many possible paths to test this are there?" and anything with more than 1-2 paths is likely significant.

vovakulikov (approved):

handbook/engineering/developer-insights/code-insights/processes.md
```diff
@@ -16,28 +16,80 @@ Attendees are encouraged to add comments on talking points asynchronously before
...
+
+- During an iteration, teammates **work on their assigned issues for the iteration in the order they are listed** in the ["Current iteration" view](https://github.com/orgs/sourcegraph/projects/200/views/1) of the board. When starting work on a task, the teammate **updates its status column to "In Progress"** to communicate it to their team. This gives a good overview in the ["Current iteraton" view](https://github.com/orgs/sourcegraph/projects/200/views/1), which can also be viewed in [Kanban layout](https://github.com/orgs/sourcegraph/projects/200/views/1?layout=board), on how the iteration is tracking.
+
+- If one or more issues that were planned for an iteration are looking to **not get finished** (which includes testing) in the [current iteration](https://github.com/orgs/sourcegraph/projects/200/views/1) (while maintaining sustainable work practices) the assignee **raises this as soon as possible asynchronously** to the team (including the PM and EM), e.g. on the GitHub issue or Slack. These issues then **become _proposed_ issues for the next iteration** (meaning nothing carries over automatically, but we also don't just drop and forget missed issues).
```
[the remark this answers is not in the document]
felixfbecker: What is preventing us from "just dropping and forgetting" missed issues is that the process says we _discuss_ every "Proposed" issue before it gets added to "Todo". The outcome of that discussion _can_ be moving the issue to the Backlog, but the point is that it is only done after discussion and making an explicit choice to do so. The other possible outcome is that we do put it on the next iteration plan.
This is contrary to a model where all missed issues get added to the Backlog by default.

handbook/engineering/developer-insights/code-insights/processes.md
```diff
@@ -16,28 +16,80 @@ Attendees are encouraged to add comments on talking points asynchronously before
...
+   The goal of this phase is to be "ready to go" for the implementation phase – having a plan (with reasonable confidence) that only needs to be executed.
+
+3. **Implementation and testing** (usually 1-4 weeks)<br>
+   Engineers execute on the implementation plan, putting a set of issues from the tracking issue into each iteration.
+   This also includes that each sub-implementation-task is sufficiently tested, meaning by the end of this phase the project is ready to ship to customers with confidence.
+
+We sequentialize and parallelize projects in a way that we can _plan_ projects (step 1 and 2) while another project is currently being _implemented_, so that we always have the next project ready to be implemented by the time a project has finished implementation.
+We will however make sure to never have multiple projects in planning phase at the same time, as this leads to cognitive overload while also delivering on an implementation for another projects.
```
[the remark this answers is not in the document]
felixfbecker: The goal ultimately is to avoid cognitive overload on individuals, so you can think of it as "not more than one project going on at the same time _per person_". It is very likely that as our team grows, we'll need to loosen this up to allow multiple groups of people working on different projects simultaneously. But at the same time, [we prefer to operate together as a team rather than a working group](https://about.sourcegraph.com/handbook/company/values#work-as-a-team) where everyone is working on different things. At least right now, most big projects we do will involve backend and frontend together, and we're just 2+2 engineers. We also have only one designer and one PM. So I believe at the moment if we do multiple large projects at the same time, it _will_ result in increased cognitive load for the team, which I want to avoid. And again, this may change in the future!

handbook/engineering/developer-insights/code-insights/processes.md
```diff
@@ -16,28 +16,80 @@ Attendees are encouraged to add comments on talking points asynchronously before
...
+This means a teammate may have to think at most about 2 projects at any given time: one that is being implemented, and one that is being planned, and a teammate will never have to implement more than one larger project at the same time.
+
+<object data="./project_planning.svg" width="1100" height="263" aria-label="Diagram showing three example projects going through all three project phases, with each product and design phase in parallel to the implementation phase of the previous project."></object>
+
+#### Tracking issues
+
+To plan projects that span multiple iterations and need to be broken up into sub-tasks, we make use of [tracking issues](../../tracking_issues.md).
+The tracking issue is created by one of the teammates with a descriptive accompanying label, e.g. <span class="badge bg-info">insights-dashboards-v1</span> (milestones are not used for this, as they are used for iterations).
```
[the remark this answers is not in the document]
felixfbecker: We could technically also use labels for iterations and milestones for projects. We can't use both, because GitHub only allows one milestone per issue, but issues need to be possible to associate with an iteration AND a project at the same time – so we just need to align on one. Iterations are gonna be created more frequently, and they have a workflow of "closing" and due dates that labels don't have and are a bit more natural to display and edit in the new GitHub board (as it is a dedicated column). You also can't group by labels (or label "pattern") in the new boards, but you can group by milestone, which is used for the "All issues" view, which would not be possible if iterations were tracked by labels.

handbook/engineering/developer-insights/code-insights/processes.md
AlicjaSuska: ok, makes sense. thank you

handbook/engineering/developer-insights/code-insights/processes.md
felixfbecker: Added full automation.

felixfbecker:
Thank you for the feedback everyone, looks like everyone is happy with now! I'll merge and don't forget we can keep iterating on it (feel free to also open PRs for any tweaks!)

PR #4495, opened 2021-09-08
Expand on Growth company value and add initiatives and resources

marekweb:
"Continuously grow" is one of our company values but it currently has less detail and explanation than the other values . Yet this is an important value and it intersects with a lot of the things we do at Sourcegraph.

This PR is an attempt to improve this by expanding on the description of the "Continuously grow" value, adding more details about how we live this value, and listing initiatives and resources that tie into it.

Hopefully as we grow more internal initiatives in growth, learning & development, and education we can add more to this section.

### DIFF
--- a/handbook/company/values.md
+++ b/handbook/company/values.md
@@ -110,3 +110,20 @@ Our code and PRs are public, so lack of transparency is only obfuscation, not se
 ### Continuously grow
 
 You strive to continuously grow and learn by genuinely soliciting feedback early and often, and humbly reflecting on your past mistakes.
+
+We value growth at Sourcegraph in the form of continuous learning:
+
+- We actively seek out learning opportunities instead of avoiding them.
+- We identify gaps in our knowledge and share them openly and honestly.
+- We adopt a growth mindset.
+
+Here are some ways that we support learning, teaching, and knowledge-sharing at Sourcegraph:
+
+- We aim to provide a high-quality [onboarding experience](../people-ops/onboarding/index.md) for new teammates by providing them with the learning materials and guidance they need to get started.
+- We have an [education budget](../people-ops/travel.md#professional-development-and-education) for professional development. This budget is dedicated to helping teammates expand their knowledge, whether it's through books, courses, training, or other resources.
+- We organize occasional hackathons internally, which are opportunities to learn, experiment, and share knowledge.
+- Product and Engineering teams host regular [Demo days](../product/demo_day.md) in which we share progress and learn from each other.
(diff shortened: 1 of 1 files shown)


handbook/company/values.md
```diff
@@ -110,3 +110,20 @@ Our code and PRs are public, so lack of transparency is only obfuscation, not se
 ### Continuously grow
 
 You strive to continuously grow and learn by genuinely soliciting feedback early and often, and humbly reflecting on your past mistakes.
+
+We value growth at Sourcegraph in the form of continuous learning:
+
+- We actively seek out learning opportunities instead of avoiding them.
+- We identify gaps in our knowledge and share them openly and honestly.
+- We adopt a growth mindset.
```
nicksnyder, line 118: The first sentence here lists two things, but then we follow up with more things, which feels a bit weird and is maybe duplicative. It kind of feels like the first summary sentence needs to change.

Other minor points
- align on "You" vs "we"
- `We adopt a growth mindset.` feels like a circular definition, so recommend removing it

```suggestion
You strive to continuously grow and learn.

- You genuinely solicit feedback early and often.
- You humbly reflect on past mistakes.
- You actively seek out learning opportunities instead of avoiding them.
- You identify gaps in knowledge and share them openly and honestly.
```

nicksnyder (approved):
Thanks for demonstrating agency to improve our values, I love the specific examples!

handbook/company/values.md
```diff
@@ -110,3 +110,20 @@ Our code and PRs are public, so lack of transparency is only obfuscation, not se
...
+
+Here are some ways that we support learning, teaching, and knowledge-sharing at Sourcegraph:
```
felixfbecker, line 120: I'd add our 360º peer review cycles as a major point here, I think they're a big pillar of continuous growth and open feedback culture 🙂 

Also regular retrospectives (which we have a handbook page for too I think) which are a big part of continuously improving as teams and blameless culture.

felixfbecker (commented):
Great idea!

handbook/company/values.md
InesRoitman: @cassiemelani

InesRoitman (approved):
Love it! Thank you :) I agree with Felix comment. And also think this will have more detail once we have our L&D function ready

handbook/company/values.md
cassiemelani: Great idea @felixfbecker - I'll add some things here!

jeanduplessis (approved):

PR #4544, opened 2021-09-10
Gently remind authors of stale PRs

felixfbecker:
So I am normally NOT a fan of stale bot and I would NEVER introduce it to our main code repository for issues.
But hear me out: I think for our handbook repo, it would help a lot of folks.

We have many people joining every month that may have never used a pull request flow before or are not fluent with GitHub's UI and might completely forget to merge their PRs, never noticing their changes never went live. At the moment, we have over 125 pull requests that have been open for over a month.

I think the _perfect_ thing to do about this is whenever a PR has been open for 2 weeks without any review, comment, new commit pushed, etc, to post a friendly comment that reminds the author, which should then be delivered to their email inbox.
If the PR still sits still for another 7 days (no comment, review request, etc) it will be auto-closed. However, it can be easily brought back by reopening the PR if that was a mistake, nothing is _deleted_.

This is applied only to PRs, not issues.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,30 @@
+name: 'Close stale PRs'
+on:
+  schedule:
+    - cron: '30 1 * * *'
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v4
+        with:
+          days-before-pr-stale: 14
+          days-before-pr-close: 7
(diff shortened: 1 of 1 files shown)


marybelzer (approved):
@felixfbecker I really like this idea, and I think the Handbook is a great place for a bot like this. I see you have it set to update in #handbook-announce, I'm also going to find a place to add details about this in the Handbook for future reference. Thanks for looking into this!

PR #4569, opened 2021-09-13
Add Liz Fong-Jones DTT recap

csbailey5t:


blogposts/2021/dev-tool-time-liz-fong-jones.md
```diff
@@ -29,25 +29,25 @@ Liz's desk isn't the only place she works, though. Despite an ergonomic setup, L
...
+This automation goes hand in hand with Liz's attention to [observability](https://www.honeycomb.io/what-is-observability/), which emphasizes the capacity to introspect a system and understand its internal state. Automated builds and infrastructure configuration, using tools such as [Chef](https://www.chef.io/) and [Terraform](https://www.terraform.io/), increase the transparency of the systems with which Liz regularly works. As part of her work for [Honeycomb](https://www.honeycomb.io/), Liz uses Honeycomb, and shifting into its interface to understand workflows, processes, and debug errors or failures is a key part of her workflow.
```
rebeccadee, line 42: I usually go by the rule of using a hyphen where the phrase appears before a noun (acting as an adjective). I can get this into our handbook as I see it isn't specified: https://about.sourcegraph.com/handbook/communication/content_guidelines/style_and_mechanics#dashes-and-hyphens)

rebeccadee (approved):
@csbailey5t Made some minor tweaks. I love the addition of the video clips! Could we add captions to those? If someone is skimming it's not clear what the clips are about so we might be able to draw people in with a teaser of what it's about. (We can also try this in a future iteration if you're short on time for this one)

csbailey5t:
> @csbailey5t Made some minor tweaks. I love the addition of the video clips! Could we add captions to those? If someone is skimming it's not clear what the clips are about so we might be able to draw people in with a teaser of what it's about. (We can also try this in a future iteration if you're short on time for this one)

@rebeccadee by caption, you mean something like a `<figcaption>` rather than a VTT captions for the clip, right? I don't think that we can semantically use `<figure>` but I'll see what it looks like to add a super brief description of the video. We can also just add main text to describe it right before.

rebeccadee:
@csbailey5t 

> by caption, you mean something like a <figcaption> rather than a VTT caption for the clip, right? 

I think yes, there are some examples in this post: https://about.sourcegraph.com/blog/zoekt-memory-optimizations-for-sourcegraph-cloud/

csbailey5t:
@rebeccadee I switched to `figure` and `figcaption` for the video embeds. It feels like a bit of a stretch of the semantics of `figure` but probably still counts for how we're using the videos in the context of the document.

rebeccadee:
@csbailey5t Thanks for adding the figcaptions! 

> Do you think we need short explainers like in previous DTT recaps, or written out links as in the podcast recaps? Or is it fine as is?

I think whatever is going to be of more value to the reader is what matters. For some episodes, that might be going deep on the specific tools, and for others (I suspect for this one) talking about the guest's philosophy/approach is more interesting.

blogposts/2021/dev-tool-time-liz-fong-jones.md
```diff
@@ -0,0 +1,90 @@
...
+published: true
+---
+
+When [Liz Fong-Jones](https://www.lizthegrey.com/), Principal Developer Advocate at Honeycomb and experienced Site Reliability Engineer (SRE), joined us for [Dev Tool Time](https://info.sourcegraph.com/dev-tool-time) on September 1, 2021, she shared how her setup and workflow bridge the current state of development and the future that's already arriving.
+
+You can watch the recording of this episode at the [bottom of this blog post](#Watch-the-recording), but first let's check out Liz's main desk setup:
+
+## Hardware setup
+
+Liz's setup stresses ergonomics, and includes two monitors and a mechnical keyboard. Liz also uses a Blue Yeti mic and Logitech (?) webcam. On her desk are two computers: an AMD64 box and an ARM64 box with a 16 cores at 2.2GHz, 32 GB ram.
```
rebeccadee, line 20:
```suggestion
Liz's setup stresses ergonomics, and includes two monitors and a mechnical keyboard. On her desk are two computers: an AMD64 box and an ARM64 box with a 16 cores at 2.2GHz, 32 GB ram.
```

@csbailey5t I forgot to mention this in my review, but I think the ? is there because we aren't sure of the brand of webcam? It looks like an error or a note to ourselves so I'm going to remove this line just in case. We can add it back when we've confirmed.

PR #4575, opened 2021-09-14
Fix nested list on product plans

malomarrec:
The notebook does not render nested list of depth > 2, see [thread](https://sourcegraph.slack.com/archives/C021KFCA5MF/p1631206706016700).
Temporarily fix: remove some nesting, replace with headers.


jyavorska:
Thanks @malomarrec, and thanks for the link to the thread. That also seems fine as a permanent fix.

jyavorska (approved):

PR #4584, opened 2021-09-14
Remove unused delivery plans page

malomarrec:
We haven't used delivery plans, and dropped using tracking issues widely for nearly a year. This page is not maintained.

I came across this page when updating other content and was confused. I suggest we remove it (and maybe add a brand new page about how we release/deliver features when we have a workflow we want to use)


jyavorska (approved):

jyavorska:
Looks good to me @malomarrec, looks like there might be some inc links to fix. Please mention in the #product-internal channel after merge so everyone is aware, and maybe at the same time ask if anyone is actually using tracking issues anymore or if it can be deleted as well.

Thanks for taking time to keep the handbook accurate!

PR #4600, opened 2021-09-15
Reorder direction page

jyavorska:


malomarrec:
Love it!
My only (minor) worry is that we are mixing What's next and why, which should be ~ quarterly with Themes, which are likely to be longer term. If that's not a problem, we should just merge this!

malomarrec (approved):

jyavorska:
Lets merge and iterate in the main branch :)

PR #4617, opened 2021-09-16
Maturity tracker proof of concept

jyavorska:
This contains fake data but demonstrates driving a feature maturity tracker off of a JSON data file, just like how https://about.gitlab.com/direction/maturity/ works. This would be a potential alternative to a private spreadsheet like https://docs.google.com/spreadsheets/d/101JXaau2EPvi322AOFmNeoeuXSJqlruD8gBBsHl1fmI/edit#gid=0 which was also meant to solve the issue of lack of clarity around maturity levels for features.

- Are these the right maturity definitions? Maybe do we want fewer?
- GitLab tracks the dates different maturity levels were reached. Do we care?
- GitLab also tracks when _planned_ maturity increases are going to happen. That might be more interesting as it gives a sense of where we're focusing.
- Is there other metadata we'd like to show?
- Would we actually use the data on other pages, such that it makes sense to have it in a separate file (as opposed to just manually editing a maturity tracker)

The code is proof of concept level and would need to be rewritten for the real deal. Possibly this content makes sense to be generated at build time instead of run time as well.

image


handbook/direction/maturity/index.md
```diff
@@ -0,0 +1,62 @@
...
+
+- ○: not yet available, but on our roadmap
+- ◔: available in the product, but may not be ready for production use yet
+- ◑: usable for primary jobs to be done, but may lack depth
+- ●: a complete feature used extensively by our customers
+- ♥: a lovable, complete, and market leading experience
+
+<script>
+
+  const harvey_balls = ["○", "◔", "◑", "◕", "●", "♥"]
```
jyavorska, line 15:
```suggestion
  const harvey_balls = ["○", "◔", "◑", "●", "♥"]
```

handbook/direction/maturity/index.md
```diff
@@ -0,0 +1,62 @@
+# Feature Maturity Tracker
+
+Sourcegraph keeps track of how mature each of our feature areas are, to help you make the best decisions about how to use our product. Clicking on the feature name will take you to its documentation.
```
jyavorska, line 3:
```suggestion
Sourcegraph keeps track of how mature each of our feature areas are, to help you make the best decisions about how to use our product. Clicking on the feature name will take you to its documentation.

If you do not see content below, check you have JavaScript enabled then contact Product Management.
```

handbook/direction/maturity/index.md
```diff
@@ -0,0 +1,62 @@
...
+## Maturity levels
+
+- ○: not yet available, but on our roadmap
+- ◔: available in the product, but may not be ready for production use yet
+- ◑: usable for primary jobs to be done, but may lack depth
+- ●: a complete feature used extensively by our customers
+- ♥: a lovable, complete, and market leading experience
+
+<script>
+
```
jyavorska, line 16:
```suggestion
```

handbook/direction/maturity/index.md
```diff
@@ -0,0 +1,63 @@
...
+Sourcegraph keeps track of how mature each of our feature areas are, to help you make the best decisions about how to use our product. Clicking on the feature name will take you to its documentation.
+
+If you do not see content below, check you have JavaScript enabled then contact Product Management.
+
+## Maturity levels
+
+- ○: not yet available, but on our roadmap
+- ◔: available in the product, but may not be ready for production use yet
+- ◑: usable for primary jobs to be done, but may lack depth
```
felixfbecker, line 11: Is "may lack depth" gonna be true for every initiative? I can imagine that this progression may be specific to the product or team, where sometimes it makes sense to go depth-first, and sometimes breadth-first, and sometimes a different dimension entirely. For example, API docs seems to be going depth-first by focusing on one specific language first and making the experience _deep_ for that language before going _wide_ and supporting many languages. I think we should keep that flexibility for different initiatives.

handbook/direction/maturity/index.md
```diff
@@ -0,0 +1,63 @@
+# Feature Maturity Tracker
+
+Sourcegraph keeps track of how mature each of our feature areas are, to help you make the best decisions about how to use our product. Clicking on the feature name will take you to its documentation.
+
+If you do not see content below, check you have JavaScript enabled then contact Product Management.
+
+## Maturity levels
+
+- ○: not yet available, but on our roadmap
+- ◔: available in the product, but may not be ready for production use yet
```
felixfbecker, line 10: Curious which products you think would go here?
If I am the PM of a team, I would be very hesitant to put my product here because "not ready for production use" will likely deter customers from trying it out and implies a _risk_ that may not actually be realistically there, even if I _don't_ see the _next_ level already being true for my product ("usable for primary jobs to be done" -> maybe we don't actually support all primary jobs yet, but we _are_ confident about the reliability of what we support, there being little _risk_ and it therefor being "production-ready").

I guess my thought overall here (considering my comments below too) is it feels really hard and potentially dangerous to map all of our products into this one linear maturity scale.

handbook/direction/maturity/index.md
```diff
@@ -0,0 +1,63 @@
...
+- ◑: usable for primary jobs to be done, but may lack depth
+- ●: a complete feature used extensively by our customers
+- ♥: a lovable, complete, and market leading experience
+
+<script>
```
felixfbecker, line 15: I wonder if this generation code is worth it compared to just maintaining the "result" in Markdown? It seems more approachable to edit the markdown directly or look at the version history (don't need to find the JSON file to look at).
A JSON file with generator code can enforce consistency, but if there are some examples already present folks would probably intuitively follow the same format so I wouldn't be too worried about that.

felixfbecker (commented):
I have thoughts on the contents in this PR, but at a higher level it would be helpful to talk about what problems we're trying to solve with this/what the goal is 🙂 

E.g. are we trying to solve a problem where customers don't understand what they can adopt or not? I see some overlap here with our [defined product stages](https://about.sourcegraph.com/handbook/product/beta_and_experimental_feature_labels)s of experimental, beta, GA etc so if this is the goal it would be useful to talk about where that process is lacking currently so we don't end up with redundant or contradictory messaging.

Or is the goal here that we can _internally_ better track on a high level which "GA" features still need more or less investment, so we don't just "ship things and be done with them" (if that is a problem we have?)?

The GitLab page certainly looks really cool with all the graphs but I'm also lacking the context why _they_ invested in maintaining such a detailed view. Maybe you have some insight there?

handbook/direction/maturity/index.md
```diff
@@ -0,0 +1,63 @@
...
+
+If you do not see content below, check you have JavaScript enabled then contact Product Management.
+
+## Maturity levels
+
+- ○: not yet available, but on our roadmap
+- ◔: available in the product, but may not be ready for production use yet
+- ◑: usable for primary jobs to be done, but may lack depth
+- ●: a complete feature used extensively by our customers
+- ♥: a lovable, complete, and market leading experience
```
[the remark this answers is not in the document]
jyavorska: Sure, that's fine. Removed.

```suggestion
```

handbook/direction/maturity/index.md
jyavorska: Good point, added 'or breadth'.

```suggestion
- ◑: usable for primary jobs to be done, but may lack depth and/or breadth
```

handbook/direction/maturity/index.md
jyavorska: These would match our existing definitions of beta or experimental, linking to that from here.

```suggestion
- ◔: available in the product, but may not be ready for production use yet (i.e., [experimental or beta features](../../product/beta_and_experimental_feature_labels))
```

handbook/direction/maturity/index.md
jyavorska: Yeah, that's a good question mentioned in the description. The benefit of having content driven by data files is that it can be used in multiple places without having to update multiple places. This would be the first place this was used, but you could imagine info on maturity showing up in other places too.

jyavorska:
> I have thoughts on the contents in this PR, but at a higher level it would be helpful to talk about what problems we're trying to solve with this/what the goal is slightly_smiling_face
> 
> E.g. are we trying to solve a problem where customers don't understand what they can adopt or not? I see some overlap here with our [defined product stages](https://about.sourcegraph.com/handbook/product/beta_and_experimental_feature_labels)s of experimental, beta, GA etc so if this is the goal it would be useful to talk about where that process is lacking currently so we don't end up with redundant or contradictory messaging.
> 
> Or is the goal here that we can _internally_ better track on a high level which "GA" features still need more or less investment, so we don't just "ship things and be done with them" (if that is a problem we have?)?
> 
> The GitLab page certainly looks really cool with all the graphs but I'm also lacking the context why _they_ invested in maintaining such a detailed view. Maybe you have some insight there?

The potential problem to solve, and the reason GitLab made the page, is transparency to customers and sales/marketing teams about the relative maturity levels of different feature sets - a slightly different approach than https://docs.google.com/spreadsheets/d/101JXaau2EPvi322AOFmNeoeuXSJqlruD8gBBsHl1fmI/edit#gid=0 which is what we have today. This can drive sales and marketing decisions, as well as customers choosing what features make sense for them to adopt (or at least going into a newer/more experimental feature without being confused that it is more mature than it is.) It's not really anything to do with not finishing features after shipping them, or at least wasn't at GitLab.

Separate, but another potential benefit would be that GitLab also has a lot of pages that are autogenerated from this content around lists of features, feature comparisons, and so on that we could eventually do something like that. But that's not really part of this proof of concept.

handbook/direction/maturity/index.md
```diff
@@ -0,0 +1,62 @@
+# Feature Maturity Tracker
+
+Sourcegraph keeps track of how mature each of our feature areas are, to help you make the best decisions about how to use our product. Clicking on the feature name will take you to its documentation.
+
+If you do not see content below, check you have JavaScript enabled then contact Product Management.
+
+## Maturity levels
+
+- ○: not yet available, but on our roadmap
+- ◔: available in the product, but may not be ready for production use yet (i.e., [experimental or beta features](../../product/beta_and_experimental_feature_labels))
```
jyavorska, line 10:
```suggestion
- ◔: available in the product, but may not be ready for production use yet (i.e., [experimental or beta features](../../product/beta_and_experimental_feature_labels.md))
```

malomarrec:
I love the idea of replacing spreadsheets with handbook pages, and having a single-source of truth approach like this.

I wish we would keep one or the other (a spreadsheet, or a page). To do that, we would also need to have the sub-features in this tracker:
Screenshot 2021-09-16 at 14 58 43

For customers and the sales team, knowing that Batch Changes has "usable, but may lack depth" for Monorepos, is way more useful than knowing that Batch Changes overall is "usable, but may lack depth".

jyavorska:
> I love the idea of replacing spreadsheets with handbook pages, and having a single-source of truth approach like this.
> 
> I wish we would keep one or the other (a spreadsheet, or a page). To do that, we would also need to represent the maturity of the sub-features:
> 
> For customers and the sales team, knowing that Batch Changes has "usable, but may lack depth" for Monorepos, is way more useful than knowing that Batch Changes overall is "usable, but may lack depth".

Yeah agreed, so probably add one more click down if we do add this page. If we end up with both the new maturity page, and the spreadsheet, that's a fail.

jyavorska:
This PR is not mergeable due to the handbook refactor, closing. It can still be reviewed conceptually @christinaforney, and if we like it I can reimplement it in the new handbook.

PR #4627, opened 2021-09-16
Add clarity for adding images

marybelzer:
@InesRoitman I've added some details about adding images to the handbook, as I found the instructions a little confusing before. Can you review and see if this makes sense/if it seems like anything is missing? Thanks!


InesRoitman (approved):
This is definitely better! I do think there is some extra explanation needed. Mainly because I've never done it and I don't think I could do it just by reading this text. Particularly the Adding Images section sound a bit complicated to me.
I do think all of your additions made it better :)

marybelzer:
@InesRoitman Rearranged a bit based on your feedback, and updated that section to include some Handbook-specific language so people know a bit more what they're looking for. Thank you for your feedback!

PR #4666, opened 2021-09-17
Move Code Graph teams according to FY22Q3 reorg and create org page

felixfbecker:
This creates a folder and basic index page for the **Code Graph org** reflecting [last week's reorg](https://docs.google.com/document/d/1d8Z8zN6DjKHfXGaCQerKDeJo5qEVxBTku8RcZtw7Di4/edit).

I did not yet delete the `developer-insights` folder because there is quite a bit of content there that I think would be worth adding to general handbook pages, and pages for the other orgs don't exist yet so it would be a lot of work to do this for _all_ new orgs and teams. I only moved the `code-insights` team out of it and moved all the Code Graph teams together. I did add a note though that there is a migration in progress.

I added redirects for all pages so no links should be broken.

I added a basic mission statement [taken from the reorg document](https://docs.google.com/document/d/1d8Z8zN6DjKHfXGaCQerKDeJo5qEVxBTku8RcZtw7Di4/edit#heading=h.7g8rpai7u3bs) that serves as the index listing all teams & linking out to them.

Also updated every place that referenced the teams or the new org (without linking to it yet) and even some old places that had outdated info.


handbook/engineering/code-graph/index.md
```diff
@@ -0,0 +1,14 @@
+# Code Graph engineering org
+
+The Code Graph org's mission is to build the **code graph**.
+Each team within our org implements a different _verb_ on the code graph:
+
+- The [Search team](./search/index.md) enables _searching_ the code graph
```
yinquanteo, line 6: could we add the following nesting below this line and change "Search team" to "Search teams" – 

```
    - [Search Core team](./search/core/index.md)
    - [Search Product team](./search/product/index.md)
```

felixfbecker: Good idea

felixfbecker: Done

jyavorska:
Wow, thanks for making this big change @felixfbecker !

keegancsmith:
@felixfbecker this broke a whole bunch of links we have everywhere. We should probably setup redirects from the old links rather than fixing everything on the internet. For example google results fail (eg I searched `search core sourcegraph`), and the link to the team page in our engineering manager job post is now outdated. https://boards.greenhouse.io/sourcegraph91/jobs/4013257004

This is something I just noticed now, I am guessing there are lots of examples of this for any of these sorts of renames happening.

felixfbecker:
@keegancsmith thanks for noticing that – I don't understand why it's not redirecting, I had added redirects for all pages to the redirects file, including for that URL: https://github.com/sourcegraph/about/pull/4666/files#diff-ce703ecf6ce3dcbed963f8969cc977c74025db3589fde6d0ed732dab9e725732R76

Any ideas?

felixfbecker:
I don't understand this. If I run the handbook locally, the redirect works perfectly for that link...

felixfbecker:
Okay, I figured it out – thanks to @jyavorska for the tip. After redirects are added, the handbook/docsite process has to be restarted with `kubectl` to pick up the new redirects. Verified that they're now working as expected.

PR #4720, opened 2021-09-21
Add note about what in-product trials are

jyavorska:
In-product trials are related to beta and experimental features, but not the same thing. This adds a note to the beta and experimental features page.


jyavorska:
Closing since in-product trials don't actually seem to exist.

PR #4721, opened 2021-09-21
Add T.S. with start date

christinaforney:


pdubroy:
AFAIK, they will be working with both Frontend Platform and Dev Experience. Can you add under both?

Also, should we add to the separate team pages too?

jeanduplessis (changes requested):
T.S will be working across both Dev Experience and Frontend Platform.

PR #4741, opened 2021-09-22
Add guidance re: uploading images to GCP bucket

rebeccadee:
Best practices for uploading and storing images


rebeccadee:
FYI @nickmyyz this captures what we talked about yesterday!

PR #4743, opened 2021-09-22
Correct quotation marks in past DTT posts

rebeccadee:


rebeccadee:
FYI @content-hacker @jennybergen @csbailey5t @nickmyyz  – this doesn't replace [adding SEO metadata to blog posts](https://app.asana.com/0/1199711242382929/1201024700451906/f) but does solve our issue of Google using partial titles due to reading quotation marks incorrectly. When I was reviewing another post I realised we haven't been consistent with [our style guide](https://about.sourcegraph.com/handbook/communication/content_guidelines/style_and_mechanics#quotation-marks) and if we actually use the curly quotation marks and apostrophes, these aren't read as code, so don't interfere with the title 🎉

PR #4745, opened 2021-09-22
Submitting a GitHub issue and how the product team processes them

malomarrec:
- to raise an issue with a team, label it with `team/team-name`
- each team sets up the process they want from there to accommodate their tooling. Automation is recommended.
- the PM on feedback rotation is responsible for making sure there are no unlabeled issues


handbook/product/surfacing_product_feedback.md
```diff
@@ -12,14 +12,22 @@ You're welcome to tag teammates or teams, but that's not necessary – once the
...
-GitHub issues with the `feedback` label will automatically post to the #feedback Slack channel on creation; you do not need to manually copy the issue link into #feedback.
+1. Check if an issue for the feedback [already exists](https://github.com/sourcegraph/sourcegraph/issues?q=is%3Aissue). If there is already an issue with similar feedback, add a comment to that issue. If the issue is closed, read why and if it makes sense, re-open it with a comment.
+2. If there is no issue with your feedback, create a [new feedback issue](https://github.com/sourcegraph/sourcegraph/issues/new?assignees=&labels=feedback&template=customer_feedback.md&title=) or a [new bug issue](https://github.com/sourcegraph/sourcegraph/issues/new?assignees=&labels=&template=bug_report.md&title=) to share with the customer.
+3. Add the right team label (prefixed with `team/`) so that the issue gets [automatically](https://github.com/sourcegraph/sourcegraph/blob/main/.github/workflows/label-move.yml) added to the team's board for triage. GitHub issues with the `feedback` label will also automatically post to the #feedback Slack channel on creation; you do not need to manually copy the issue link into #feedback.
```
Joelkw, line 19: @malomarrec  the original intent of not having someone tag team was that it's very hard, and somewhat unrealistic as product team grows, for anyone outside product to always know the current org + ownership chart of the product org. I would suggest rephrasing this so that team seems optional – IIRC, the original intent of this process was to be as lightweight as possible so anyone else in the company felt comfortable following it rather than keeping the feedback to themselves. 
```suggestion
3. If you know which team(s) owns the product area(s) involved, you can add the right team label (prefixed with `team/`) so that the issue gets [automatically](https://github.com/sourcegraph/sourcegraph/blob/main/.github/workflows/label-move.yml) added to the team's board for triage. If you do not know, that's okay! GitHub issues with the `feedback` label will also automatically post to the #feedback Slack channel on creation and product teammates will help route them. 
```

serinadawn: I agree with @Joelkw on not asking that the person who submits feedback understand how to navigate our internal team structure.  I do wonder how we can internally triage these pieces of feedback and get them back to the right team members?

malomarrec: @Joelkw How about bugs?

PR #4751, opened 2021-09-22
Add meta tag specific title and description for blog posts

csbailey5t:
This PR adds optional fields to the frontmatter for blog posts: `seoTitle` and `seoDescription`. 

When present, these are used to set `<title>` and `<meta name="description" />`. When they aren't present, the values for `title` and `description` are used.

### DIFF
--- a/blogposts/2021/dev-tool-time-seth-vargo.md
+++ b/blogposts/2021/dev-tool-time-seth-vargo.md
@@ -1,6 +1,8 @@
 ---
 title: 'Dev Tool Time with Seth Vargo: Productivity hacks and .gitconfig tips'
 description: Check out the recording of our first episode of Dev Tool Time, in which Google Cloud Engineer Seth Vargo shares his tips for a productivity-optimized desk setup, efficient window management, and keyboard shortcuts.
+seoTitle: Productivity Hacks & .gitconfig Tips – Dev Tool Time
+seoDescription: We’re talking about setting up a new computer in record time on Dev Tool Time. Seth Vargo shares .gitconfig tips, favorite Mac window manager, and more.
 author: Vanesa Ortiz
 authorUrl: https://twitter.com/vanesacodes
 publishDate: 2021-04-28T10:00-07:00
--- a/blogposts/2021/dev-tool-time-seth-vargo.md
+++ b/blogposts/2021/dev-tool-time-seth-vargo.md
@@ -1,8 +1,8 @@
 ---
 title: 'Dev Tool Time with Seth Vargo: Productivity hacks and .gitconfig tips'
 description: Check out the recording of our first episode of Dev Tool Time, in which Google Cloud Engineer Seth Vargo shares his tips for a productivity-optimized desk setup, efficient window management, and keyboard shortcuts.
-seoTitle: Productivity Hacks & .gitconfig Tips – Dev Tool Time
-seoDescription: We’re talking about setting up a new computer in record time on Dev Tool Time. Seth Vargo shares .gitconfig tips, favorite Mac window manager, and more.
+externalTitle: Productivity Hacks & .gitconfig Tips – Dev Tool Time
+externalDescription: We’re talking about setting up a new computer in record time on Dev Tool Time. Seth Vargo shares .gitconfig tips, favorite Mac window manager, and more.
 author: Vanesa Ortiz
 authorUrl: https://twitter.com/vanesacodes
 publishDate: 2021-04-28T10:00-07:00
--- a/website/src/components/Layout.tsx
+++ b/website/src/components/Layout.tsx
@@ -42,11 +42,7 @@ export default class Layout extends React.PureComponent<LayoutProps> {
         return (
             <div className={`flex flex-column fill-height ${this.props.className || ''}`}>
                 <Helmet>
-                    {metaProps.seoTitle ? (
-                        <title>{metaProps.seoTitle}</title>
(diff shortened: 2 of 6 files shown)


website/src/components/blog/postTypes.tsx
```diff
@@ -18,6 +18,8 @@ export interface Post {
     frontmatter: {
         title: string
         description: string
+        seoTitle: string
+        seoDescription: string
```
felixfbecker, line 22: I would recommend documenting these fields here:
Also should mark these as optional with `?`.
```suggestion
        /** Controls the page's `<title>` for SEO and the browser tab label. Defaults to {@link title}. */
        seoTitle?: string
        /** Controls the page's `<meta name="description">` for SEO. Defaults to {@link description}. */
        seoDescription?: string
```

blogposts/2021/dev-tool-time-seth-vargo.md
```diff
@@ -1,6 +1,8 @@
 ---
 title: 'Dev Tool Time with Seth Vargo: Productivity hacks and .gitconfig tips'
 description: Check out the recording of our first episode of Dev Tool Time, in which Google Cloud Engineer Seth Vargo shares his tips for a productivity-optimized desk setup, efficient window management, and keyboard shortcuts.
+seoTitle: Productivity Hacks & .gitconfig Tips – Dev Tool Time
```
felixfbecker, line 4: The name "seoTitle" seems a bit misleading considering that `<title>` also controls the tab title 🤔

website/src/components/Layout.tsx
```diff
@@ -52,7 +55,12 @@ export default class Layout extends React.PureComponent<LayoutProps> {
                     <meta property="og:image" content={metaProps.image} />
                     <meta property="og:description" content={metaProps.description} />
```
felixfbecker, line 56: Should this also use `seoDescription` considering that some other link unfurlers are likely to look at `<meta name="description">` first? Same for `twitter:*` tags and `*:title` tags
```suggestion
                    <meta property="og:description" content={metaProps.seoDescription || metaProps.description} />
```

website/src/components/Layout.tsx
```diff
@@ -52,7 +55,12 @@ export default class Layout extends React.PureComponent<LayoutProps> {
...
 
-                    <meta name="description" content={metaProps.description} />
+                    {metaProps.seoDescription ? (
+                        <meta name="description" content={metaProps.seoDescription} />
+                    ) : (
+                        <meta name="description" content={metaProps.description} />
+                    )}
```
felixfbecker, line 62: Can be simplified :)
```suggestion
                    <meta name="description" content={metaProps.seoDescription || metaProps.description} />
```

website/src/components/Layout.tsx
```diff
@@ -40,7 +42,8 @@ export default class Layout extends React.PureComponent<LayoutProps> {
         return (
             <div className={`flex flex-column fill-height ${this.props.className || ''}`}>
                 <Helmet>
-                    <title>{metaProps.title}</title>
+                    {metaProps.seoTitle ? <title>{metaProps.seoTitle}</title> : <title>{metaProps.title}</title>}
```
felixfbecker, line 45: Can be simplified :)
```suggestion
                    <title>{metaProps.seoTitle || metaProps.title}</title>
```

felixfbecker (approved):
Would be useful to understand the motivation behind this a bit – what motivates having the title be different?

csbailey5t:
> Would be useful to understand the motivation behind this a bit – what motivates having the title be different?

As I understand it, the main motivation is separating the title displayed on the blog post page from what appears in Google search results to be more streamlined for the search viewer/reader and for the search algorithm to better surface posts. We could reverse this and explicitly set a display title for the post page on our site, but I'm not sure if one approach is better than the other.

website/src/components/Layout.tsx
csbailey5t: It wasn't clear from the Slack and sync conversations about this whether the `og:*` and `twitter:*` tags should reflect the search specific title/description, or the content on our site, so I was being conservative here. We can merge this for now, and I can follow up with stakeholders for more clarification.

blogposts/2021/dev-tool-time-seth-vargo.md
csbailey5t: This might be a compelling reason to have the "seoTitle" just be "title," and explicitly set the visual title that goes on the post page

felixfbecker:
Ah, I see. So basically having two titles:
- An **internal title and description** which is shown within the context of the blog: Page title header, blog post index list. 
- An **external title and description** which is shown outside of the blog: On Google, when link unfurling (posting to Twitter, Slack, LinkedIn, etc), browser tab bar. These may contain more context in the text since it is not shown within the context of the blog.

How about naming them this way? So making the `seoTitle` and `seoDescription` named `externalTitle` and `externalDescription`, respectively, and then following the above logic to decide where each one is used.
We could also reverse it, i.e. add new fields `internalDescription` and `internalTitle` to override those and making the `title` and `description` the _external_ and default ones for the internal fields.

csbailey5t:
Internal vs external is a great way to frame it. I think for now I'm inclined to switch to the first option, `externalTitle` and `externalDescription`, and if we do want the `og:*` and `twitter:*` tags to also switch, then move forward with the second version and rewrite as needed.

PR #4818, opened 2021-09-27
Add revenue considerations to prioritization model

jyavorska:
Our current prioritization guidance didn't mention revenue, this adds a bit about how we can think through the implications on the business of prioritization.


handbook/product/product_management/prioritizing.md
```diff
@@ -20,6 +20,19 @@ We receive tons of feature requests and bug reports, more than we can handle. Th
...
+
+- What is the story that ties my work to future revenue?
+  - What is the revenue opportunity? (more is better)
+  - How long will it take to realize that revenue? (sooner is better)
+- Who is the likely buyer of this feature, and how well does it correlate with our ideal customer profile?
+- How does this feature fit into our pricing model? Are parts of it free and paid and, if so, what is the upgrade path?
+- Are there other things we should consider doing that would have a bigger impact on the business instead?
+
+All of this is not to say that we don't prioritize features that drive monthly active users, or that we are myopically focused on driving revenue up. It's just that we are stewards of the business, and none of us would be here without building a solid, profitable business. To do so we mindfully balance revenue growth with user growth and features users immediately need with innovation in the future, and by ensuring that considering revenue implications is part of our prioritization process.
```
nicksnyder, line 34: A piece that feels missing here and maybe worth calling out: we want to build a long-term sustainable business and this requires making investments, experiments, and bets that we think will drive future revenue. Not all bets and experiments will be successful, and that is ok! What is important is that:
1. We continue making bets like these
2. These bets are prioritized and staffed proportionally to the opportunity (they shouldn't disproportionally take resources away from other things that are more likely/guaranteed to yield revenue on shorter time horizons).

nicksnyder (approved):
Thanks for drafting this up, looks good! I just have one minor comment.

handbook/product/product_management/prioritizing.md
jyavorska: Great addition @nicksnyder! I've integrated that content here as well.

jyavorska:
```suggestion
All of this is not to say that we don't prioritize features that drive monthly active users, or that we are myopically focused on driving revenue up. It's just that we are stewards of the business, and we want to build a long-term sustainable business and this requires making investments, experiments, and bets that we think will drive future revenue. Not all bets and experiments will be successful, and that is ok! What is important is that:

1. We continue making bets like these
1. These bets are prioritized and staffed proportionally to the opportunity (they shouldn't disproportionally take resources away from other things that are more likely/guaranteed to yield revenue on shorter time horizons).
```

PR #4874, opened 2021-09-29
Remove link from blog titles

rebeccadee:
It has always struck me as odd that the title on our blog post pages is a link to the same blog post 🤔  This is confusing for readers and serves no purpose that I can tell! This PR removes the link from blog post titles.

### DIFF
--- a/website/src/components/blog/BlogPost.tsx
+++ b/website/src/components/blog/BlogPost.tsx
@@ -37,9 +37,7 @@ export const BlogPost: React.FunctionComponent<Props> = ({
         <Tag className={`blog-post ${className}`}>
             <header className={headerClassName}>
                 <h1 className={titleClassName}>
-                    <Link to={url} className={`d-block ${titleLinkClassName}`}>
                         {post.frontmatter.title}
-                    </Link>
                 </h1>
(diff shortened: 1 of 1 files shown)


rebeccadee:
@felixfbecker Please let me know if you don't want to be my go-to reviewer 😆 If the link in the title serves a purpose I'm unaware of I'm happy to be wrong!

rebeccadee:
Aaaaah, I think I get it now. The titles are no longer clickable on the index page: https://deploy-preview-4874--sourcegraph.netlify.app/blog

We do have the "Read more" button but I wonder if there's any way to isolate the clickable titles on the card from the titles on the posts themselves?

csbailey5t:
@rebeccadee It's a bit tricky to track, but the `PostsList` (https://github.com/sourcegraph/about/blob/main/website/src/components/blog/PostsList.tsx#L21) component which renders the list of posts reuses the `BlogPost` component, which is why your change affects both single post pages and the list of posts. We would need to add some logic to render the title as a link or not depending on context or an explicitly passed option. 

Given the upcoming website rebuild, is it compelling enough to make this change now, or should it be noted in a design doc for the rebuild?

PR #4876, opened 2021-09-29
Update triage filter used by frontend-platform

valerybugakov:


pdubroy (approved):

pdubroy (changes requested):
Actually, I wondering if we should in fact track these issues on our board. What do you think? Might it not be useful for the team to be aware of what GitStart is working on?

valerybugakov:
Let's discuss it in Slack, and I will update the filter in case we decide to do so.

PR #4928, opened 2021-10-05
Add smartypants to auto-correct punctuation

csbailey5t:


marekweb (approved):
Awesome. I checked some posts and it seems to be working as expected, looks good!

PR #4960, opened 2021-10-25
Update maternity leave post with keyword

nickmyyz:
@vanesa These are the updates I'm planning! A few hundred people are searching "welcome back from maternity leave" every month, so if we add that keyword throughout the post, more people might see it!

@rebeccadee first of the optimized posts is ready for your review!


rebeccadee (commented):
@nickmyyz LGTM, just added one suggestion to make the case consistent with how we do things now. @vanesa do you want an opportunity to review Nick's changes before we merge?

nickmyyz:
> One other thought I just had, @nickmyyz – maybe as part of this process you could add some "More posts like this" to the end of these posts, since you're already editing them?

Ooh, that's a good idea! I'll add that to this and to the other posts going forward.

rebeccadee:
@nickmyyz @vanesa I'm going to merge this – the changes aren't substantial and I don't want to delay this longer. We can always make tweaks in a further PR 😃

rebeccadee:
@nickmyyz Been going through blog post-related issues in GitHub and came across this! When you have a moment could you add the related posts section? You should be able to merge this yourself but let me know if you are being asked for an approval first. Thanks!

nickmyyz:
@rebeccadee changes made! Two things: It looks like it does require a review and there are conflicting files. I'm not sure what do do about the latter, do you know what to do?

rebeccadee:
Paging @bretthayes – hoping this is quick to review/resolve. This (admittedly old) PR is just to update a blog post but we're getting conflicts on a .json file which is not my area of expertise 😅 Any ideas? 🙏

bretthayes:
No worries @rebeccadee! 😁  The PR just needed to be updated with the main branch since it fell a bit behind. LGTM! The merge conflict is fixed so you should be good to go! 🚀

PR #4973, opened 2021-11-03
Optimize and combine codebase overwhelm posts

nickmyyz:


nickmyyz:
@rebeccadee have you gotten a chance to look at this one? we have the redirect set up now so the posts are ready to be combined

rebeccadee (commented):
@nickmyyz This was on my list for today 😆 Sorry for the delay. I just pushed a fix for a link that wasn't going direct to right team page entry. 

I'm going to merge this so we have the content from the blog post that's now being redirected here. It is running very long now that we've combined the 2 (the leaders one was already pretty long!) so I wanted to ask if you would take another pass at trimming down the content to focus the piece? No rush, but I think we can integrate the content of the two more tightly and probably cut some of the scene-setting from the first piece. What do you think?

nickmyyz:
> @nickmyyz This was on my list for today 😆 Sorry for the delay. I just pushed a fix for a link that wasn't going direct to right team page entry.
> 
> I'm going to merge this so we have the content from the blog post that's now being redirected here. It is running very long now that we've combined the 2 (the leaders one was already pretty long!) so I wanted to ask if you would take another pass at trimming down the content to focus the piece? No rush, but I think we can integrate the content of the two more tightly and probably cut some of the scene-setting from the first piece. What do you think?

@rebeccadee I went through and did some general trimming, mostly the scene setting from part two, as you suggested, and integrated a few of the tips. Let me know what you think of it now!

PR #4974, opened 2021-11-03
Deprecate and redirect

nickmyyz:
@rebeccadee do you know what the best way is to delete this post and redirect traffic to the new one? 

maybe @csbailey5t knows? we're looking to combine two posts and redirect any lingering organic traffic from this one (the content of which is to be merged into another post https://github.com/sourcegraph/about/pull/4973 ) to the new one. I figure it has something to do with a 301 redirect but I don't know how to actually implement that 😅


csbailey5t:
@nickmyyz I would ask in the #any-questions Slack channel. I'm not sure if we need to redirect at the level of the repo, or on our host (Netlify). 

> @rebeccadee do you know what the best way is to delete this post and redirect traffic to the new one?
> 
> maybe @csbailey5t knows? we're looking to combine two posts and redirect any lingering organic traffic from this one (the content of which is to be merged into another post #4973 ) to the new one. I figure it has something to do with a 301 redirect but I don't know how to actually implement that 😅

rebeccadee:
Closing in favour of https://github.com/sourcegraph/about/pull/4990

PR #4992, opened 2021-11-15
Edit title on secrets blog post

rebeccadee:
This post hasn't been getting as much traffic as we'd hoped and I have a theory that the title isn't resonating because people aren't that familiar with what code search means yet. I'd like to tweak the title to focus on the problem people might be seeking to solve. I've made a change in this PR but would love to get your input @andreeleuterio on what someone might be looking for that would be answered by the post, and how we can write the title and description to appeal to those readers 🙏


andreeleuterio (approved):
Looks good!

PR #5105, opened 2022-01-31
Add internal telemetry

dadlerj:
@mgyorffy the tracker code here largely replicates what we use on the webapp, and sends anonymous pagesview events to the same data warehouse. The code was reviewed by Farhan and the frontend platform eng team. You can see it at https://github.com/sourcegraph/eventlogger. Let us know if you have any questions!


marybelzer:
@dadlerj Heads up that today is the last day of Mate's contract - I have asked the other engineers on the CPT to take a look at this so someone will review soon!

website/src/html.tsx
```diff
@@ -103,6 +103,13 @@ export default class HTML extends React.Component<HtmlProps> {
                         `,
                         }}
                     />
+
+                    {
+                        /* Add Sourcegraph tracker */ <script
+                            type="text/javascript"
+                            src="https://storage.googleapis.com/sourcegraph-assets/analytics/eventlogger.bundle.js"
+                        />
+                    }
```
bretthayes, line 112: Is there a specific reason this is wrapped in curly braces? Can this be written as follows?

```jsx
{/* Add Sourcegraph tracker */}
<script
    type="text/javascript"
    src="https://storage.googleapis.com/sourcegraph-assets/analytics/eventlogger.bundle.js"
/>
```

Also, it looks like the bundle isn't compressed and is 771 Kb. This could impact performance a bit. Is the event logger compressed before being deployed to GCP? Is this possible?

Another approach we could look at is shipping the Event Logger to npm `@sourcegraph/eventlogger` so we can import and ship it with the bundled production site.

bretthayes (changes requested):

website/src/html.tsx
dadlerj: Thanks! And great questions. Can you point me to any compression tools?

And: dumb question, but if I put it in NPM (which would be great, it'd be nice to have versioning), would it be possible to load it on this html.tsx page? I can't do a node-style require/include here, so without a `<script>` tag option, it just might not work.

bretthayes: That's actually a good question regarding importing. Since we're using Gatsby in the current stack, and Next JS in the future stack, if you export the Event Logger and publish it to an npm package, we'd able to import it here and it would get compressed when we build the site for production. So you actually wouldn't have to do an extra step to compress it in the Event Logger repo. Unless of course you want to support this style of importing via a script tag.

If you'd like to support the script style importing, I would use Webpack since it's a bundler and we can make use of [compression out of the box](https://webpack.js.org/guides/production/#minification). You can also support [different module methods](https://webpack.js.org/api/module-methods/).

You could also make this into a React Hook for react applications. 🤓

dadlerj: Thanks. I'd love to use NPM, but I honestly just can't figure out _how_ to import a script into this file using anything other than `<script>` notation (since I need the script itself to run on the output page, i.e. to load asynchronously, not to run in the Node context). I tried to just import this by dropping the eventlogger.tsx file into the repo, so it would get built as part of the site itself, but couldn't get it to work. 

Perhaps using Helmet? I'm sort of lost on how to actually make this work given the Gatsby structure and the way that this file is built, haha. 

In the meantime, I'll explore Webpack.

dadlerj:
Updated! And this should (?) be much, much smaller:

```
$ du -sh dist/main.js
8.0K	dist/main.js
```

Thanks for the advice to use webpack @bretthayes .

bretthayes (approved):
Sweet! Looks good as a first go!

As a follow up improvement for importing via script tags, we could probably cache bust the file with versions in the filename. I can open a PR for you if you want if I can get write access to the event logger repo. 🤓 

It would also be cool to see some CI/CD on merge in the future.

Let's ship it! 🚀

dadlerj:
Awesome, thanks! Just added you to that repo.

And yes, would love to get that set up. I'm definitely not the right long-term owner of this codebase or the ops around it. We are hiring more on our data eng/data ops team, and it may fall to them, or to someone on core eng (eventually). For now, though, any contributions or improvements would be much appreciated.

PR #5127, opened 2022-02-10
Temporarily remove mentions

rebeccadee:


Joelkw (approved):
Thanks @rebeccadee ! I think this may even be more than we need (some of these seem like harmless mentions, and the 3.32 post was made with this future consideration in mind already and is vague) but this definitely gets them! Defer to @amieroth ultimately, I'm fine with this either way.

rebeccadee:
@Joelkw I may have gone overboard 🙈 will let @amieroth decide what's best!

PR #5134, opened 2022-02-11
Add script to check Node version and install pinned dependencies

### LINKED ISSUE #5133: Add security measure to ensure dependencies match pinned Node version

katjuell:
Closes #5133 by ensuring that we have some means of guaranteeing that everybody's on the pinned version of Node for this project and installing the pinned dependencies.

## To Test
- Check out the branch and remove your `node_modules` in `website`
- Change your version of Node, run the install script `./script/install.sh`, make sure it exits.
- Set the `YARN` var in the script to something you don't have installed, make sure it exits.
- Ensure your dependencies install correctly.


katjuell:
@bretthayes I added a quick demo page for the code insights component in [70cc82c](https://github.com/sourcegraph/about/pull/5134/commits/70cc82c8180352c41a941a0ad0f0400bb19f3534), which I will revert — but curious if both of those graphs display for you on Safari under `/code-insights`. The `CodeInsightExampleType.Search` type does not for me (on Safari `15.0`). This does not seem to be an issue on staging, which has our pinned Node version and the correct dependencies, so I'm just curious if you can reproduce this (screenshot from an earlier iteration of the page):
Screen Shot 2022-02-09 at 10 44 41 AM

website/script/install.sh
```diff
@@ -0,0 +1,22 @@
...
+fi
+
+echo "Checking that yarn is installed"
+
+if [ ! -x "$YARN" ]; then
+    echo "Please install yarn"
+    exit 1
+fi
+
+echo "Install dependencies from the lock file"
```
bretthayes, line 20: Installing*

bretthayes (changes requested):
The Terraform versions chart displays for me in Safari 14 and 15 but not the Search

bretthayes (approved):

Issue #5118, opened 2022-02-07
Break points in get started flow


marybelzer:
[This Slack thread](https://sourcegraph.slack.com/archives/C02HL0RNTGV/p1644204976234379) resurfaced an issue that came up in the get started flow discussions a while back. Mate adjusted the size of these boxes to help a bit, but it's still a lot of white space on some screen sizes:
![image (2)](https://user-images.githubusercontent.com/73856596/152828599-e7b0c2b9-875a-4a8c-be5b-2b4798571768.png)

Looking for design input on how we should handle this and we can then have an engineer adjust the page. Thanks!

Joelkw:
Opening, this was not supposed to close – #5140 was a quick partial fix.

PR #5140, opened 2022-02-15
Quick update to the getting started page?

Joelkw:
[Here's what this change does and how it looks](https://www.loom.com/share/8805b6a2ba834ec0a9a3a26439f74530).

This seems to be a problem [here](https://sourcegraph.slack.com/archives/CNC4YAL1E/p1644182558618839) and [here](https://sourcegraph.slack.com/archives/C01SD4A49E1/p1644878053814279?thread_ts=1644867510.292599&cid=C01SD4A49E1). 

This quick fix is **not a final fix** – so it does *not* resolve [this issue](https://github.com/sourcegraph/about/issues/5118) IMO – because there's still a lot of white space on this page and I'm not sure what the design intention is long term, but it is a "short term improvement" fix

https://user-images.githubusercontent.com/11967660/154136058-d91da229-8401-4074-ad0d-46063c8a1fae.mp4

.


benvenker (approved):

katjuell (approved):
LGTM! Just need to make sure #5118 stays open so we we can have a designer review the flow

bretthayes (approved):
Thanks @Joelkw!

I agree with @katjuell. I would say this is a good quick fix for now until we get more design input via #5118.

PR #5189, opened 2022-03-09
Add tracking for buttons

### LINKED ISSUE #5108: Tracking on signup flow
This request came in from Erica Freckleton in Marketing, who handles tracking + analytics. We recently made changes to the sign-up flow on about.sourcegraph.com and need to be able to track user behavior. Below is a suggestion on how we tackle this - but either way, we need to establish some standards of how we implement tracking like this. Whoever takes this, I would recommend reaching out to Erica to work with her directly on the implementation as she'll be an ongoing partner for us on these types of things!

----

I'd have your devs add something like "cta-btn" to the <class> name on all the action buttons you want to track.  That way we could we create a simple trigger in GTM that just looks for {{Click Classes}} contains "cta-btn"... and we wouldn't need all that extra logic involving the {{Click Text}}, {{Page Path/URL}}, etc.

In addition, you could include some other info such as the style or location on the page.  If we embed that in an HTML data attribute ("data-"), we could use a data layer variable to parse it out.

For example, here's what the HTML code looks like for the "Continue with GitHub" button right now:

<a href="https://sourcegraph.com/.auth/github/login?pc=https%3A%2F%2Fgithub.com%2F%3A%3Ae917b2b7fa9040e1edd4&amp;redirect=%2Fwelcome&amp;_ga=2.158632569.2108224894.1643087719-145800987.1643087719" class="btn github custom-btn">Continue with GitHub</a>

You could add the following (see bold):

<a href="https://sourcegraph.com/.auth/github/login?pc=https%3A%2F%2Fgithub.com%2F%3A%3Ae917b2b7fa9040e1edd4&amp;redirect=%2Fwelcome&amp;_ga=2.158632569.2108224894.1643087719-145800987.1643087719" class="btn github custom-btn cta-btn">Continue with GitHub</a>

Note: 
Populating numbers for the data-button-style and data-button-location would keep things a little cleaner in the code.  However, you'd need look-up tables to translate the values into something meaningful.  For example:

data-button-style:
1 = White font on a black background
2 = White font on purple background

data-button-location:
1 = Header
2 = Body
3 = Footer

(or you could even be more granular/specific for #2)

Alternatively, you could embed the descriptions themselves (rather than codes).  But you'd need to keep the text pretty short/concise so that may not work too well if there's a lot of info you need to cram in.

— 

I like his approach with populating numbers for the data-button values to keep things clean, so long as we can keep a record of what aligns with what value.

katjuell:
Closes #5108 by adding tracking for buttons on the site. For more info about the data attributes, check out the discussion (with screenshots) in [this doc](https://docs.google.com/document/d/1Q1JMzyRmwNKk65KxxtT9EOnxoSf2dHEGkUAPbJkKzI4).

Note that this does **not** add tracking on GA. This will still need to happen.

Tl;dr on the schema in use for this PR:
**data-button-style**:
1 — Bootstrap primary button
2 — Bootstrap outline button
3 — ArrowRightIcon with text
4 — ArrowRightBoxIcon with text
5 — Bootstrap outline button with ArrowRightBoxIcon on the inside

**data-button-location**:
1 — Nav
2 — Hero
3 — Body (for body CTAs, we are only tracking calls to demo Sourcegraph. See the discussion doc for more details).
4 — ”Try Sourcegraph” sections. The sections that offer different callouts to try Sourcegraph, above the footer.

Pages that do not have tracking (but please let me know if they should!):
- `/jobs`
- `/resources/abcs-ebook`
- `/white-papers/remote-work-easier`

## To Test
- Navigate to some of the pages updated in the PR and check that the data attributes have been added.

### DIFF
--- a/website/src/components/Header.tsx
+++ b/website/src/components/Header.tsx
@@ -126,7 +126,9 @@ export default class Header extends React.Component<HeaderProps, any> {
                                     )}
 
                                     <Nav.Link
-                                        className="btn btn-outline-primary ml-3 px-5 py-2"
+                                        className="btn btn-outline-primary ml-3 px-5 py-2 cta-btn"
+                                        data-button-style="2"
+                                        data-button-location="1"
                                         href="https://info.sourcegraph.com/demo-request"
                                         title="Request a demo"
                                     >
@@ -136,6 +138,8 @@ export default class Header extends React.Component<HeaderProps, any> {
                                     {!this.props.hideGetStartedButton && (
                                         <Nav.Link
                                             className="btn btn-primary ml-3 px-5 py-2"
+                                            data-button-style="1"
+                                            data-button-location="1"
                                             href="/get-started"
                                             title="Get started"
                                         >
--- a/website/src/components/TrySourcegraph.tsx
+++ b/website/src/components/TrySourcegraph.tsx
@@ -17,17 +17,17 @@ export const TrySourcegraph: React.FunctionComponent<{ demoFormURL?: string; cla
             </div>
             <div className="col-md-6 pt-3 align-self-center text-center">
                 {demoFormURL !== '' && (
-                    <Link 
+                    <Link
                         className="btn btn-outline-secondary mx-2 mb-3"
                         data-button-style="2"
                         data-button-location="4"
-                        to={demoFormURL} 
+                        to={demoFormURL}
                         title="Request a demo"
                     >
                         Schedule a demo
                     </Link>
                 )}
-                <Link 
+                <Link
                     className="btn btn-primary mx-2 mb-3"
                     data-button-style="1"
                     data-button-location="4"
--- a/website/src/components/TrySourcegraph.tsx
+++ b/website/src/components/TrySourcegraph.tsx
@@ -17,11 +17,23 @@ export const TrySourcegraph: React.FunctionComponent<{ demoFormURL?: string; cla
             </div>
             <div className="col-md-6 pt-3 align-self-center text-center">
                 {demoFormURL !== '' && (
-                    <Link className="btn btn-outline-secondary mx-2 mb-3" to={demoFormURL} title="Request a demo">
+                    <Link 
+                        className="btn btn-outline-secondary mx-2 mb-3"
+                        data-button-style="2"
+                        data-button-location="4"
+                        to={demoFormURL} 
+                        title="Request a demo"
+                    >
                         Schedule a demo
                     </Link>
                 )}
-                <Link className="btn btn-primary mx-2 mb-3" to="/#get-started" title="Try Sourcegraph now">
+                <Link 
+                    className="btn btn-primary mx-2 mb-3"
+                    data-button-style="1"
+                    data-button-location="4"
+                    to="/#get-started"
+                    title="Try Sourcegraph now"
+                >
                     Try Sourcegraph now
                 </Link>
             </div>
--- a/website/src/pages/pricing.tsx
+++ b/website/src/pages/pricing.tsx
@@ -251,7 +251,13 @@ export default ((props: any) => (
                         </p>
                     </div>
                     <div className="col-md-6 pt-3 align-self-center text-center">
-                        <Link className="btn btn-primary mx-2 mb-3" to="../#get-started" title="Try Sourcegraph now">
+                        <Link 
+                            className="btn btn-primary mx-2 mb-3"
+                            data-button-style="1"
+                            data-button-location="4"
+                            to="../#get-started" 
+                            title="Try Sourcegraph now"
+                        >
                             Try Sourcegraph now
                         </Link>
                     </div>
--- a/website/src/pages/pricing.tsx
+++ b/website/src/pages/pricing.tsx
@@ -251,11 +251,11 @@ export default ((props: any) => (
                         </p>
                     </div>
                     <div className="col-md-6 pt-3 align-self-center text-center">
-                        <Link 
+                        <Link
                             className="btn btn-primary mx-2 mb-3"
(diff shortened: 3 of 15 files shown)


website/src/components/Header.tsx
```diff
@@ -126,7 +126,9 @@ export default class Header extends React.Component<HeaderProps, any> {
                                     )}
 
                                     <Nav.Link
-                                        className="btn btn-outline-primary ml-3 px-5 py-2"
+                                        className="btn btn-outline-primary ml-3 px-5 py-2 cta-btn"
```
bretthayes, line 129: I think we should be cautious about using classes as a reference point for tracking since it may not be scalable long-term. If we ever make use of markup and style compression techniques, it may obfuscate our classes similarly how scss modules uses hashes. Also, if we're not styling anything with the `cta-btn` class, I think it would be more appropriate to add this data to a data-attribute since it may confuse future eyes thinking there is a `cta-btn` style somewhere. What do you think about adding `data-button-type="cta"` instead?

website/src/components/Header.tsx
```diff
@@ -126,7 +126,9 @@ export default class Header extends React.Component<HeaderProps, any> {
...
+                                        data-button-style="2"
```
bretthayes, line 130: Could we add a constant file as a source of truth to map these numbers to variable names? Magic numbers can be confusing to read without documentation. What do you think about the following?

```js
const buttonStyle = {
  primary: 1,
  outline: 2,
  arrowWithText: 3,
  arrowBoxWithText: 4,
  arrowBoxButton: 5,
}
```

```js
const buttonLocation = {
  nav: 1,
  hero: 2,
  body: 3,
  trySourcegraph: 4,
}
```

```jsx
import { buttonStyle, buttonLocation } from 'tracking'

<button
  data-button-type="cta"
  data-button-style={buttonStyle.primary}
  data-button-location={buttonLocation.hero}
>
    My Button
</button>
```

bretthayes (changes requested):

website/src/components/Header.tsx
katjuell: Love it!

website/src/components/Header.tsx
katjuell: I would be 100% fine with this, for all of the reasons you've mentioned. Tagging @adammaxla here, since I believe he made the initial recommendations on #5189. Adam, would it be cool on your end if we did it this way?

website/src/tracking.ts
```diff
@@ -0,0 +1,14 @@
+export const buttonStyle = {
+    primary: 1,
+    outline: 2,
+    arrowWithText: 3,
+    arrowBoxWithText: 4,
+    arrowBoxOutlineButton: 5,
```
bretthayes, line 6: What do you think about calling this `arrowBoxOutlined` since we know it's a button via `buttonStyle` 💡

katjuell: Works for me!

bretthayes (commented):
Looks great! Just waiting on a reply from @adammaxla regarding `data-button-type="cta"` versus the class approach.

adammaxla:
> 

I am good with this - I don't remember making the suggestion but I think it makes sense.

adammaxla:
@katjuell @bretthayes  I am good with this convention, I honestly don't remember making this recommendation but I think it makes sense.

marybelzer:
@katjuell is this ready to be shared with marketing stakeholders?

katjuell:
@marybelzer on the code side it should be good to go — just needs another +1. On the backend/ GA side, the functionality still needs to be implemented.

marybelzer:
@katjuell cool, I will share with Erica. She will then manage the implementation on the GA side. Thanks!

website/src/pages/pricing.tsx
```diff
@@ -255,7 +256,14 @@ export default ((props: any) => (
                         </p>
                     </div>
                     <div className="col-md-6 pt-3 align-self-center text-center">
-                        <Link className="btn btn-primary mx-2 mb-3" to="../#get-started" title="Try Sourcegraph now">
+                        <Link
+                            className="btn btn-primary mx-2 mb-3"
+                            data-button-style={buttonStyle.primary}
+                            data-button-location={buttonLocation.trySourcegraph}
+                            data-button-type="cta"
+                            to="../#get-started"
```
bretthayes, line 264: Is `../#` a typo? I think it should be `/get-started`

website/src/components/TrySourcegraph.tsx
```diff
@@ -17,11 +18,25 @@ export const TrySourcegraph: React.FunctionComponent<{ demoFormURL?: string; cla
...
                         Schedule a demo
                     </Link>
                 )}
-                <Link className="btn btn-primary mx-2 mb-3" to="/#get-started" title="Try Sourcegraph now">
+                <Link
+                    className="btn btn-primary mx-2 mb-3"
+                    data-button-style={buttonStyle.primary}
+                    data-button-location={buttonLocation.trySourcegraph}
+                    data-button-type="cta"
+                    to="/#get-started"
```
bretthayes, line 37: I think this should be `/get-started`

bretthayes (changes requested):
Just a few small changes I just noticed.

Also, if we ever add other button types in the future, I guess at that point we can abstract that logic into the tracking variables file and add `buttonTypes`.

website/src/pages/pricing.tsx
katjuell: Ohh yeah that's a typo! Thanks for catching.

bretthayes (approved):
Awesome! Lgtm!

PR #5191, opened 2022-03-10
Move animation resources to be local

katjuell:


felixfbecker (changes requested):
We can't add videos to the repo, they are way too large and bloat the Git repository size forever (they would be in the history forever).

This PR should fix the issue too: https://github.com/sourcegraph/about/pull/5192

PR #5192, opened 2022-03-10
Ignore insights self-hosted video from Cookiebot

felixfbecker:
Cookiebot blocks cross-domain videos by default until consent is given because videos _might_ set cookies. This breaks the animation from autoplaying on the first page visit. However, we know that GCS does not set cookies (Cookiebot just doesn't know) and it's not an iframe so it can't use local storage, so we can safely disable the blocking for this video.

### DIFF
--- a/website/src/pages/code-insights.tsx
+++ b/website/src/pages/code-insights.tsx
@@ -340,6 +340,8 @@ export const CodeInsightsPage: React.FunctionComponent<PageProps> = props => (
                         loop={true}
                         playsInline={true}
                         controls={false}
+                        // GCS does not set cookies, so we don't want Cookiebot to block this video based on consent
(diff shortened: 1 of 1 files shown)


coury-clark (approved):

rebecca-rissinger:
@felixfbecker thanks for sharing this! To confirm, the only change we are making is disabling the blocking for videos across sites? Was this change small enough that I would not need to do any QA? Or do you think it'd be best if I ran a few tests on my end?

katjuell (approved):
lgtm!

felixfbecker:
It only changes the blocking for this specific video (that currently doesn't load). I checked everything looks alright in the deploy preview: https://deploy-preview-5192--sourcegraph.netlify.app/code-insights

PR #5201, opened 2022-03-11
Update iframe videos

katjuell:
Removes `data-cookieconsent="ignore"` from iframed videos.

## To Test
- Check records in PR in the deploy environment in incognito mode and ensure that they load properly.


bretthayes (approved):
Looks good! Tested and working on all pages!

PR #5202, opened 2022-03-14
Final edits to onboarding post

rebeccadee:
Melissa's final edits came in late as she was off sick. Pushing these now instead


rebeccadee:
@nickmyyz Just some small edits – I would merge myself but this new approval requirement is preventing me 🙃

PR #5179, opened 2022-03-07
CloseFactor Social Pages

bretthayes:
- Creates two pages and closes #5107
- Adds a few new colours to our colours sass partial
- Adds a few new mixins and background utility classes to our gradients sass partial

Notes:
1. Someone from marketing will be styling the forms in HubSpot otherwise we run the risk of overwriting styles and having conflicts in HubSpot styling if they're ever changed in HubSpot and not in code.
2. The YouTube videos in the Hero's are private so they aren't displaying atm.

### How to test
- Visit the [accelerate-dev-onboarding](https://deploy-preview-5179--sourcegraph.netlify.app/accelerate-dev-onboarding) page
- Visit the [fixing-vulnerabilities](https://deploy-preview-5179--sourcegraph.netlify.app/fixing-vulnerabilities/) page

Issue #5106, opened 2022-02-01
CloseFactor (previously Fast Start) Social Design


marybelzer:
We need design and dev support to launch a target social campaign early in FY23. The scope includes: design a page template and produce four landing pages.

- Figma File (link when ready)
- [Copy doc including outline](https://docs.google.com/document/d/1xczYELbgcxrbubwHmW61BD4AOx0CcUou5vXIGIRZ-XM/edit)

Marketing has asked us to set a delivery date based on our current workload, and they are planning on two design reviews. @stephzabala can you put an estimated timeline in the comments and I will update expectations with Marketing? This is not an urgent request, just want to give them an estimate with your other work in mind. Thanks!

[Asana Card](https://app.asana.com/0/1200095757619647/1201745821158096/f)

Schedule:

- [ ]  Draft 2021-10-02
- [ ] First review 2021-18-02

fabicastp:
Thank you, Mary and Steph! We're looking for a very simple design, similar to https://info.sourcegraph.com/demo-request - I don't expect more than one round of review. 

The request isn't urgent but we would like to launch this as early on Q1 as possible :) Thank you!!

marybelzer:
This will be handled by Grafik

marybelzer:
@bretthayes We are seeing an error where the page is just stuck on loading for some folks. One of our stakeholders contacted Chilipiper support, who said this in response:

> This almost always indicates that our code is on the page twice. I inspected the page and do see it on there twice, can you remove one version?

Can you take a look? Thanks!

bretthayes:
@marybelzer yes for sure! I suspect I know what the problem is. I'll take a look asap!

PR #5208, opened 2022-03-19
HubSpot/ChiliPiper multi-form bug fix

bretthayes:
### What should this PR do?

This fixes a bug where HubSpot and ChiliPiper would have unexpected loading results due to the scripts being loaded more than once on the same page. This caused the ChiliPiper scheduler to hang ([noted here](https://github.com/sourcegraph/about/issues/5106#issuecomment-1072883997)).

This solution improves the ability to have [multiple HubSpot forms](https://github.com/sourcegraph/about/pull/5179/files#diff-eed8c0ceaaee7bcf8769e4a7cb2dce59af170c8e1a48c27aef3695fd88004de4R9-R11) of the same type on a single page where designs request it.

### How to test

Check out the following pages:
- /fixing-vulnerabilities
- /accelerate-developer-onboarding

Test each form with `yourEmailPrefix+testing@sourcegraph.com and ensure the form submits and the ChiliPiper scheduler displays in a modal and doesn't hang with a loader.

### DIFF
--- a/website/src/hooks/hubSpot.ts
+++ b/website/src/hooks/hubSpot.ts
@@ -64,20 +64,39 @@ interface HubSpotForm {
     onFormReady?: ($form: HTMLFormElement) => void
 }
 
-function createHubSpotForm({ region, portalId, formId, targetId, onFormSubmit, onFormReady }: HubSpotForm): void {
-    // Load HubSpot script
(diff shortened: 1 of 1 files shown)


katjuell (approved):
lgtm!

PR #5213, opened 2022-03-23
Add review section to Careers page

### LINKED ISSUE #5056: Add reviews section to Careers page
The reviews are under "Left to do" in the [Careers page QA doc (requirements)](https://docs.google.com/document/d/1pn2le8VFiQE-ZuFSnVeqExsI5BRUF5knEi5slu0t_VM/edit)

---
[Asana ticket](https://app.asana.com/0/1200095757619647/1201611726728716) for PM

st0nebraker:
Closes #5056 by adding a glassdoor reviews section to the Careers page

### To Test
- Navigate to `/jobs` to find new Reviews carousel
- (There is no design spec)

image

### Notes
- There was no design spec
- No order specified. I placed this between the values & apply to jobs section. Let me know if it should be elsewhere


website/src/pages/jobs.tsx
```diff
@@ -316,6 +309,31 @@ const CareersPage = props => {
                     </Carousel.Item>
                 </Carousel>
 
+                <Carousel
+                    className="reviews-carousel"
+                    prevIcon={<ArrowLeftIcon color="#808080" />}
+                    nextIcon={<ArrowRightIcon color="#808080" />}
+                >
+                    <Carousel.Item>
+                        <img src="../careers/review1.png" />
```
bretthayes, line 318: I know this page already contains a lot of repetitive manual markup that could be abstracted into components and used in loops, but as an improvement, you could refactor and simplify this given the url structure where the assets live to something like this:

```jsx
{new Array(5).fill(null).map((el, i) => (
    <Carousel.Item key={`review-${i}`}>
        <img src={`/careers/review${i}.png`} />
    </Carousel.Item>
))}
```

You'll have to rename your assets to start at `0` given the loop.

Also, this is out of scope, but feel free to change all of our urls on this page to be absolute paths instead of relative paths since assets are served from the root directory. Notice Gatsby builds everything under `public` when you run `yarn build` 🤓

website/src/css/pages/_jobs.scss
```diff
@@ -245,6 +245,42 @@
...
+
+        .carousel-item {
+            text-align: center;
+            padding: 4%;
+            background-color: #fff;
+            height: 800px;
+
+            @media (min-width: 500px) and (max-width: 768px) {
+                height: 570px;
+                padding: 0px;
```
bretthayes, line 260: Whenever you declare `0` values, you don't have to specify the unit of measurement so you can omit the `px`. Typically, prettier would automate and catch things like this for us 🙂

bretthayes (changes requested):
Nice work! Just a few comments 🙂 

Also since there was no design spec for this, we should get a stakeholder to sign off on this @marybelzer

website/src/pages/jobs.tsx
st0nebraker: Ah, I was thinking of doing a loop but bypassed it. Thanks!

st0nebraker: Ah, I was thinking of doing a loop but bypassed it. Thanks!

website/src/css/pages/_jobs.scss
```diff
@@ -245,6 +245,42 @@
         }
     }
 
+    .reviews-carousel {
```
katjuell, line 248: I would try to see how many of these styles you can set using Bootstrap utility classes in the `tsx` itself. We've got custom heights and max-widths now as well! Using the built-in breakpoints will let you avoid writing media queries in many cases as well! So you could do `px-lg-5 px-md-3 px-0` to set horizontal padding on an element at the set breakpoints, for example.

katjuell (dismissed):
Awesome work! And this was good context for me as well, to catch up with another carousel (in this case, the default Bootstrap one) in use around the site (tl;dr a lot of my work recently has been on carousels! Of which we have many!).

Let me know if you have any questions :)

website/src/css/pages/_jobs.scss
st0nebraker: Thanks! Can you explain our custom heights and max-widths? I had a hard time finding the documentation on Bootstrap. I'm using the utility classes for all the background & spacing styles here now though 😄

katjuell: Yay! Awesome :)

Agreed that the custom stuff isn't well-documented! For height, you can do `height-<whatever>` (including `auto`) — there are defined breakpoints as well, which align with Bootstrap defaults. The `_spacing` file will show you what's available.

For max width, it would be `mw-<whatever>` — no defined breakpoints, but feel free to add them if you need them!

website/src/css/pages/_jobs.scss
```diff
@@ -246,27 +246,19 @@
     }
 
     .reviews-carousel {
-        padding: 0 0 100px 0;
         background-color: #fff;
```
katjuell, line 249: Can swap this out for `bg-white` as well — [Bootstrap default](https://getbootstrap.com/docs/4.0/utilities/colors/#background-color)!

st0nebraker: Whoops, I did before but didn't commit it I guess. It's in now 👍

website/src/css/pages/_jobs.scss
st0nebraker: Neat, thank you so much! Okay this change should be good now :)

website/src/pages/jobs.tsx
```diff
@@ -310,12 +298,30 @@ const CareersPage = props => {
...
 
+                <Carousel
+                    className="pb-lg-7 bg-white"
+                    prevIcon={<ArrowLeftIcon color="#808080" />}
+                    nextIcon={<ArrowRightIcon color="#808080" />}
+                >
+                    {new Array(6).fill(null).map((el, i) => (
+                        <Carousel.Item
+                            key={`review-${i}`}
+                            className="pt-sm-6 pt-5 height-800 height-md-650 height-sm-575"
```
bretthayes, line 315: I think we should vertically center the child img elements since they're all different heights

bretthayes (changes requested):

website/src/pages/jobs.tsx
st0nebraker: The Bootstrap React Carousel component is being stubborn with this, because there's a nested wrapper (`.carousel-inner`) above the `Carousel.Item` and it's positioned `relative`. So to get the vertical alignment I need to bring the [custom class back in](https://github.com/sourcegraph/about/pull/5213/commits/8c1f43535d8e6f99db5c7d61464f87e42803ef85), not using the Bootstrap classes because they can't reach it.

But also with this, I'm seeing the images bounce around sometimes, are you? >:( Any ideas what to do from here?

elzannewentzel: @st0nebraker @bretthayes Is it possible to make them all the same size to avoid that overlap over the carousel nav at the bottom?
Screenshot 2022-03-31 at 15 51 44
?

website/src/css/pages/_jobs.scss
```diff
@@ -245,6 +245,26 @@
         }
     }
 
+    .reviews-carousel {
+        height: 800px;
```
katjuell, line 249: It should be possible to clear out some more of these styles by using our `height-<whatever>` custom spacings (they also have breakpoints, which would handle the media queries below)

website/src/css/pages/_jobs.scss
```diff
@@ -245,6 +245,26 @@
...
+        display: table;
```
katjuell, line 250: `d-table` in Bootstrap

website/src/css/pages/_jobs.scss
```diff
@@ -245,6 +245,26 @@
...
+        width: 100%;
```
katjuell, line 251: `w-100` in Bootstrap

website/src/css/pages/_jobs.scss
```diff
@@ -245,6 +245,26 @@
...
+        @media (max-width: 768px) {
+            height: 775px;
+        }
+        @media (max-width: 576px) {
+            height: 650px;
+        }
+
+        .carousel-inner {
+            display: table-cell;
```
katjuell, line 260: `d-table-cell` in Bootstrap

website/src/css/pages/_jobs.scss
```diff
@@ -245,6 +245,26 @@
...
+            vertical-align: middle;
```
katjuell, line 261: `align-middle`

website/src/css/pages/_jobs.scss
```diff
@@ -245,6 +245,26 @@
...
+            .carousel-item {
+                padding-bottom: 10%;
```
katjuell, line 263: This you could do via `pb-<whatever-makes-sense-by-looking-at-it>`

website/src/css/pages/_jobs.scss
bretthayes: Just thought I'd let you know @katjuell @st0nebraker, I [enhanced these](https://github.com/sourcegraph/about-replatform/pull/67/files#diff-d5059f7f3296770e92baba16d5e81763f80d55d19610817a103e1801725c641dR20-R38) a bit over in the replatform that could make this a lot easier on us until we get on Tailwind 😄 

So you should have these available to you if we port these over here:

breakpoint is `sm`, `md`, `lg`, `xl`
spacing is 100px–1200px

```scss
.h-#{$breakpoint}-auto {
    height: auto;
}
.h-#{$breakpoint}-#{$spacing} {
    height: #{$spacing}px;
}
.min-h-#{$breakpoint}-#{$spacing} {
    min-height: #{$spacing}px;
}
.max-w-#{$breakpoint}-#{$spacing} {
    max-width: #{$spacing}px;
}
```

website/src/css/pages/_jobs.scss
st0nebraker: `.carousel-inner` is another tag added by Bootstrap in between `Carousel` and `Carousel.Item` though 🤔  , and doesn't seem I can get to it in the template by something like `Carousel.Inner`. Any idea how I can add this utility class inline to this hidden element from the template?
image

website/src/pages/jobs.tsx
st0nebraker: Omg, the extra <div> wrapper for centering FTW 🤦‍♀️  Thank you Brett 🙌

website/src/css/pages/_jobs.scss
st0nebraker: Disregard, obvious solution of adding an extra wrapper to the CarouselItem did it 🤦‍♀️  -- Custom styles here are nixed!

bretthayes: Awesome! Thanks for nuking the styles 🎉

website/src/pages/jobs.tsx
```diff
@@ -310,12 +298,29 @@ const CareersPage = props => {
...
+                    className="bg-white height-800 height-sm-800 pt-md-0 pt-1"
+                    prevIcon={<ArrowLeftIcon color="#808080" />}
+                    nextIcon={<ArrowRightIcon color="#808080" />}
+                >
+                    {new Array(6).fill(null).map((el, i) => (
+                        <Carousel.Item key={`review-${i}`}>
+                            <div className="d-flex align-center height-750 height-sm-650">
+                                <img
+                                    src={`/careers/review${i + 1}.png`}
+                                    className="d-block m-auto col-11 col-lg-5 mw-lg-600 mw-sm-500"
```
bretthayes, line 317: You could probably get away with `mw-600`. It would solve the issue of the images underlapping slide bullet point UI and the images from not stretching too far on wider screens. And then you could remove those spacing styles in this PR since we'll be porting over the new spacings 🙂

bretthayes (changes requested):

website/src/pages/jobs.tsx
st0nebraker: I need `col-11 mw-sm-500 mw-600` for the mobile sizing though. Oh yea I'll port over your spacing changes now 🌟

bretthayes (approved):

courier-new:
Hey there!! First off, I 💙 this page and I love the idea for this enhancement to it. Also, congrats @st0nebraker for getting your first PR merged!! 🙌 

Apologies for coming in with this feedback super late, but I only found this after I saw the changes already went live. 😅 I read from the Careers page QA doc that the initial suggestion from marketing was either to include screenshots or else just take quotes from these reviews and link back to Glassdoor with them. Would we consider actually going the quotes route here instead? I'm not sure the screenshots in a carousel are a good idea for a couple reasons:

- They're not accessible. This is the biggest reason for me. Since none of these images have alt text, when a screen reader encountered them, most would just read the file names out. Since there isn't any text around the carousel contextualizing the images, either, this would make for an extremely poor experience for someone using a screen reader. Since the text in the screenshots is captured at a fixed resolution, if someone has set their default font size to be a lot bigger, these images would still be stuck at a fixed, very small font size and would only get blurrier as the user tried to zoom in on the image.
- There's an awful lot of text in each image for it to be part of an auto-progressing carousel. If I don't stop the carousel, I can't read a full review before it's already progressed to the next one. I think if we want to include the full reviews, it'd make more sense to arrange them as a waterfall of cards, kinda like this example:
	
	That way you could still read all the text. I'm not opposed to keeping it in carousel-form, either, but if we do, I think keeping only the most meaty quote from each review and just providing a link back to the original review on Glassdoor would be more usable.

Again sorry for kinda popping up out of nowhere with this feedback, but I'm happy to talk about this more or collaborate to make it happen, if that would help. Since this page might be the first impression that a lot of people get about working at Sourcegraph, I want to make sure it's making the best possible one, especially in terms of accessibility! 🙂

bretthayes:
Thanks @courier-new!! Totally agree with all of your points here and definitely think a text-based carousel would be a better approach here. I think we could probably use our QuoteCarousel component to slide through all of the reviews instead of using screenshots. I've created a follow up ticket #5246 with your suggestion. Thanks again for the feedback!

st0nebraker:
Thanks for raising this @courier-new! I appreciate the feedback and it's a good reminder to stay vigilant on a11y standards. Working on #5246 today to address this.

elzannewentzel:
@courier-new Can you review[ these changes](https://deploy-preview-5248--sourcegraph.netlify.app/jobs) and let us know if it's good to go, please?

PR #5223, opened 2022-03-28
Update onboarding use cases page

katjuell:
Add feedback from the [doc](https://docs.google.com/document/d/1HR3O5IL1dNtoME0-b3jfSho7Q-yQrOqHJd5r-UjTv88)

## To Test
- Check out the onboarding use cases page and ensure that: logos are no longer company links and the copy changes in the carousel match the requirements in the doc.


website/src/css/pages/__case_studies.scss
```diff
@@ -72,6 +72,14 @@
                 font-size: 1.2rem;
                 line-height: 1.5;
             }
+
+            &--section {
```
katjuell, line 76: Ended up adding this to deal with a nested `h5` on `p` warning and figured that this was the least destructive way to not interfere with the existing case studies styles (border currently gets applied to all `p` elements under `case-studies__quote--in-content`)

website/src/components/Blockquote.tsx
```diff
@@ -23,10 +22,10 @@ export const Blockquote: FunctionComponent<{
                 }
             >
                 {border && headline ? (
-                    <p className="font-weight-normal">
+                    <div className="case-studies__quote--in-content--section">
```
bretthayes, line 25: In the replatform, we should make use of `blockquote`, `figcaption` and `cite` elements for more semantic markup instead of using divs.

bretthayes (approved):

bretthayes:
Updated our replatform ticket sourcegraph/about-replatform#41 to include these changes so we don't forget.

website/src/components/Blockquote.tsx
katjuell: Sounds good, yeah. It would be nice to have a better and more consistent way of dealing with the border as well

katjuell:
@bretthayes I noticed that the alt was hardcoded (:mybad:) so updated

bretthayes (dismissed):

bretthayes:
@katjuell I know this is a bit of scope creep, but do you want to remove the links to the logos in the fixing vulns page carousel as well? There was a comment [in the doc](https://docs.google.com/document/d/1HR3O5IL1dNtoME0-b3jfSho7Q-yQrOqHJd5r-UjTv88/edit) about it.

bretthayes (approved):

PR #5228, opened 2022-03-29
Add link to onboarding guide

rebeccadee:


rebeccadee:
@efreckelton 👋 I merged this yesterday but now I'm wondering about something and would like your feedback please :) 

I added a CTA to the onboarding guide at the end of the post, but we also have two CTAs in the body of the post. Is it bad to have so many CTAs and should we be prioritising the onboarding guide because it's gated? Or is it good to have a few options for people to explore more?

PR #5248, opened 2022-04-07
Replace Glassdoor Review screenshots with QuoteCarousel

### LINKED ISSUE #5246: Replace the review screenshots on our careers page to be text-based
As a follow up to #5213, let's replace our review screenshots to be text based quotes as an improvement towards accessibility standards.

st0nebraker:
Closes #5246. Replaces long quote screenshot's with QuoteCarousel for accessibility.

### Test:
Navigate to `/jobs`

### Open questions:
Any other indicator we want for design to point out these are Glassdoor reviews? Or is the link to full review sufficient?

image


website/src/pages/jobs.tsx
```diff
@@ -3,9 +3,55 @@ import ArrowRightIcon from 'mdi-react/ArrowRightIcon'
...
+        {
+            header: 'Fast paced startup that cares',
+            quote: 'Great teams full of extremely smart people. Management that cares about the well being of its employees. Full remote and async culture.',
+            by: 'Current Employee, Feb 4, 2021 - ',
+            linkText: 'Read full review',
+            link: 'https://www.glassdoor.com/Reviews/Employee-Review-Sourcegraph-RVW41989078.htm',
+        },
+        {
+            header: 'Great start to an awesome adventure',
+            quote: '- Professional at a level many pre-IPO companies are not - Really good product that is easy to get behind and support - Dedicated leadership that support all employees - Everyone is motivated and pulling in the same direction',
```
bretthayes, line 34: Can you remove the dash and space `- ` at the beginning of this quote?

website/src/pages/jobs.tsx
```diff
@@ -304,22 +350,11 @@ const CareersPage = props => {
...
-                            <div className="d-flex align-center h-750 h-sm-650">
-                                <img
-                                    src={`/careers/review${i + 1}.png`}
-                                    className="m-auto col-11 max-w-sm-500 max-w-600"
-                                />
-                            </div>
-                        </Carousel.Item>
-                    ))}
-                </Carousel>
+                <div className="bg-white py-8">
```
bretthayes, line 353: Can you change `bg-white` to `bg-gradient-onahau-fog`?

website/src/pages/jobs.tsx
```diff
@@ -304,22 +350,11 @@ const CareersPage = props => {
...
+                    <ContentSection>
+                        <QuoteCarousel items={glassdoorReviewItems} autoAdvance={true} />
+                    </ContentSection>
+                </div>
 
                 <div className="bg-white contact-us">
```
bretthayes, line 359: Can you add `py-8` and remove the one-off padding in the stylesheet? I think the section doesn't need that much padding and it kind of ends gracefully similarly to how the hero starts off.

bretthayes (changes requested):
Awesome! This looks great! Definitely an improvement. Made some comments on a few things, then we should be good!

website/src/pages/jobs.tsx
st0nebraker: Sure thing!

website/src/pages/jobs.tsx
st0nebraker: Ah, yea I meant to adjust the bullet point quote. Done ✅

bretthayes (dismissed):
It's beautiful!! Thanks for jumping on this so quickly!! 🎉

st0nebraker:
@elzannewentzel can QA give a final pass on this since there was no specific spec in #5056 ?

bretthayes (dismissed):

courier-new:
This looks awesome!! Thanks for addressing the accessibility concerns! 🙌 I only have two minor, non-accessibility-related questions/concerns:
1. It strikes me that all the other carousels on this screen have a smooth, animated transition between items, but this one just jumps from quote to quote. Is that difference intentional? It feels a bit weird.
2. For this quote:
    Screen Shot 2022-04-11 at 5 09 46 PM
    The big heading is also the first line of the longer quote. Would it make sense to use a different part of the quote for the big heading? None of the others have that duplication. Separately, if the original quote used a lower case "i", should we fix it here? Like: "For the first time ever.. [I] have no complaints about my place of employment."

st0nebraker:
Thanks, @courier-new ! Good eye catching the `i` in that quote, and I took a new "soundbite" for the header of that quote slide.

The `QuoteCarousel` component I'm using here in favor of re-usability is lacking that smooth animated sliding transition. It looks like all of our Custom Carousels' are missing that animated transition. @bretthayes +/or @katjuell , do you think we should open another story to give all our custom carousels an animated transition? I think it'd take a bit extra work that's worth capturing in a separate story for our new sprint format.

bretthayes:
@st0nebraker yes, I agree! We can tackle the carousel animation enhancement in a new story and this page will inherit that when it's merged.

bretthayes (approved):

PR #5291, opened 2022-04-25
Fix HubSpot form loading 

### LINKED ISSUE #5290: HubSpot form not loading
Forms won't load on client-side navigation.

katjuell:
Closes #5290 by updating the HubSpot hook to ensure that forms load on client-side navigation.

## To Test
- Head to the `/code-insights` page and check that the forms load.
- Head to another page (`/pricing` for example)
- Navigate back to the code insights page and ensure that the forms are loaded.


website/src/css/_colors.scss
```diff
@@ -63,12 +63,10 @@ $dodger-blue: #2c7ef6;
 $theme-colors: (
```
katjuell, line 63: Prettier made some changes to this file.

website/src/hooks/hubSpot.ts
```diff
@@ -85,16 +85,10 @@ interface HookProps {
...
-
-    if (!script) {
-        const scriptElement = document.createElement('script')
-        scriptElement.src = hubSpotScript
-        document.head.append(scriptElement)
-        return scriptElement
-    }
-
-    return script
+    const scriptElement = document.createElement('script')
```
katjuell, line 88: @bretthayes I don't **think** we need to do this for ChiliPiper, but lmk if you disagree.

website/src/hooks/hubSpot.ts
```diff
@@ -85,16 +85,10 @@ interface HookProps {
 
 const loadHubSpotScript = (): HTMLScriptElement | Element => {
     const hubSpotScript = '//js.hsforms.net/forms/v2.js'
-    const script = document.querySelector(`script[src="${hubSpotScript}"]`)
-
-    if (!script) {
```
bretthayes, line 90: Removing this causes the `v2.js` `hubSpotScript` to be appended to the head multiple times when navigating back and forth between pages and also if a page has multiple forms.

For pages that include multiple forms of the same formId, we need to render the form in different places but not append the hubSpot script again.

website/src/hooks/hubSpot.ts
bretthayes: Yeah, this goes along with the same process for HubSpot since we don't want to append the script multiple times if it's already present on the page.

bretthayes (changes requested):
We may need to find another treatment by triggering a re-render of the forms if that's the bug, but not append the script multiple times. It looks to be working but I'm unsure how stable this is because ChiliPiper used to hang if the script was loaded twice. Maybe that was just for the ChiliPiper script and not for the HubSpot script? But either way I don't think we should be appending scripts more than once since that may cause inconsistent results.

website/src/hooks/hubSpot.ts
katjuell: 100%

katjuell:
Thanks @bretthayes for troubleshooting with me!

elzannewentzel:
@katjuell Thank you so much for looking into this! Fabiana reported that there. are other pages that are also showing this behaviour. I can't replicate it on these pages, but can you please test it as well and see if we need to do anything, please?

These are the other pages:

- The demo request page doesn't always load the form
- The Code Insights page doesn't always show the demo button
- The Code Search page "Deploy locally" CTA doesn't work 
- The Batch Changes demo page doesn't always load the form

katjuell:
Hi @elzannewentzel — was she reporting that behavior after we deployed this fix? The fix went out EOD yesterday EST. The deploy locally CTA fix went out yesterday as well, in #5308

elzannewentzel:
@katjuell It sounds like it's after the deploy yesterday. Noted on "Deploy locally" fix, they are aware of that.

PR #5305, opened 2022-04-28
Add og url and image secure url for meta

bretthayes:
Adds `og:url` and `og:image:secure_url` for our meta as per the [spec](https://developers.facebook.com/docs/sharing/webmasters#images).


katjuell (approved):
lgtm!

PR #5332, opened 2022-05-05
Wrap labels

rebeccadee:
@bretthayes I noticed the labels weren't being auto-applied from this issue template and I think this might be the problem! 🙏

### DIFF
--- a/.github/ISSUE_TEMPLATES/blog_post.md
+++ b/.github/ISSUE_TEMPLATES/blog_post.md
@@ -2,7 +2,7 @@
 name: Blog post proposal
 about: Use this template for proposing an idea for a new blog post.
 title: 'Blog post working title'
-labels: about-blog, blog-post-proposals
+labels: "about-blog", "blog-post-proposals"
 ---
 
 ## Proposal


bretthayes (changes requested):
I'm not seeing the issue template option come up when I click on the create "New Issue" button. I think it may be because the directory is plural. I just checked out [the docs](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/manually-creating-a-single-issue-template-for-your-repository) and it should be `ISSUE_TEMPLATE`. If we rename it, that should fix it.

The double quotes will actually break it because the yaml front matter will look for labels with quotes: (ie: `"about-blog"` which doesn't exist vs. `about-blog`).

I just noticed in Settings that we're able to create issue templates with GitHub's issue template UI builder. Are you able to see [this](https://github.com/sourcegraph/about/issues/templates/edit)?

rebeccadee:
@bretthayes I was never able to find that builder via our GitHub UI 🤦‍♀️ Thank you for sharing, I will try it! 

My "fix" with the quotation marks was based on [this comment](https://github.community/t/apply-labels-automatically-at-issue-creation/1218/17), I thought maybe the hyphens were causing problems but let me see what I can do with the builder. Stay tuned!

rebeccadee:
Closing in favour of https://github.com/sourcegraph/about/pull/5333

bretthayes:
> @bretthayes I was never able to find that builder via our GitHub UI 🤦‍♀️ Thank you for sharing, I will try it!

@rebeccadee you're welcome! ☺️ 

> My "fix" with the quotation marks was based on [this comment](https://github.community/t/apply-labels-automatically-at-issue-creation/1218/17), I thought maybe the hyphens were causing problems but let me see what I can do with the builder. Stay tuned!

Ah yes! For this case, since there are colons within the labels, if they weren't surrounded by quotes then it would break the yaml since it would be a syntax error. Slight edge case here 🤓

PR #5363, opened 2022-05-19
Test commit

st0nebraker:
### Test

1. Ensure linting passes.
2. Ensure prettier has standardized the proposed changes.
3. Ensure dev and production builds work.

### DIFF
--- a/src/components/Blog/BlogListItem.tsx
+++ b/src/components/Blog/BlogListItem.tsx
@@ -32,6 +32,33 @@ export const BlogListItem: FunctionComponent<PostIndexItem> = ({
                     <time dateTime={frontmatter.publishDate}>{formatDate(frontmatter.publishDate)}</time>
                 </p>
             )}
+            {frontmatter.authors?.length && (
+                <p className="text-align-center text-secondary mb-0">
+                    {frontmatter.authors.map((a, index) => (
+                            <span key={a.name}>
+                                {a.url ? (
+                                    a.url.includes('http') ? (
+                                        <a href={a.url} target="_blank" rel="nofollow noreferrer">
+                                            {a.name}
+                                        </a>
+                                    ) : (
+                                        <Link href={a.url}>{a.name}</Link>
+                                    )
+                                ) : (
+                                    a.name
+                                )}
+                                {index === frontmatter.authors!.length - 1 ? ' ' : ', '}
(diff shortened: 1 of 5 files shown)


zlonko:
@st0nebraker I pushed a quick fix one of the blogpost markdown files, which was causing one of the authors not to render correctly: 5cc031e

zlonko:
@st0nebraker While checking, I came across two blogposts that were rendering incorrectly when opened. This appears to be due to `style: short-inline-title`, so I removed those and moved the `canonical` link into the text of one of the posts. This looks like it's from an old pattern that we can trim this outside of this issue. Just putting a fix in here so we can get this bug fix out. a89e642

zlonko (dismissed):

src/components/Blog/BlogListItem.tsx
```diff
@@ -26,10 +26,31 @@ export const BlogListItem: FunctionComponent<PostIndexItem> = ({
...
+                            )}
+                            {/* eslint-disable-next-line @typescript-eslint/no-non-null-assertion */}
+                            {index === frontmatter.authors!.length - 1 ? ' ' : ', '}
+                        </span>
+                    ))}
+
+                    {frontmatter.publishDate && (
+                        <p className="text-align-center text-secondary mb-0">
+                            <time dateTime={frontmatter.publishDate}>{formatDate(frontmatter.publishDate)}</time>
+                        </p>
```
bretthayes, line 52: This is actually invalid HTML syntax. You can't nest p tags inside p tags.

Error in the console:
```
Warning: validateDOMNesting(...): <p> cannot appear as a descendant of <p>.
```

I would suggest moving this down as a sibling of the other p tag and then wrapping the whole block in a fragment `<></>`

bretthayes (changes requested):
Just a small syntax change, but other than that looks good!

src/components/Blog/BlogListItem.tsx
st0nebraker: Good catch! I had this for `PodcastListItem` & `PostcastPost` but forgot to replicate it here. Fixed in [62cdb04](https://github.com/sourcegraph/about/pull/5363/commits/62cdb04cc6397981bdf025958908acf30821064b)

bretthayes (approved):

PR #5410, opened 2022-05-31
Update 3.40 to show it is safe for GitLab users

iskyOS:
Patch 3.40.1 has been released and is now safe for GitLab code host users. Updating the message box to reflect this.


abschu (approved):
Looks good - thanks @iskyOS

PR #5395, opened 2022-05-25
Navigation updates, Code Security name change

zlonko:
This closes #5266. It updates navigation in the header to expose use cases and reflect the renamed Code Security page. Links are also updated to support Code Security and the removal of the Product Specialist page, a.k.a. "Become One".

Specs are located in this [doc](https://docs.google.com/document/d/1WWlrDoND2-YC1n1P81JDhFW3fJQ2tKP4THqt3MLiuuQ/edit#).

### Changelog
- Restructured the desktop and mobile dropdowns for "Customers" in `Header`
  - Renamed dropdown to "Use Cases"
  - Removed links to `/use-cases` and `/product-specialist`
  - Added links for the five use case pages
- Flipped order of "Case Studies" and "Use Cases" in `Footer`
- Deleted `product-specialist.tsx` per specs
  - Added redirect from `/contact/product-specialist` to `/demo`
  - Changed in-repo links to this page to `demo`
- Updated "Fixing Vulnerabilities" to "Code Security"
  - Renamed `vulnerabilities.tsx` to `code-security.tsx`
  - Updated [meta](https://docs.google.com/spreadsheets/d/1_mZxd_RLysQ0VMUzHK_gQZ8Jak-o7cTyaWS8tqYLZvg/edit#gid=149103332) on `code-security.tsx`
  - Changed in-repo links and added redirect from `/use-cases/vulnerabilities` to `/use-cases/code-security`
  - Updated anchor link on `/use-cases`

### Bug Fixes
- The mobile navs on Batch Changes and Code Reuse were invisible, so those have been fixed
- Code Insights was added back into the nav on this issue: https://github.com/sourcegraph/about/issues/5391

### Notes
- Re: Linking the Use Case dropdown header to `/use-cases`
  - react-bootstrap doesn't appear to support adding an href on NavDropdown: [docs](https://react-bootstrap.github.io/components/navs/#nav-dropdown-props)
  - This also seems to pose a usability issue: a user could click the header wanting to expand the menu, but end up following the link (or vice versa). Also, a user may not expect this behavior, because neither the Product nor Resource dropdowns will repeat this pattern. Perhaps it is more declarative to have another nested item for "All Use Cases"? For example:
 
     Screen Shot 2022-05-25 at 5 03 37 PM

### Test
1. Ensure linting passes.
2. Ensure prettier has standardized the proposed changes.
3. Ensure dev and production builds work.
4. Please check that the nav links under the "Use Case" dropdown work as expected on mobile and desktop
[...]

PR #5420, opened 2022-06-02
Follow up for nav updates

bretthayes:
Follow up for #5395, removing "PDF" from the Sourcegraph Overview copy

### Test

2. The copy "pdf" is removed from the "Sourcegraph Overview" link


st0nebraker (approved):
Groovy

PR #5435, opened 2022-06-06
Code Security Copy Update

### LINKED ISSUE #5405: Update Fixing vulns page to Code security
The Find & fix vulnerabilities use case is being renamed to Code security. We already updated the main nav and Use case landing page in #5266. 
The changes required for this are explained in this [example](https://app.markup.io/markup/856a513d-8fff-404f-8a1c-f48373ca3ce1)

Asana: https://app.asana.com/0/1200095757619647/1202325344671144/f

zlonko:
This closes #5405. It updates the copy on the Code Security Use Case page.

### Changelog

- Changes made to copy on `/use-cases/code-security`
- Discovered inconsistencies in the spacing, buttons, and responsiveness of the "Try Sourcegraph" sections on the Use Case pages

### Notes

- No spec, but the copy changes are captured in this Markup: [link](https://app.markup.io/markup/856a513d-8fff-404f-8a1c-f48373ca3ce1)

### Test

2. Navigate to `/use-cases/code-security`
3. Please ensure copy is transcribed

### DIFF
--- a/src/pages/use-cases/code-security.tsx
+++ b/src/pages/use-cases/code-security.tsx
@@ -227,7 +227,7 @@ const UseCasePage: FunctionComponent = () => (
                                     Find, fix, and track vulnerable code across your entire codebase in minutes, not
                                     days
                                 </div>
-                                <div className="flex-column flex-md-row d-md-flex align-items-center">
+                                <div className="flex-column flex-md-row d-md-flex text-center">
                                     <div className="mb-3 mb-md-0">
                                         <Link href="/demo" passHref={true}>
                                             {/* eslint-disable-next-line jsx-a11y/anchor-is-valid */}
@@ -352,7 +352,7 @@ const UseCasePage: FunctionComponent = () => (
                         <Link href="/demo" passHref={true}>
                             {/* eslint-disable-next-line jsx-a11y/anchor-is-valid */}
                             <a
-                                className="btn btn-primary max-w-200 w-100"
+                                className="btn btn-primary max-w-350 w-100"
                                 title="Request a Demo."
                                 data-button-style={buttonStyle.primary}
                                 data-button-location={buttonLocation.bodyDemo}
--- a/src/pages/use-cases/code-security.tsx
+++ b/src/pages/use-cases/code-security.tsx
@@ -227,12 +227,12 @@ const UseCasePage: FunctionComponent = () => (
                                     Find, fix, and track vulnerable code across your entire codebase in minutes, not
                                     days
                                 </div>
-                                <div className="max-w-400 flex-column flex-md-row d-md-flex align-items-center">
+                                <div className="flex-column flex-md-row d-md-flex align-items-center">
                                     <div className="mb-3 mb-md-0">
                                         <Link href="/demo" passHref={true}>
                                             {/* eslint-disable-next-line jsx-a11y/anchor-is-valid */}
                                             <a
-                                                className="btn btn-primary w-100"
+                                                className="btn btn-primary w-100 max-w-350"
                                                 title="Request a Demo."
                                                 data-button-style={buttonStyle.primary}
                                                 data-button-location={buttonLocation.hero}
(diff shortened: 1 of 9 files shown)


src/pages/use-cases/code-security.tsx
```diff
@@ -343,15 +341,18 @@ const UseCasePage: FunctionComponent = () => (
...
+                            security.
+                        </p>
                     </div>
-                    <div className="d-flex flex-column">
+                    <div className="text-center col-12 px-0">
                         <Link href="/demo" passHref={true}>
                             {/* eslint-disable-next-line jsx-a11y/anchor-is-valid */}
                             <a
-                                className="btn btn-primary"
+                                className="btn btn-primary min-w-200"
```
st0nebraker, line 355: Maybe we should loop in Tracey here to get her opinion. I think all these buttons should take up screen width on mobile, especially since the same button in the hero does this. When we have a standardized style, maybe we should make a re-usable component for this section used on all the Use Cases to ensure they stay standardized. It would help us require tracking too, if we require it as a prop + if we do that sort of approach on our commonly used CTA's.
(Although a new `max-w` class on the hero buttons is giving us a weird UI bug there too heh)
image
image

st0nebraker (changes requested):
Copy changes look good 💯  - Just an open question on the shared CTA spacing

src/pages/use-cases/code-security.tsx
bretthayes: I don't think we should use min-width almost anywhere throughout the site unless there's a really good use for it. When we use it, we should override it at the next below or above breakpoint. But in most cases we should use max-width which is more efficient for responsive flexibility and reduces the amount of code for breakpoint specific width overrides.

If we want to span the buttons a little bit bigger on mobile, I think we should only go up to a max of 350px wide, otherwise it's just too big IMO.

@zlonko this should give you what you're looking for to match @st0nebraker's first screenshot, or `max-w-350` for the latter.

```suggestion
                                className="btn btn-primary max-w-200 w-100"
```

Either way, I agree that we should have some consistency throughout the site but let's scope that in another ticket to avoid scope creep here. Same thing regarding a common/shared button component. We could potentially create a Button component with a `variation` prop to consolidate a mapping of a few different CTAs we have (ie: `demo`, `try`, etc.) but I wouldn't want to overcomplicate it since not all button and link CTAs are the same.

bretthayes (changes requested):
Copy changes look good. Just need to swap out the min width classes for max width classes.

src/pages/use-cases/code-security.tsx
st0nebraker: The `Request demo` buttons spacing should still match each other while we wait for any brand design changes. That's what the spec has outlined [here](https://www.figma.com/file/kQTq5eWvjbMcMmsvZ8wvyF/Code-reuse?node-id=106%3A53). If the buttons width has a `max-w-350` I think it should be centered to the screen. I agree if it's scope creep a follow up ticket can be made for this. Thoughts?

zlonko: @st0nebraker It looks like I misunderstood, although I am encountering some trouble with the link due to a Figma outage. To clarify: 
- Both Request Demo buttons should have a max-width of 350
- Both Request Demo buttons should be centered to the screen on mobile

Is the screenshot what you are looking for on mobile? Please let me know what I may be missing. I can then apply these changes to the other pages. 

Screen Shot 2022-06-07 at 11 48 58 AM

Screen Shot 2022-06-07 at 11 52 55 AM

Screen Shot 2022-06-07 at 11 44 25 AM

st0nebraker: Yes that screenshot matches the spec! 👍🏻

zlonko: Great! Your suggestion is a great idea. I fixed the width and alignment of the Request Demo buttons on the other four use cases pages as well.

st0nebraker: Thank you! Looks good 😄

st0nebraker (approved):

zlonko:
@elzannewentzel At this point, I think we can share these updates with our stakeholders. Here is the [Netlify link](https://deploy-preview-5435--sourcegraph.netlify.app/use-cases/code-security)!

zlonko:
Per Fabiana's feedback, the section title and table of contents are updated on the Use Case index. 9bfe7c7

bretthayes (approved):

Issue #5409, opened 2022-05-31
Hubspot master forms


elzannewentzel:
We have identified improvements that can be made to how we set up Hubspot forms.

https://docs.google.com/document/d/1qa393uAW_tjwNtH9jHf_d__PCJgmbsxNoxi9cO4XOcE/edit

PR #5442, opened 2022-06-07
Webinar Recording pg: Code Insights

### LINKED ISSUE #5412: Gated page: Code insights webinar recording
We are making the recording of the Code Insights webinar available to users. 
Reuse the existing [webinar landing page](https://about.sourcegraph.com/webinars/code-insights-turning-your-metrics-into-action) #5317 
After filling in the form, the user should land on the page with the video.
The recording page should be accessible without filling in the form.

🔗[Figma](https://www.figma.com/file/GASbA761UFtEe9LEE4j43a/Code-Insights-webinar?node-id=79%3A3021) 
The speakers are placeholders but they should be the same component from the original gated pages
[🔗Meta ](https://docs.google.com/spreadsheets/d/1SF6taZ-Jdw6KiV-6JLz8mzOjJTh0qTcAgKh1Gcyo0Kk/edit#gid=1942549407)
📹[Video ](https://drive.google.com/file/d/143csAoQZ2eQK1DaCNOekeV0MdqFX61tb/view)
YouTube video: https://youtu.be/dXKvetMozB0

To create consistency, can we also update the header image on  [this page](https://about.sourcegraph.com/webinars/code-insights-turning-your-metrics-into-action) to the orange background (like [Nutanix](https://about.sourcegraph.com/webinars/preparing-for-the-next-log4j))

Asana: https://app.asana.com/0/1200095757619647/1202325344671170/f

st0nebraker:
Closes #5412 by adding a `/watch-now` recording page for the Code Insights webinar

### Changelog
- Adds `watch-now` sub-path to the Code Insights webinar page with the recording available
- Webinar gated page submission now pushes to this page
- Opens up `GatedResourceLayout` to account for different body for Recording pages
- Updates `GatedResourceLayout` hero bg to new requirements

### Test
2. Nav to `/webinars/code-insights-turning-your-metrics-into-action` + ensure form submission brings the user to the recording page
3. Nav to `/webinars/code-insights-turning-your-metrics-into-action/watch-now` and ensure page matches spec


zlonko (changes requested):
This looks great. Everything works as expected, and the previous pages were intact when I checked. Just a few minor notes!

src/pages/webinars/code-insights-turning-your-metrics-into-action/watch-now.tsx
```diff
@@ -0,0 +1,62 @@
...
+            learnMoreCTA={
+                <ContentSection className="d-flex flex-column align-items-center py-lg-8 py-7">
+                    <h1 className="font-weight-bold text-center">Want to learn more about Code Insights?</h1>
+                    <Link href="/contact/request-code-insights-demo" passHref={true}>
+                        {/* eslint-disable-next-line jsx-a11y/anchor-is-valid */}
+                        <a
+                            className="btn btn-primary mt-4 col-12 col-md-3 col-xl-2"
+                            title="Request a Demo"
+                            data-button-style={buttonStyle.primary}
+                            data-button-location={buttonLocation.trySourcegraph}
```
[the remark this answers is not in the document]
st0nebraker: I agree! Committed the suggestion, thanks Tim 😄

src/pages/webinars/code-insights-turning-your-metrics-into-action/watch-now.tsx
```diff
@@ -0,0 +1,62 @@
...
+        <GatedResourceLayout
+            title={
+                <>
+                    Sourcegraph Code Insights:
+                    <br /> Turning Metrics into Action
+                </>
+            }
+            subtitle="How to track what really matters to you and your team"
+            speakers={speakers}
+            videoSrc="https://storage.googleapis.com/sourcegraph-assets/webinars/code-insights.mp4"
```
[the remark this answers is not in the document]
st0nebraker: Yep! You beat me to the punch commenting before I had a moment to update. See updates in [6feef93](https://github.com/sourcegraph/about/pull/5442/commits/6feef93c42d1e7765c600372887a1629b081bedb)

src/pages/webinars/code-insights-turning-your-metrics-into-action/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+                        <br /> Turning Metrics into Action
+                    </>
+                }
+                subtitle="How to track what really matters to you and your team"
+                speakers={speakers}
+                formLabel="Watch the on-demand webinar"
+                form={{
+                    formId: '66361163-5e08-4be3-8ab0-6590b70df69e',
+                    onFormSubmitted: () =>
+                        router.push('/webinars/code-insights-turning-your-metrics-into-action/watch-now'),
```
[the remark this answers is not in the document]
st0nebraker: I imagine #5409 will take care of this. But I went ahead and updated the styling on HS to be unified here while we wait for #5409 to complete 👍🏻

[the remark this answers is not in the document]
bretthayes: Yes for sure! #5409 should consolidate our forms to be consistent with the renamed form input labels and the submit button text.

src/pages/webinars/code-insights-turning-your-metrics-into-action/watch-now.tsx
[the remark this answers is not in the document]
bretthayes: Beauuuuty! ✨

src/components/GatedResourceLayout/index.tsx
```diff
@@ -58,25 +62,21 @@ export const GatedResourceLayout: FunctionComponent<Props> = ({
     const isWebinarPg = useRouter().pathname.split('/').slice(1)[0] === 'webinars'
     const isGuidePg = useRouter().pathname.split('/').slice(1)[0] === 'guides'
 
-    const hubSpotConfig: HubSpotForm = {
+    const hubSpotConfig: HubSpotForm | null = {
         portalId: '2762526',
-        formId: form.formId,
+        formId: form?.formId ?? '',
```
bretthayes, line 67: The `formId` is always required if you're using the HubSpot hook so I don't think we should allow it to ever be optional. I see why you're allowing this now that I've looked at the rest of the code - more on this in further comments.

```suggestion
        formId: form.formId,
```

src/pages/webinars/code-insights-turning-your-metrics-into-action/watch-now.tsx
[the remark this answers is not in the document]
zlonko: Nice! This looks good 👍

zlonko (approved):
Thanks for taking a look at those notes, Becca. This looks good to go!

src/pages/webinars/code-insights-turning-your-metrics-into-action/watch-now.tsx
```diff
@@ -0,0 +1,62 @@
...
+                            Request a demo
+                        </a>
+                    </Link>
+                </ContentSection>
+            }
+        />
+    </Layout>
+)
+
+export default Recording
```
bretthayes, line 62: Instead of creating a duplicate page, we can avoid some duplicate content here and I think we can optimize the `GatedResourceLayout` component to use a `?watch-now` query parameter. Then you can use that to conditionally show the `videoSrc` and swap that content. You can also use the param condition to not run the `useHubSpot` hook which will remove the need for an optional `formId`. This page with or without the `watch-now` parameter can be measured against each other for analytics.

src/components/GatedResourceLayout/index.tsx
```diff
@@ -132,20 +132,43 @@ export const GatedResourceLayout: FunctionComponent<Props> = ({
...
+                    </ContentSection>
 
-                {children}
-            </section>
+                    {children}
+                </section>
+            )}
+
+            {/* ---- RECORDING BODY VARIATION ---- */}
+            {videoSrc && learnMoreCTA && (
```
bretthayes, line 154: I think `videoSrc` would be sufficient and you don't necessarily need both conditions. But, with my recent suggestion of using a query param, you can use that to conditionally show this content.

src/data/tracking.ts
```diff
@@ -1,3 +1,5 @@
+/** Documentation at: https://docs.google.com/document/d/1Q1JMzyRmwNKk65KxxtT9EOnxoSf2dHEGkUAPbJkKzI4/edit */
```
bretthayes, line 1: Sweet! Thanks for adding this 🎉

src/hooks/hubSpot.ts
```diff
@@ -168,6 +168,10 @@ export const useHubSpot = ({
     onFormSubmitted,
 }: HookProps): void => {
     useEffect(() => {
+        /** Some pg's conditionally require HS forms, exit HS creation when formId is falsey */
+        if (!formId) {
+            return
+        }
```
bretthayes, line 174:
```suggestion
```

bretthayes (changes requested):

src/pages/webinars/code-insights-turning-your-metrics-into-action/watch-now.tsx
st0nebraker: I see what you mean. I like the query param path better, less redundant, slightly different slug but better practice. Although, the hook will still run in the background when the form portion is hidden because we can't conditionally call hooks as part of the [React hook rules](https://reactjs.org/docs/hooks-rules.html). I tried disabling the linter to do this, but then it won't compile. Do you have another idea to exit the hook when it's not needed? Or think it's fine to leave running?

bretthayes: Ah, I forgot about that hook rule for conditionals. In that case, I don't think we should leave the hook running since it would load for folks directly navigating to the resource.

I was going to suggest using `useState` and passing that to `onFormSubmitted` in the `GatedResourceLayout` so you could do something like `hasWatchNowQuery || formSubmitted` instead of `router.push`, but then that would prevent us from tracking those analytics between the pages for users submitting forms. But we'd still run into the hook being fired for users navigating directly to it.

My apologies, I didn't think this through thoroughly. I think we may have to revert to your previous solution but with a few optimizations since the URL is cleaner with `page-name/watch-now`. I'd suggest sharing the duplicated data between the two pages so you can add them into each `GatedResourceLayout` for each page conveniently. You could export the data from one page, and import in the `watch-now` page.

zlonko (dismissed):

st0nebraker:
This is ready for stakeholder review, @elzannewentzel ! Staging link to the video is [here](https://deploy-preview-5442--sourcegraph.netlify.app/webinars/code-insights-turning-your-metrics-into-action?watch-now), webinar landing page is [here](https://deploy-preview-5442--sourcegraph.netlify.app/webinars/code-insights-turning-your-metrics-into-action) (form submission taking the user to the video)

PR #5475, opened 2022-06-13
Collocated home redirect trailing slash fix

bretthayes:
Minor fix to the `/home` redirect which should point to `/` when there is and isn't a trailing slash. This forces the redirect no matter what to the root `/`.

Note, this implementation may change in the future with more added support for collocated/nested layouts in the next version of Next.js with their [layouts RFC](https://nextjs.org/blog/layouts-rfc).

### Test
1. Ensure `/home/` or `/home` or `/home/_hero` redirects to `/`

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -142,4 +142,5 @@
 [[redirects]]
   from = "/home"
   to = "/"
+  force = true
 # ========== END INDIVIDUAL REDIRECTS ==========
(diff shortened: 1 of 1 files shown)


zlonko (approved):
Successfully tested the different redirects. This looks good 👍

st0nebraker (approved):
▶️ ▶️ 🏠  looks good!

PR #5476, opened 2022-06-13
Icon Styling

### LINKED ISSUE #5432: Update icon styling
Update the icon library we use on the platform (maybe) 
Update the styling of all icons on the platform:
- color
- background
Example of new styling: https://www.figma.com/file/o1QRtdQI0ozKq0n7ATrKlx/Brand-DLS?node-id=7341%3A49748

zlonko:
This closes #5432. It centralizes our icons in a flexible component and updates styling across the About site.

### Changelog

- Moves icons into `Icon` component
  - Accepts icon name, size, onClick, additional classes, etc.
  - Implements [new icon styling](https://www.figma.com/file/o1QRtdQI0ozKq0n7ATrKlx/Marketing-DLS?node-id=4422%3A86874) on `variant="boxed"`
  - Allows for plain icon when `variant` prop is omitted
- Uses the [Material UI icon library](https://www.npmjs.com/package/@mui/icons-material) identified in conversation with Tracey

New icon style:
Screen Shot 2022-06-16 at 9 01 55 AM

### Notes
- Spotify and Twitch logos are the only remaining mdi-react icons. They were [not available in the library](https://mui.com/material-ui/material-icons/?theme=Sharp). Are we open to replacing these with SVGs?
- I ran into an issue rendering blogposts with `BlockquoteWithBorder` due to the inclusion of the `Icon` component. TIL when a component is passed into another component, and then used as in MDX, Next does not use the Babel-aliased path when reading the component. So, `Icon` needs to be imported with a relative path in `BlockquoteWithBorder` for MDX to find it.
- Still working on optimizing the bundle size (see [comments](https://github.com/sourcegraph/about/issues/5432#issuecomment-1155667416) in [issue](https://github.com/sourcegraph/about/issues/5432#issuecomment-1156593359)). From the [MUI Docs](https://mui.com/material-ui/guides/minimizing-bundle-size/) it seems that tree-shaking should be happening automatically, and the site seems load quickly despite the larger bundle size measured on the production build. I am looking to learn more about this, so I appreciate any feedback!

### Test

2. Check that icons are rendered according to the new style on the following pages
  - `/use-cases` and Use Case pages
  - `/code-insights`
  - `/case-studies/nutanix-fixed-log4j-with-sourcegraph`
3. Please check that there are no breaking changes with plain icons
  - Arrows in carousels and blockquotes appear and function as expected
  - Company logos (GitHub, Twitter, etc.) render at consistent sizes


src/components/Icon.tsx
```diff
@@ -4,10 +4,11 @@ import * as icons from '@material-ui/icons'
...
+    size: number
 }
 
-export const Icon: FunctionComponent<Icon> = icon => {
-    const IconName = icons['SearchSharp']
+export const Icon: FunctionComponent<Icon> = ({ icon, size }) => {
+    const IconName = icons[icon]
 
-    return <IconName className="mb-4 bg-violet-mist text-vivid-violet rounded p-2" style={{ fontSize: 48 }} />
+    return <IconName className="mb-4 bg-violet-mist text-vivid-violet rounded p-2" style={{ fontSize: size }} />
```
zlonko, line 13: @bretthayes Following up on your [comment in the issue](https://github.com/sourcegraph/about/issues/5432#issuecomment-1155413136), here is the proposed `Icon` component.

bretthayes: Cool! I'd suggest using `name` as the prop instead of `icon` since we already know it's an icon from the name of the component 🤓

src/components/Icon.tsx
```diff
@@ -0,0 +1,27 @@
...
+    name: keyof typeof icons
+    size: number
+    className?: string | undefined
+    color?: string
+    key?: string
+    onClick?: (event: SyntheticEvent) => void
+    variant?: 'boxed'
+}
+
+export const Icon: FunctionComponent<Icon> = ({ name, size, className = '', color, key, onClick, variant }) => {
```
st0nebraker, line 15: I'm seeing this console error on some pages like `/code-insights`, we should probably give key a default value. I was bringing in this [random id creator function](https://github.com/sourcegraph/about/blob/improve-syntax-highlighting/src/util/createRandomId.ts) from the Learn site in my working PR, maybe bring that into here?
image

src/components/Icon.tsx
```diff
@@ -0,0 +1,27 @@
...
+    const IconName = icons[name]
+
+    return (
+        <IconName
+            className={`${className} ${variant === 'boxed' ? 'bg-violet-mist text-vivid-violet rounded p-2' : ''}`}
+            style={{ fontSize: size }}
```
st0nebraker, line 21: I wonder if we can put in a default size too so it won't always be needed for the majority of icons put in the body. Seems like 48 or 21? Was that talked about with Tracey? I think it'd help reinforce our standardizations.

src/components/Icon.tsx
```diff
@@ -0,0 +1,27 @@
...
+    onClick?: (event: SyntheticEvent) => void
+    variant?: 'boxed'
+}
+
+export const Icon: FunctionComponent<Icon> = ({ name, size, className = '', color, key, onClick, variant }) => {
+    const IconName = icons[name]
+
+    return (
+        <IconName
+            className={`${className} ${variant === 'boxed' ? 'bg-violet-mist text-vivid-violet rounded p-2' : ''}`}
```
st0nebraker, line 20: This is small, but I don't think a ternary is necessary if one result is empty. I'd prefer
```suggestion
className={classNames(className, variant === 'boxed' && 'bg-violet-mist text-vivid-violet rounded p-2')}`
```

or 
```suggestion
className={classNames(className, { 'bg-violet-mist text-vivid-violet rounded p-2': variant === 'boxed' })}
```

st0nebraker (changes requested):
This is sweet, Tim! Just a few comments below

src/components/Icon.tsx
zlonko: Thank you for noting this; it is definitely the better way! In addition to your suggestion, I added an import for `classNames`: 31a0f7d.

src/components/Icon.tsx
zlonko: This is an excellent point, considering the potential for variations! Tracey's Figma specifies the new icon style with a size of 48, so let's go with that as the default. We can flag this as a decision point for her in stakeholder review. 

In 9af1fa6, I added this as the default size and removed instances where `size={48}` is used on the component. 

You have me thinking. Ultimately, we might emulate what we're doing here with `variant` so that `size` can only be chosen from categories determined in this component. For example, the `Icon` component would accept a string corresponding with the uniform sizes of nav arrows, iconography, or social media icons. Because that adds some maintenance, let's first checkout what Tracey has planned. Great to know what you think of this idea though.

src/components/Icon.tsx
zlonko: First of all, what a helpful script. Wow!

It looks like the error is related to the presence of the `key` prop in `Icon.tsx`, because it goes away once the prop is removed from the `FunctionComponent` constant. However, we can still use `createRandomId()` to assign the random key value within the return. For instances where the `Icon` component needs a key value, such as the `map()` on `CustomCarousel` creating the Code Insights error, a key value still needs to be assigned at the function level.

See: 1273f57

src/components/Layout/Footer.tsx
```diff
@@ -1,13 +1,11 @@
...
 
-import GithubIcon from 'mdi-react/GithubIcon'
-import LinkedinIcon from 'mdi-react/LinkedinIcon'
 import SpotifyIcon from 'mdi-react/SpotifyIcon'
 import TwitchIcon from 'mdi-react/TwitchIcon'
-import TwitterIcon from 'mdi-react/TwitterIcon'
-import YouTubeIcon from 'mdi-react/YoutubeIcon'
 import Link from 'next/link'
 
+import { Icon } from '@components'
```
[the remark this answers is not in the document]
zlonko: Right? This foot finally fits in a shoe!

src/components/Icon.tsx
st0nebraker: Exactly! Sounds good , thanks :)

src/components/Icon.tsx
st0nebraker: I like that idea! Takes it a step further & can help us standardization body icons, command icons (for arrows), etc.

st0nebraker (approved):
🚀

zlonko:
Hi, @elzannewentzel, this is ready for stakeholder review. Becca and I are curious what Tracey anticipates as the range of icon sizes on About. We suggest that, rather than using open numbers, we use standard sizes categories ([thread](https://github.com/sourcegraph/about/pull/5476#discussion_r900445937)). This would add a little maintenance, but increase our conformity and ease for updating. It could be a useful tradeoff!

src/components/Layout/Footer.tsx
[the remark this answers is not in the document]
bretthayes: Haha looooove this! 🔥 👟

src/util/createRandomId.ts
```diff
@@ -0,0 +1,7 @@
+const rand4 = (): string => (((1 + Math.random()) * 0x10000) | 0).toString(16).slice(1)
```
bretthayes, line 1: Maybe we should consider using [uuid](https://www.npmjs.com/package/uuid)

bretthayes (changes requested):
A few suggestions, but this looks killer @zlonko!! 😄 🔥 Absolutely stoked about this!

Also, to your comment about mdi-rect, I'm totally cool with using a few SVGs there if there's only Spotify and Twitch logos left. Then we can nuke that package and reduce some bloat.

package.json
```diff
@@ -44,6 +44,10 @@
     }
   },
   "dependencies": {
+    "@emotion/react": "^11.9.3",
+    "@emotion/styled": "^11.9.3",
```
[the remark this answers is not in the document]
zlonko: This is a bit mysterious. It looks like `@emotion` is used in the larger [material-ui repo](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/mui/material-ui+emotion+&patternType=literal), but not in the [icons package](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/mui/material-ui/packages/mui-icons-material+emotion&patternType=literal). This is supported by the [dependency list on npm](https://www.npmjs.com/package/@mui/icons-material). I will remove `@emotion` and see if anything is disrupted.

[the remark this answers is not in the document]
zlonko: I just finished some debugging. It looks like `@mui/icons-material` requires `@mui/material`, which uses the `@emotion` packages. The app needs all four to build. This is the error returned when `@mui/material`, and `@emotion/react`, `@emotion/styled` are removed and a build is attempted.

Screen Shot 2022-06-21 at 3 21 58 PM

bretthayes:
@zlonko can you look into why the builds are failing?

zlonko:
@bretthayes I am testing out different fixes, but I haven't found the solution yet. The build is failing because certain imported components are not recognized in the blog architecture. Specifically, the error occurs on components with nested imports (such as `Icon` imported into `BlockquoteWithBorder` --> "BlockquoteWithBorder not recognized, did you forget to import it?"). 

This is similar to the error that @st0nebraker is encountering on the [syntax highlighting ticket](https://github.com/sourcegraph/about/issues/5403) from our group chat in Slack. I posted a solution to our chat that worked at the time, but that is no longer working. I had found some discussion that suggested that the babel `@` alias was not recognized by the blog parsing process, so I changed the import to a local path (`./Icon`) which removed the error. I am unsure what is different now, so I am taking another look.

bretthayes:
Looks like there's some merge conflicts to fix from our Carousel standardization PR. @zlonko 

The parsing issue sounds like there may be something going on with our alias config that we could look into. I can take a look! 👀

zlonko:
Thanks, @bretthayes! I appreciate anything you see. Otherwise, I resolved the merge conflicts you mentioned and I am continuing to look for a solution to the alias issue.

bretthayes:
Closing this due to too many global changes causing too many merge conflicts. We will revisit this with an alternative solution in the near future.

PR #5499, opened 2022-06-21
Fix Shift Left Video

zlonko:
This fixes a bug duplicating one of the videos in the blogpost. This was due to error transcribing the video path.

### Changelog
- Updates markdown with correct video

### Test

2. Please check that the first and second videos are different.

### DIFF
--- a/content/blogposts/2022/shift-left-wtf-does-it-mean.md
+++ b/content/blogposts/2022/shift-left-wtf-does-it-mean.md
@@ -63,7 +63,7 @@ In order to take full advantage of existing libraries, you need to be able to un
 Investing in great documentation helps, but this is a hard change to make if your codebase isn't already well-documented. The best way to learn is by example. A single real usage example can be worth a thousand words of documentation.
 
 <Video 
-  source={{webm: 'blog/shift-left/02-find-references', mp4: 'blog/shift-left/01-code-reuse'}} 
+  source={{webm: 'blog/shift-left/02-find-references', mp4: 'blog/shift-left/02-find-references'}} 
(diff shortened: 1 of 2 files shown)


st0nebraker:
I don't see the backup videos rendering on Safari, do you @zlonko ?

zlonko:
@st0nebraker I see the same. I am not sure why, though, because I was able to view the videos in Safari before shipping the original blogpost. It seems like reversing the order of the mp4 and webm videos gets them to appear in Safari: f99d86c. Do you see the same on staging?

st0nebraker (approved):
> @st0nebraker I see the same. I am not sure why, though, because I was able to view the videos in Safari before shipping the original blogpost. It seems like reversing the order of the mp4 and webm videos gets them to appear in Safari: [f99d86c](https://github.com/sourcegraph/about/commit/f99d86c1487d219ae29ec6229240946bbf662c99). Do you see the same on staging?

That's odd. The rearranged order seems to work on staging! Lets keep an eye on it after this ships. Thanks Tim ⭐

bretthayes:
Just wanted to note the video element's sources are loaded in the order they are declared in. Here's [the spec for reference](https://html.spec.whatwg.org/multipage/media.html#media-element-load-algorithm:~:text=immediately%20resolve%20or%20reject%20those%20promises%20in%20the%20order%20the%20corresponding%20tasks%20were%20queued.). 🤓 Although mp4 is more widely supported than webm, it's weird that one or the other didn't get loaded as a fallback option? 🤔 

cc @zlonko @st0nebraker

PR #5520, opened 2022-06-27
Syntax Highlighting

### LINKED ISSUE #5251: Code block syntax highlighting
Joe provided feedback on the way we style code block syntax highlighting on the blog. 

https://about.sourcegraph.com/blog/introducing-migrator-service/ 
the code block colour seems too bright
we also use thin font for code (much thinner than the regular text, which kinda makes it even harder to read)

The way it's styled on the docs site is easier to read: https://docs.sourcegraph.com/batch_changes/how-tos/creating_a_batch_change

bretthayes:
Closes #5251 and adds syntax highlighting to code blocks in markdown.

### Changelog
- added prism rehype plugin to our markdown parsing lib
- added prism-themes and included the `prism-one-light` theme
- Updated next and mdx-remote
- removed old code element styles
- removed unused font families and fixed our font family declarations
- a few bug fixes in [fb0a31b](https://github.com/sourcegraph/about/pull/5520/commits/fb0a31b38515b9498b317bb1465ff4885d977aad) mentioned in [this comment](https://github.com/sourcegraph/about/pull/5520#issuecomment-1168965377)

### Test

2. Check some post types for syntax highlighting (ie: `/blog/graphql/handling-authentication-and-authorization-in-graphql`)


st0nebraker (approved):
Looks good. I pushed up a commit for the `yacc` language, `go` works since they're similar. I'm happy to take on a follow up ticket for this to audit code blocks missing a language for highlighting. I think we should bump the code font-size down to match our body paragraphs since that was a piece of the feedback on this issue.
image

Feeling silly for getting stuck on `[prism-react-renderer](https://www.npmjs.com/package/prism-react-renderer)`, I like `[rehype-prism](https://www.npmjs.com/package/rehype-prism)`, seems like a simpler solution, even though the package looks way less popular 🤔  Thanks for looking into this @bretthayes !

zlonko (approved):

bretthayes:
Awesome, thanks for pushing up that change @st0nebraker! Doing a follow up audit for all code blocks that are missing languages would be great! These are all of Prism's [supported languages](https://prismjs.com/#supported-languages).

For a little bit of backstory, I actually tried out rehype-highlight with highlight.js as seen in [my first commit](https://github.com/sourcegraph/about/pull/5520/commits/668704ec30c0a54ca6c39d41afac5e217b8b387a#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R56-R63), but there were less supported languages in the core so I opted out and went for Prism. Prism has [239~ supported languages](https://prismjs.com/#supported-languages:~:text=This%20is%20the%20list%20of%20all%20293%20languages) versus [~40 in highlight.js](https://github.com/highlightjs/highlight.js/blob/main/SUPPORTED_LANGUAGES.md).

As for our font-size, the computed font-size is actually the same for code blocks and body text (16px / 1rem). However, it appears slightly different because of the font-family that code blocks use. I just did an audit of our fonts included on the site since we have 4 different fonts with different weights. I just removed some unused font families so that should increase our performance/load times a bit. I included a matching monospace font to compliment our base font so that should resolve the readability issue. 🤓 

Also just added a few bug fixes I noticed in [this commit](https://github.com/sourcegraph/about/pull/5520/commits/fb0a31b38515b9498b317bb1465ff4885d977aad). The code insights page video wasn't aligned and I noticed some video elements on blog posts that weren't within the container width constraint.

zlonko (approved):
Thank you for your attention on these bug fixes, Brett! Good catch 👍

unknwon:
Thank you!

bretthayes:
Just noting here for historical purposes; we gained a 6% overall performance boost by fixing our font families in commit [3cba42d](https://github.com/sourcegraph/about/pull/5520/commits/3cba42ddaf16a28e1e87272c389f3cd1008900e5). Doing a Lighthouse audit just before this PR was merged and after this PR was merged resulted in the following results:

- overall we went from 54% to 60% (6% difference)
- FCP went from 2.3 to 2s
- LCP went from 3.6s to 2.1s, and;
- CLS went from 0.014 to 0.013

image

PR #5526, opened 2022-06-28
Update Criteo logo

### LINKED ISSUE #5512: Update Criteo logo
Stakeholder: Fabiana
What? Update the Criteo logo with the new version on these pages:
https://about.sourcegraph.com/case-studies
https://about.sourcegraph.com/case-studies/criteo-tackles-big-code
https://about.sourcegraph.com/pricing (bottom of the page)
Why? Criteo updated their logo and we want to ensure we are using the latest brand assets.

Image
Image

Asana: https://app.asana.com/0/1200095757619647/1202502794982933/f

st0nebraker:
Closes #5512

### Changelog
- Updates Criteos logo
- Small fix for broken NavBar spacing, smallest screensize before the mobile menu kicks in

**BEFORE:**
image
**NOW**:
image

### Test
2. Ensure Criteos logo is updated with no breaking changes on: `/case-studies`, `/case-studies/criteo-tackles-big-code`, `pricing`
3. Ensure NavBar does not break to 2 lines at any screen size


src/components/CaseStudies/CaseStudyCard.tsx
```diff
@@ -54,7 +54,7 @@ export const CASESTUDIES: CaseStudy[] = [
     },
     {
         name: 'Criteo',
-        logo: '/external-logos/criteo-logo.svg',
+        logo: '/external-logos/criteo-logo.png',
```
zlonko, line 57: The new logo looks sweet! I did not notice any issues rendering it on `pricing`, `case-studies`, or `case-studies/criteo-tackles-big-code`. 

I noticed that this is the only PNG file we are using for a customer logo, while the other logos are SVG files. If we don't have a preference, would you be open to using the SVG from the issue instead? 

One advantage of using the SVG file is that we don't need to worry about resolution on different devices and screens. An SVG is also much smaller than the PNG, which can improve speed (see below). This [short article](https://www.growfox.co.uk/blog/5-reasons-you-should-be-using-svgs-over-pngs) provides a few more reasons as well.

Screen Shot 2022-06-30 at 9 41 26 AM

src/components/Layout/Header/DesktopNav.tsx
```diff
@@ -15,7 +15,7 @@ interface Props {
 
 const DesktopNav: FunctionComponent<Props> = ({ navLinks, hideGetStartedButton }) => (
     <>
-        <Nav className="left-nav me-auto ml-md-2">
+        <Nav className="left-nav me-auto ml-xl-2">
```
zlonko, line 18: I'm really glad you caught this spacing issue on the nav bar. When I tested locally, I did see that the break persists between 1099px-1109px width (below). I am testing on a Macbook.

This is not in scope for the fix in question, but this does make me think about the two CTAs on the nav bar. I have seen chatter in Slack discussing whether "Request a Demo" and "Get Started" are redundant. Today (6/30) we are pushing an update that will redirect the Get Started CTA to `/get-started/self-hosted`, eliminating the decision tree the CTA provides. Perhaps this is a question for our product folks as they consider the second stage of their plan to conceptualize self hosted vs. cloud offerings.

Screen Shot 2022-06-30 at 9 28 15 AM

zlonko (changes requested):
Thanks, @st0nebraker. This rocks! I left a couple of observations regarding the file and navbar below.

src/components/Layout/Header/DesktopNav.tsx
st0nebraker: Hm I can't reproduce that. At 1099px the mobile nav kicks in. @bretthayes are you noticing this bug persist?

I do agree the CTAs are a bit redundant!

https://user-images.githubusercontent.com/59381432/176945526-1a029eb7-2c0e-47bf-83e9-572e7b0f3de3.mov

src/components/CaseStudies/CaseStudyCard.tsx
st0nebraker: Good point! I think it makes sense for all our customer logos to be the file type too. I updated this in [f1a4dd2](https://github.com/sourcegraph/about/pull/5526/commits/f1a4dd2a24c98999c61eb036d1b1f5f43e750d94)

src/components/Layout/Header/DesktopNav.tsx
bretthayes: I'm not seeing the wrapping bug persist, but I do notice the height of the nav changes slightly which causes a bit of a jump. Since this is a logo update PR, I don't want to introduce scope creep here. So looks good to me for the time being. I can clean up some of the styles/classes in my sticky nav PR since that's related to the nav.

bretthayes (approved):

Issue #5529, opened 2022-06-28
Standardize Figures, Alerts, and other Blog Components


zlonko:
Within the blog, there are many instances of figures, alerts, and tables. By using components, we improve the maintenance of our styling standards, while also making it easier for blog contributors to create content.

e.g., instead of having writers use:

```
<figure>
    <img src="..." alt="..." />
    <figcaption>...</figcaption>
</figure>
```

...where alt text can sometimes be dropped, and `figcaptions` can be used improperly, we recommend in the Starter Pack:

```
<Figure src="..." alt="..." caption="..." />
```

PR #5528, opened 2022-06-28
Blog/Mobile-friendly tables and Components

zlonko:
This closes #5454 and #5529. It implements components to abstract common design patterns in the blog. It also makes tables mobile-friendly.

### Changelog
- Adds `Alert`, `Figure` and `TableWrapper` components
- Standardize `figure`/`figcaption` instances with `Figure` component
- Implements the following components
  - `Alert`
  - `Figure`
  - `TableWrapper` 
  - `Video`
  - `YouTube`

### Notes
- `img`s are converted to `Figure`s in https://github.com/sourcegraph/about/pull/5536 to merge into this PR
- Opted for a wrapper component around tables because they are often written in Markdown instead of HTML. This allows for some flexibility for our author's preferences. Bootstrap's `table-responsive` utility did not appear to read within the markdown, so the class `table-overflow` was created.
- Noting that some tables were difficult to read even when made scrollable on mobile.
- One table on `blog/optimizing-a-code-intel-backend` is changed into a series of figures to make the images more accessible.
 
### Test
1. Ensure prettier has standardized the proposed changes.
2. Navigate to `blog/a-programmers-guide-to-find-and-replace` or any other blogpost with tables to check that tables are rendered with a horizontal scroll on mobile
3. Please see that images and videos render as expected on blogposts

Issue #5513, opened 2022-06-27
Blog post starter pack


elzannewentzel:
Stakeholders: Internal users that want to create a new blog post
What?: A blog post starter pack with a template containing all the different options we currently offer 
Why?: We (CPT) are seeing an incline in requests for creating & publishing blogposts. The Marketing team provides [guidance](https://handbook.sourcegraph.com/departments/marketing/blog/) in the handbook on how to do this, but this process can be hard to follow for teammates that are not familiar with markdown or git. 

We have heard from our stakeholders that they would like this process to be easier and from our recent experience with assisting with these requests, we identified the following needs:

1. Create a starter/draft .md template that includes examples of what is included on the handbook page
- front matter example
- text options with markdown 
- image example
- video example
- Hubspot form example
- about the author section
3. Update instructions on the handbook page (we should notify and review changes with the Marketing team) (this can be done as a separate task)
4. Update instructions on the handbook page to encourage contributors to add an `About the author` section to blogposts.
5. Make the template package available somewhere (in the repo ideally) so that we can communicate it internally
As a company, we want to make it so that everyone can code and if we can make it easier for our teammates to create and publish their blog posts, we contribute to our company's mission

- [ ] #5553
- [ ] sourcegraph/handbook#4186
- [ ] #5541

PR #5536, opened 2022-06-29
convert img to figure in blog

zlonko:
This supplements #5528 for #5529. It converts images used in the blog from `img` tags to the `Figure` components. This is to consolidate the treatment of images across the blog.

This pull request intends to merge into https://github.com/sourcegraph/about/pull/5528.

### Changelog
- Converts `img` to `Figure` in Blog
- Trims unneeded and outdated styling around Blog images

### Test
2. Please select a few blogposts and check that images render as expected


content/blogposts/2020/3.18/custom-search-pages.md
```diff
@@ -15,49 +15,44 @@ This is a great resource that we hope will make these communities accessible to
...
-<div>
-<img alt="K8s search page" src="https://sourcegraphstatic.com/blog/3.18/k8s-search-page.png" />
-</div>
+<a target="_blank" href="https://sourcegraph.com/kubernetes">
+  <h2 className="text-center">Kubernetes</h2>
+  <Figure 
+    alt="K8s search page" 
+    src="https://sourcegraphstatic.com/blog/3.18/k8s-search-page.png"
+  />
 </a>
```
st0nebraker, line 24: I'm noticing when there's a link wrapping `<Figure />` the image alt tag isn't picked up by the screen reader b/c of how nested the image is. So on this page, these blocks read "Link: Kubernetes. Link: Go". Maybe we should re-think how we treat links with our custom Figures so alt tags get picked up properly? Or ensure a descriptive title is in each anchor tag?
On this page, I think the image should definitely be described considering what it's showing.
image

st0nebraker (dismissed):
Just an open question on treating links + Figures with a11y concerns.

Thanks for this big change, Tim! 💥

content/blogposts/2020/3.18/custom-search-pages.md
zlonko: Good point, @st0nebraker. One way we can handle this is to use a wrapper component, such as `LinkWrapper` in b0f3707. Otherwise, I think we can require writers to include a `title` on the `<a>` tags in our standards for the blog. The former ensures that links are implemented consistently, but the latter may be more simple.

If we choose to just go with guidance in the template, I can revert this last commit and add a note to the template we're creating in #https://github.com/sourcegraph/about/issues/5513 and the Handbook page. Let me know what you think!

zlonko:
Noticing the merge conflicts; I will resolve those once we make a decision on link accessibility!

bretthayes (approved):
I'd say this looks good for now and then we can enhance the Figure element to include links/titles to meet a11y standards as @st0nebraker mentioned.

content/blogposts/2020/3.18/custom-search-pages.md
zlonko: Updating the `Figure` sounds like the way to go, because we will have designed a11y into the components themselves. I am happy to work on it before my last day. I set up a ticket here: https://github.com/sourcegraph/about/issues/5554.

bretthayes: You're awesome! Thanks @zlonko!

PR #5545, opened 2022-06-30
Sticky navbar and improvements

### LINKED ISSUE #5401: Desktop main navigation: pin to top
Design reported that our navigation in the about/marketing desktop site should be pinned to the top of the page. 

**Requirements:**
1. Fix the position of the main nav on desktop to the top of the page
2. Change the background of the main nav to white on scroll so that the content behind it is hidden. 
Note: We have two variations of main nav styling: 1. Without a background: for example product and use case landing pages and 2. With a background: for example the homepage and individual use case pages

Example: https://www.hedgebase.io/
* Always show the CTA's in the nav bar

bretthayes:
Closes #5401 and implements a sticky nav feature.

### Changelog
- Fixed a bug correcting the right click event button. Clicking our logo now properly downloads our brand kit 🥚
- Updated our mobile navbar button/hamburger menu to match our DLS
- Implemented sticky nav functionality so when scrolling down it sticks

### Test

2. Right click our logo to ensure the brand kit downloads
3. Resize the browser to ensure our mobile nav toggle button displays on mobile. Check light/dark version (ie: use cases page)
4. Scroll down and ensure our navbar is fixed while scrolling


src/styles/components/_Header.scss
```diff
@@ -1,5 +1,7 @@
 // stylelint-disable declaration-property-unit-whitelist
 .header {
+    transition: all 0.3s ease-in-out;
```
zlonko, line 3: The nav bar is sticky on all of the mobile devices I tested using Chrome. Due to the background color's transition time, a user can briefly see page content beneath the navbar if they are scrolling quickly. What do you think of this?

Because the effect is so brief, I don't think it is a blocking issue. I tried shortening the transition time to `0.1s`, but it persisted. We might circumvent this with a solid background color across the entire site, although the transition is more visually interesting, especially on pages like `/use-cases`!

Screen Shot 2022-07-01 at 10 00 04 AM

zlonko (approved):
Looking groovy, @bretthayes! I added note about the transition's UX below. Glad to know your thoughts 👍

st0nebraker (approved):
LGTM! Works as expected. Thanks for fixing that 🥚 bug too! The sticky nav bar is so sleek

src/styles/components/_Header.scss
bretthayes: It's a trade off I accepted considering this will only happen for scroll-happy users haha. 😝 It's subtle, but while we have page templates like the `use-case/*` pages which include the nav within the hero, I think we can apply `bg-white` to the nav parent element to avoid this. Another solution I had was to negatively position the nav above the viewport and have it fade down into view after scrolling down, but maybe this can be a secondary approach or something we can suggest during a future redesign.

zlonko: Thank you for walking through the options! Changing it to `bg-white` as you did to the parent in 21dd476 sounds good.

zlonko (approved):
Looking great, Brett!

st0nebraker (approved):

Issue #5383, opened 2022-05-24
Clean up interfaces


zlonko:
Standardize our interfaces with the same naming convention and file structure. Please see [this comment](https://github.com/sourcegraph/about/pull/5365#discussion_r886157444) for reference.

bretthayes:
Here's what I'm thinking to clean up and standardize our interfaces with file organization and naming conventions:

## Rules of thumb/style guide:
1. If one or more interfaces are needed for a specific component or page, scope and keep them within the component or page
2. If one or more interfaces are needed for many components or pages within the same directory, collocate a file and export the interfaces to import them in the necessary components or pages
3. If one or more interfaces are needed in many pages or components across the codebase, add them to our common/shared interfaces directory and export them

### Example of a scoped interface inside a component

In this case, keep scoped interfaces within the component. Note how it's possible to keep the same namespace since they are thrown away at runtime. Open to standardizing the naming convention to always be `{Name}Props` for convenient search and batch changes.

`components/Button.tsx`
```jsx
import { FunctionComponent } from 'react'

interface Button {
    text: string
}

export const Button: FunctionComponent<Button> = ({ text }) => (
    <button type="button">
        {text}
    </button>
)
```


### Example of an interface inside a component that's been imported

In this case, if the interface is used within the same directory for many sibling/colocated components/pages, add them to a collocated file.

`components/Button/buttonProps.ts`
```ts
export interface ButtonProps {
    text: string
}
```

`components/Button/index.tsx`
```jsx
import { FunctionComponent } from 'react'

import { ButtonProps } from './buttonProps'

export const Button: FunctionComponent<ButtonProps> = ({ text }) => (
    <button type="button">
        {text}
    </button>
)
```

### Example of an interface inside a component that's been imported from our common/shared interfaces directory

In this case, if the interface is used within many components/pages across the codebase, add them to our common/shared interfaces directory to be imported in the necessary components/pages.

`interfaces/index.ts`
```ts
export * from './buttonProps'
```

`interfaces/buttonProps.ts`
```ts
export interface ButtonProps {
    text: string
}
```

`component/Button.tsx`
```jsx
import { FunctionComponent } from 'react'

import { ButtonProps } from '@interfaces'

export const Button: FunctionComponent<ButtonProps> = ({ text }) => (
    <button type="button">
        {text}
    </button>
)
```

zlonko:
[...]

PR #5547, opened 2022-06-30
Interface Clean Up

zlonko:
This closes [#5383](https://github.com/sourcegraph/about/issues/5383). 

### Changelog
- Uses the naming convention defined in the issue
- Ensures interfaces are used in components
- Renames TypeScript files with exported interfaces
- Destructures `React` imports across the repo (e.g., `React.ReactNode` -> `ReactNode`)

### Notes
- Collocation seemed like the best call for the existing interfaces, with the exception of `postProps.ts`
- Please let me know if you see additional opportunities for aligning these to the new standards!

### Test

2. Please check that there are no breaking changes due to updated names and imports


src/components/Blockquote.tsx
```diff
@@ -5,24 +5,47 @@ import Link from 'next/link'
...
 
-interface Logo {
+interface LogoProps {
     src: string
     alt: string
     href?: string
 }
 
-interface Link {
+interface LinkProps {
```
st0nebraker, line 14: Since `Logo` and `Link` aren't components, should they have this naming convention, @bretthayes ?

src/components/Carousels/QuoteCarousel.tsx
```diff
@@ -7,7 +7,8 @@ import Carousel from 'react-bootstrap/Carousel'
 import { BlockquoteWithLogoBottom } from '@components'
 import { breakpoints } from '@data'
 import { useWindowWidth } from '@hooks'
-interface Blockquote {
+
+interface BlockquoteProps {
```
st0nebraker, line 11: This is probably a good spot to re-use `BlockquoteWithLogoBottomProps` from `Blockquote.tsx`, no? Then we can import it at the top

src/components/Carousels/TabCarousel.tsx
```diff
@@ -9,22 +9,22 @@ import { useCarousel } from '@hooks'
...
-    items: Template[]
+    items: TemplateProps[]
     currentItem?: ReactNode
     previousItem?: ReactNode
     currentItemIndex?: number
     autoAdvance?: boolean
 }
 
-interface Template {
+interface TemplateProps {
```
st0nebraker, line 19: Same question here on naming convention. To me, it doesn't make sense to have the suffix be `Props` if the interface isn't directly applied to a component, but is nested w/in. But maybe I'm misunderstanding our convention. Can you clarify, @bretthayes ?

src/components/CaseStudies/CaseStudyCard.tsx
```diff
@@ -102,7 +102,7 @@ export const CASESTUDIES: CaseStudy[] = [
     },
 ]
 
-export const CaseStudyCard: FunctionComponent<{ study: CaseStudy; bwLogo?: boolean }> = ({
+export const CaseStudyCard: FunctionComponent<{ study: CaseStudyProps; bwLogo?: boolean }> = ({
```
st0nebraker, line 105: There should be an interface here for `CaseStudyCardProps`, and `CaseStudy` is separate for each study, not the components props, right?

src/components/CaseStudies/CaseStudyLayout.tsx
```diff
@@ -1,31 +1,31 @@
...
     image?: string
 }
 
-interface Props {
+interface CaseStudyLayoutProps {
     customer: string
     title: string
     logo?: string
-    quote?: Quote
+    quote?: QuoteProps
```
st0nebraker, line 19: Same convention query for `QuoteProps`

st0nebraker (commented):
Stopping for an open question around our naming convention, @bretthayes . I personally don't like naming every single interface with a `Props` suffice as not every interface is applied as component props, so to me this seems confusing.

src/components/Blockquote.tsx
bretthayes: I agree! No, I don't think they should have the same naming convention. I'm reconsidering the naming convention and I think we should revisit the discussion in the issue for clarification.

src/components/Carousels/TabCarousel.tsx
bretthayes: Yeah, I agree! I think we can forgo the Props suffix for interfaces that aren't directly applied to components. Let's go with that!

src/components/CaseStudies/CaseStudyCard.tsx
bretthayes: Yes definitely! That makes a lot more sense!

src/components/CaseStudies/NewCaseStudyLayout.tsx
```diff
@@ -1,35 +1,35 @@
...
 import { CaseStudyCard, CASESTUDIES } from './CaseStudyCard'
 
-interface Quote {
+interface QuoteProps {
     text: string
     author: string
 }
 
-interface Logo {
+interface LogoProps {
```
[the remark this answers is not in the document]
bretthayes: The shape of some interfaces are different but share the same name so it's a bit confusing. If I see some similarities though I'll consolidate and share them globally.

src/components/CaseStudies/UseChallengeSolutionResults.tsx
```diff
@@ -6,29 +6,34 @@ import ClipboardTextOutlineIcon from 'mdi-react/ClipboardTextOutlineIcon'
...
-interface Props {
-    useCases: TextLink[]
-    challenges: TextLink[]
-    solutions: TextLink[]
-    results: TextLink[]
+interface UseChallengeSolutionResultsProps {
+    useCases: TextLinkProps[]
+    challenges: TextLinkProps[]
+    solutions: TextLinkProps[]
+    results: TextLinkProps[]
```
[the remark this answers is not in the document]
bretthayes: Yeah we can remove the suffix

bretthayes:
Closing and abandoning this PR due to too many global changes, merge conflicts, and reconsidering our naming conventions for interfaces.

PR #5550, opened 2022-07-01
Add CTA link at the bottom of every post

mplacona:
### Test

### DIFF
--- a/src/components/Blog/PostLayout.tsx
+++ b/src/components/Blog/PostLayout.tsx
@@ -3,7 +3,7 @@ import { FunctionComponent } from 'react'
 import { MDXRemote } from 'next-mdx-remote'
 import Link from 'next/link'
 
-import { BlockquoteWithBorder, EmbeddedHubSpot, Video, YouTube } from '@components'
+import { BlockquoteWithBorder, EmbeddedHubSpot, Video, YouTube, TrySourcegraph } from '@components'
 import { PostComponentProps } from '@interfaces/posts'
 import { formatDate } from '@util'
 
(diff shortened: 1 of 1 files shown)


src/components/Blog/PostLayout.tsx
```diff
@@ -73,6 +73,10 @@ export const PostLayout: FunctionComponent<PostComponentProps> = ({
                 <div className={`blog-post__html ${contentClassName}`}>
                     <MDXRemote {...content} components={components as Components} />
                 </div>
+                <div className="try-sourcegraph">
+                    <h4>Interested in seeing a demo of Sourcegraph?</h4>
+                    <p><a href="https://about.sourcegraph.com/demo">You can request a demo here.</a></p>
+                </div>
```
bretthayes, line 79: We actually have a `TrySourcegraph` common component that we could use here which includes a relative Link rather than an absolute link to the demo page. I think that would be preferred for design consistency, code-reuse, and the included tracking set on the link. Would that be suitable? You can see this section on our [Use Cases page](https://about.sourcegraph.com/use-cases) right before the footer.

bretthayes, line 79: We actually have a `TrySourcegraph` common component that we could use here which includes a relative Link rather than an absolute link to the demo page. I think that would be preferred for design consistency, code-reuse, and the included tracking set on the link. Would that be suitable? You can see this section on our [Use Cases page](https://about.sourcegraph.com/use-cases) right before the footer.

mplacona: Can you please a look at [this commit](fe511e7d9e1c68a53fc5b4f964aa7eceae1b3cac) and tell me if this is how you see it working? I kept the box the same size as the post so it looks like it's part of it rather than just a floating container.

bretthayes: Awesome! That works! Thanks for updating that!

bretthayes (approved):

PR #5560, opened 2022-07-07
Redirect lingering get started links to self hosted

bretthayes:
Fixes some lingering links to point to `/get-started/self-hosted` as a follow up of #5508


netlify.toml
```diff
@@ -160,10 +160,8 @@
 [[redirects]]
   from = "/get-started"
   to = "/get-started/self-hosted"
-  force = true
 
 [[redirects]]
   from = "/get-started/cloud"
   to = "/get-started/self-hosted"
-  force = true
```
zlonko: I included `force = true` in the #5508 because the redirects appeared to require them in order to work. When I test Netlify, `/get-started` and `/get-started/cloud` do not redirect to `/get-started/self-hosted`. Is there another way to ensure this?

zlonko: Just noticed you updated this in 2697b2e4c32c2164729aa3ce1ab78a3288297f5d! Nvm! Also, "use the force"... thank you for this 🥸

PR #5593, opened 2022-07-20
Remove approval workflow

bretthayes:
As part of SOC2, we shouldn't have this workflow and we should always require approvals for PRs.


st0nebraker (approved):
🔐 🌈 👏🏻

Issue #5378, opened 2022-05-11
Create Tailwind Theme


bretthayes:
Create a Tailwind theme that aligns with our Brand Guidelines, Marketing DLS, design consistency concerns, and UI standardizations as the foundation for our Tailwind Conversions in sourcegraph/about#5381.

PR #5615, opened 2022-07-27
Code Search Page Refresh

### LINKED ISSUE #5539: /code-search page refresh
Stakeholders: Product Marketing
What?: Updates to the Code Search product page to bring it more in line with the Code Insights page
Why?: Code Intelligence (the product) is now code navigation! After lots of research & consideration, we’re excited to roll out a name change for our Code Intelligence product. Going forward, Code Intelligence will become code navigation, which will better match what our customers refer to it as and what other technologies (like IDEs) call this functionality.
**What’s changing?**
We are changing the customer-facing, external name of the Code Intelligence product. We are not changing our platform positioning nor renaming the Code Intel team
The frontend navigation functions — such as Find References — will now be code navigation
We will refer to LSIF & SCIP indexer data as code graph data to give this technology extra weight & importance. This data — and precise code navigation — is a big competitive differentiator for Sourcegraph
Assets:
[Figma file](https://www.figma.com/file/3DjJkWg9tYOpT7KyfqQsTZ/Code-search-refresh?node-id=0%3A1) 
[Other edits to about](https://docs.google.com/document/d/14JVbWBanripFhd1Jdg8AarlslpI9sJZY0v-zoqeLDG4/edit#)

**When will this change take place?**
We are announcing the beta release of auto-indexing in 2 weeks, on July 28. We will introduce this new naming in that release’s blog post and update our docs + website to reflect the change

Asana: https://app.asana.com/0/1200095757619647/1202519591393084/f

bretthayes:
This closes #5539 with a new refresh for the code search page.

### Changelog
- Moved our animations to the public directory since they're now being used in other places
- Added textured backgrounds, illustrations, and a new Background component via our DLS
- Updated container widths, breakpoints, grid sizes, heading font sizes, and heading font weights globally to meet our new brand guidelines/DLS in preparation for #5378
- Updated `BlogResourceItem` to meet new designs
- Renamed `FeatureWalkthrough` to `CoreFeatures`
- Updated `IntegrationsSection` to meet new designs
- Removed Code Intelligence from top nav
- Renamed a few instances of Code Intelligence to Code Navigation as per other edits requested in the issue
- Added a `ProductHero` component specific to product pages and updated all product pages to use it. There is potential to consolidate this into a catchall `Hero` component but I didn't want to overcomplicate this component for this PR and include all variants.
- Added a `TwoColumnSection` component. Same concern here with overcomplicating and not including all variants. There needs to be more discussion about Atomic Design and consolidation for our DLS to make development easier.
- Added `grayscale-200` and `grayscale-500` to our colours - didn't want to include all new brand colours in this PR since they'll be included in #5378 
- Updated the gradient venus utility class to meet new designs under backgrounds. Same concern here with not including all updates since #5378 will cover this next
- Removed code insights and and batch changes styles/tech debt

### Test


src/components/BlogResourceItem.tsx
```diff
@@ -19,24 +21,37 @@ interface Blog {
...
-            </a>
-            <p className="text-gray-5">{blog.type}</p>
-            <p>{blog.description}</p>
-        </div>
-        <div className="col-sm-4 col-md-3 d-flex align-items-center">
-            {blog.img && <img className="flex-1 w-100" alt={blog.img.alt} src={blog.img.src} />}
+export const BlogResourceItem: FunctionComponent<Props> = ({ blog }) => {
+    const defaultThumbnails: string[] = ['darkNebulous1Sm', 'darkNebulous2Sm', 'darkNebulous3Sm', 'darkNebulous4Sm']
+    const thumbnail: string = defaultThumbnails[Math.floor(Math.random() * 4)].toString()
+    const variant: Background['variant'] = thumbnail as Background['variant']
```
bretthayes, line 27: This creates a random thumbnail if no img is provided for `BlogResourceItem`. I figured this would be better than not having an image at all and gives some differentiation between posts.

st0nebraker (changes requested):
Looks good! Lots of enhancements here, thanks Brett 💥 . Just a few comments below

src/components/ProductHero.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { Background, HubSpotForm } from '@components'
+
+interface ProductHero extends Background {
+    product: 'code search' | 'batch changes' | 'code insights'
+    title: string | ReactNode
+    description: string
+    displayUnderNav?: boolean
+}
+
+export const ProductHero: FunctionComponent<Omit<ProductHero, 'className' | 'children' | 'illustration'>> = ({
```
st0nebraker, line 12: Can you explain why you're omitting some properties on here? `Omit` is a bit new to me

src/styles/_colors.scss
```diff
@@ -48,6 +48,10 @@ $gray-4: #8a8a8a;
 $gray-3: #c4c4c4;
 $gray-2: #e7e7e7;
 
+// New DLS colors
+$gray-200: #ebeff3;
+$gray-500: #484B51;
```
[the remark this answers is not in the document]
bretthayes: Yes, I plan to clean this up and delete all old colours once when I tackle #5378

src/components/CoreFeatures.tsx
```diff
@@ -6,15 +6,6 @@ import Link from 'next/link'
 
 import { buttonStyle, buttonLocation } from '@data'
```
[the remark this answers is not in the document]
bretthayes: Awesome! Thanks for catching this! Now that those DLS specs are in, I'm sure lots more one-off styles we've added in the past can be removed 🤓

src/components/Background/index.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+interface IllustrationStyle {
+    [key: string]: {
+        size: string
+        position: string
+    }
+}
+
+// Background variant to image mapping
+const backgrounds: { [key: string]: StaticImageData } = {
```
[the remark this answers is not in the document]
bretthayes: That's the plan! I'd like for us to rework that so we don't have to inject classes like that in the Layout 🤓

src/components/Background/index.tsx
```diff
@@ -0,0 +1,175 @@
...
+ * @param props.illustration - an optional string for an illustration bg image
+ * @param props.className - optional classNames
+ * @param props.style - optional CSS style properties
+ * @returns ReactNode
+ */
+export const Background: FunctionComponent<Background> = ({ variant, children, illustration, className, style }) => {
+    const windowWidth = useWindowWidth()
+    const isMobile = windowWidth < breakpoints.lg
+
+    const backgroundSource: string = backgrounds[variant].src
```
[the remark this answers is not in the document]
bretthayes: It's unnecessary since the variant is always required and it's a union type that lists the available values.

src/components/ProductHero.tsx
bretthayes: Yeah for sure! It's because we don't want to allow those props in this component since they're not used directly and there's no need to do prop drilling. We also wouldn't want to do unnecessary prop drilling here. It's better to be more explicit about your props in your component composition than to allow all the things.

bretthayes: Also, this component may or may not be consolidated into a global Hero component. I think there needs to be some more consolidation in our DLS so we can have a clear path about props and prevent overcomplicating components.

st0nebraker: Yea I think this component can be used for all heros too! -- On the omission, am I missing something, because `ProductHero` interface doesn't have `'className' | 'children' | 'illustration'` properties. So is this necessary?

src/components/Background/index.tsx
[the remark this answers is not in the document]
st0nebraker: Ah I see! Nevermind 😄

st0nebraker (approved):

src/components/ProductHero.tsx
bretthayes: The `ProductHero` interface extends the `Background` interface and then adds a few properties that are specific to the Product Hero 🤓

st0nebraker: Oh geez, missed that part 🤦🏻‍♀️  - thanks for explaining!

PR #5634, opened 2022-07-29
remove note about in-development functionality

rvantonder:
This functionality should not be advertised broadly to the public at this time, it is only undergoing internal testing currently.

### Test
change to post

### DIFF
--- a/content/blogposts/2022/release-3.42.md
+++ b/content/blogposts/2022/release-3.42.md
@@ -57,12 +57,6 @@ This is especially useful for engineers joining new teams or onboarding to a cod
 
 ![Search for repository by description](https://storage.googleapis.com/sourcegraph-assets/docs/images/code_search/repo-description-search.png)
 
-## Smart search suggestions with lucky search
-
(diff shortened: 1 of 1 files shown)


benvenker (approved):

rvantonder:
@iskyOS we go ahead with this merge because encouraging users to change their settings for this feature is potentially damaging to maintain and can significantly increase debt/settings issues.

rvantonder:
@iskyOS can you comment on our current process for including content in release announcements? Ever since I worked at Sourcegraph, my understanding is that the Changelog is the source of truth for whether something goes into the release blog post. In this case, this note didn't have a corresponding Changelog entry (and was my way of ensuring that it shouldn't go into the blog post). So two questions:

- If we are still relying on the Changelog as the source of truth, can we cross-check that blog post release announcements have a corresponding Changelog entry?

- If we are not fully relying on the Changelog to inform blog post content, then it is crucial that the Changelog _does_ get a corresponding entry if we decide to publish an announcement coming from some other source. This Changelog entry can then be reviewed for accuracy. Can we ensure that happens going forward?

Asking the above because either would have mitigated this announcement slipping in

PR #5655, opened 2022-08-05
Codecov Case Study/headshots

### LINKED ISSUE #5635: Add team headshots to Codecov Case Study
Who?: Lindsay, Fabiana
What?: Add team headshots to Codecov Case Study shipped last week
Why?: They were missing from the original draft, but are now ready

Asana: https://app.asana.com/0/1200804416400102/1202595954078382/f

st0nebraker:
Closes #5635. Adds team headshots to Codecov Case Study & adjusts `<AuthorBio />` to have a multi-author design.

### Changelog
""

### Test
2. Nav to `/case-studies/codecov-uses-sourcegraph-to-resolve-incidents-faster` and ensure headshots meet design
3. Ensure no breaking changes to the Nutanix case study `<AuthorBio />` section


src/components/CaseStudies/AuthorBio.tsx
```diff
@@ -1,23 +1,62 @@
-export const AuthorBio: React.FunctionComponent<{
-    customer: string
+interface Author {
     image?: string
-    author: string
+    name: string
```
bretthayes, line 3: This is so much better! 👏🏻 

Author.author ❌
Author.name ✅

src/components/CaseStudies/AuthorBio.tsx
```diff
@@ -1,23 +1,62 @@
...
+                    {authors.map(a => (
+                        <div key={a.name}>
+                            <img
+                                className="rounded-circle max-w-150"
+                                width={150}
+                                height={150}
+                                src={a.image}
+                                alt={a.name}
+                            />
+                            <h3 className="font-weight-normal">{a.name}</h3>
```
bretthayes, line 31: Looks like the Figma has some more spacing between the thumbnail and the header so we'll probably need to add a top margin here

src/components/CaseStudies/AuthorBio.tsx
```diff
@@ -1,23 +1,62 @@
...
+            // Multi author design
+            <div>
+                <section className="col-lg-6 mx-auto p-0 max-w-650">
+                    <h5 className="mb-4 text-xl font-weight-bold">About {customer}</h5>
+                    <p>{about}</p>
+                </section>
+
+                <section className="d-md-flex d-block justify-content-between col-lg-9 mx-auto mt-6 p-0">
+                    {authors.map(a => (
+                        <div key={a.name}>
```
bretthayes, line 23: Looks like the author sections have some bottom spacing between them. We should add a bottom margin here on mobile when they stack, except for the last one so it doesn't push the section further than it needs to.

Just to note, in Tailwind after #5378 gets merged, you can do this mobile-first by first setting the mobile styles, and then override the breakpoint when they align horizontally on tablet/desktop. (ie: `tw-mb-sm md:tw-mb-0`) The `xs/sm/md/lg/xl` spacings are defined from our DLS, but you could also use numbers (ie: `tw-mb-04`). The breakpoint modifiers come first with a colon that references the screen size (ie: `md:tw-mb-0` would reference the `md` breakpoint and set the bottom margin to `0`).

You can add these when `index !== index.length - 1`

bretthayes (changes requested):
Both singular and multiple author layouts look good. Just a few changes need to be made:

- noted some spacing differences between this and the Figma, and;
- we should crop and compress the thumbnails to be fairly similar. I'd probably pick the smallest thumbnail and go with that for the rest. Mitchell's thumbnail is 2.17 MB so that's quite high. ⚠️ We should always aim to be under or around the 100 KB range for thumbnails of this size, depending on quality and image details.

src/components/CaseStudies/AuthorBio.tsx
st0nebraker: Good eye! Both margins are addressed in [a8dd2ef](https://github.com/sourcegraph/about/pull/5655/commits/a8dd2ef59eadfa66fd31e87fae5b5e3ab95019db) 😄

st0nebraker:
> * we should crop and compress the thumbnails to be fairly similar. I'd probably pick the smallest thumbnail and go with that for the rest. Mitchell's thumbnail is 2.17 MB so that's quite high. ⚠️ We should always aim to be under or around the 100 KB range for thumbnails of this size, depending on quality and image details.

I got Mitchells image down to 682KB but can't get it much smaller. `jpegoptim` is erroring out on it. Any other recs to compress high quality images?

bretthayes:
> > * we should crop and compress the thumbnails to be fairly similar. I'd probably pick the smallest thumbnail and go with that for the rest. Mitchell's thumbnail is 2.17 MB so that's quite high. ⚠️ We should always aim to be under or around the 100 KB range for thumbnails of this size, depending on quality and image details.
> 
> I got Mitchells image down to 682KB but can't get it much smaller. `jpegoptim` is erroring out on it. Any other recs to compress high quality images?

If you crop Mitchell's thumbnail to jeff's or jerrod's image size, and then compress it afterwards, that should do the trick!

bretthayes (changes requested):
Try cropping the image before compression

bretthayes (approved):
This looks good to go. 🚀 

Side note: I cropped the image by 50% and used 124% higher compression. Even though PNGs are lossless, JPGs tend to have a smaller file size which is better for loading performance. Since these images are so small, this seems more suitable.

For assets, let's use JPGs where possible for low file size with 70% quality, and if a PNG has a low file size  (just as good) for high quality, let's use that. Otherwise use SVGs where possible.

st0nebraker:
> Side note: I cropped the image by 50% and used 124% higher compression. Even though PNGs are lossless, JPGs tend to have a smaller file size which is better for loading performance. Since these images are so small, this seems more suitable.
> 
> For assets, let's use JPGs where possible for low file size with 70% quality, and if a PNG has a low file size (just as good) for high quality, let's use that. Otherwise use SVGs where possible.

Hm it doesn't look more cropped than my last crop. What tool did you use for cropping and compressing? Maybe would be good to update our image standards in our team docs somewhere for reference. I'm not as versed with image size rules and tools and happy to continue reinforcing our standards! @bretthayes

bretthayes:
> > Side note: I cropped the image by 50% and used 124% higher compression. Even though PNGs are lossless, JPGs tend to have a smaller file size which is better for loading performance. Since these images are so small, this seems more suitable.
> > For assets, let's use JPGs where possible for low file size with 70% quality, and if a PNG has a low file size (just as good) for high quality, let's use that. Otherwise use SVGs where possible.
> 
> Hm it doesn't look more cropped than my last crop. What tool did you use for cropping and compressing? Maybe would be good to update our image standards in our team docs somewhere for reference. I'm not as versed with image size rules and tools and happy to continue reinforcing our standards! @bretthayes

Check out the branch, pull the latest changes and check out the image info in the Finder sidebar. You can also "right click>save as" on the image in my commit and check the "get info" pane on the image as an alternative.

Your crop in [1804db6](https://github.com/sourcegraph/about/pull/5655/commits/1804db6ccfe082ea76d85be8ead523a0c1e6e7bb) was 1088x1088 and 437KB. I cropped it to 600x600 at 104KB in my commit [b70dda5](https://github.com/sourcegraph/about/pull/5655/commits/b70dda5c7a910964f9a581635dfeee2b84f41cb7). There won't be a swipe/onion skin UI in my commit since the original file wasn't modified but replaced with a new format.

For standards/rules, it depends on the size of the image but there's lots of resources and suggestions around for the best aspect ratios, image dimensions, file sizes, and formats, depending if they're a logo, full width hero image, or thumbnail. We can definitely document this more in the handbook and come up with some standard guidelines to follow. As a simple rule of thumb, SVGs are preferred, PNGs are great for transparent images but they tend to be higher file size so they increase loading times, and JPGs tend to have lower file size which is great for loading times but can sometimes have quality drawbacks.

PR #5717, opened 2022-09-06
new blog cta

beyang:
@jdorfman wdyt of blog post CTAs at the end like this?
* Subscribe to newsletter (dropped in a placeholder substack subscription page)
* Join Discord

out


jdorfman (approved):
Love it

PR #5756, opened 2022-09-20
Update multi-version base version guarantee

efritz:
### Test


efritz:
This appears outdated now.

Issue #5690, opened 2022-08-24
4.0 landing page post-launch updates


elzannewentzel:
Stakeholders: Lori, Andy, PMM
What: Updates to the existing [4.0 landing page](https://about.sourcegraph.com/sourcegraph-4) after 5PM on 27 Sept
Why: The page needs to be updated with relevant content after the launch 

Assets:
 - [4.0 landing page](https://www.figma.com/file/FrXCB2MwgntaBDtkQLvtYz/Sourcegraph-4.0?node-id=671%3A7998)
 - [Homepage banner update ](https://www.figma.com/file/FrXCB2MwgntaBDtkQLvtYz/Sourcegraph-4.0?node-id=424%3A6646)
 - Meta

Requirements:

- [x] ~Update the homepage banner~
- [x] Remove the Twitch embed
- [x] Add highlights video to the hero
- [x] Change the "See the highlights from Sourcegraph 4.0" heading to "Experience 4.0 with the Sourcegraph Engineering team" 
- [ ] Add livestream recording to the bottom of the page

PR #5791, opened 2022-09-27
Add livestream video to 4.0 landing

bretthayes:
Follow up for #5690, adding the livestream video. 

### Changelog
- added livestream video to the bottom of the page using the TwitchPlayer
- removed Bootstrap classes for YouTube video aspect ratios in lieu of Tailwinds video aspect ratio utility class

### Test

2. Ensure livestream video plays

### DIFF
--- a/src/components/YouTube.tsx
+++ b/src/components/YouTube.tsx
@@ -32,9 +32,9 @@ export const YouTube: FunctionComponent<YouTube> = ({
     showTitle = false,
 }) => (
     <figure>
-        <div className={`video-embed embed-responsive embed-responsive-16by9 ${className}`}>
+        <div className={`tw-aspect-video ${className}`}>
             <iframe
-                className="embed-responsive-item"
+                className="tw-w-full tw-h-full"
                 src={`https://www.youtube-nocookie.com/embed/${id}?autoplay=${autoplay ? 1 : 0}&cc_load_policy=${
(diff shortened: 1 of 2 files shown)


st0nebraker (changes requested):
One thing I noticed for mobile (I know there wasn't a spec for mobile), maybe the titles should left-align at small breakpoints and below? Since that's our pattern on most pages. Other aspects of the same page do this already, like the `Top updates released in Sourcegraph 4.0` section. What do you think?
image

traceyljohnson:
@st0nebraker Something to double check is that the H1 and H2's go down to their respective mobile sizes (6xl and 4xl). That will help it to not break so oddly (unless they already are those sizes and I'm just mistaken).

st0nebraker:
> @st0nebraker Something to double check is that the H1 and H2's go down to their respective mobile sizes (6xl and 4xl). That will help it to not break so oddly (unless they already are those sizes and I'm just mistaken).

Looks like they are the right size, just sits a bit odd at that 320px XS screen width

fabicastp (approved):

st0nebraker (approved):
LGTM! Tested the Youtube component on a few other pages & looks groovy 😎

bretthayes:
Awesome! Thanks Beckz and Fabz! 🎉

PR #5824, opened 2022-10-05
Rm Bootstrap > Custom Tooltip & Accordion

### LINKED ISSUE #5746: Change Bootstrap Accordion/Tooltip => Flowbite
Follow up to #5674. Migrate Accordion & Tooltip components to Flowbite-React library.

This will require configuring [flowbite-react](https://flowbite-react.com/) to play well with our Tailwind config.

This supports our effort to remove all Bootstrap libraries from the repo.

st0nebraker:
Closes #5746, removes Bootstrap Accordion & Tooltip imports in favor of custom components to reduce bundle size.

### Changelog
- Adds a custom Accordion component
- Adds a custom Tooltip component (with accessibility considerations for tabbing)

### Test
2. Nav to `/pricing` and ensure no breaking UX in the tooltips or accordion


src/components/Tooltip.tsx
```diff
@@ -0,0 +1,34 @@
...
+                role="button"
+                tabIndex={0}
+            >
+                {children}
+                {isVisible && (
+                    <div
+                        className={`tooltip ${position} tw-min-w-[250px] tw-p-xxs tw-shadow-lg tw-bg-gray-400 tw-white tw-opacity-100 tw-text-white tw-rounded-md`}
+                    >
+                        {text}
+                    </div>
```
st0nebraker, line 29: I just need to write some logic to take care of the overflow offscreen issue here. Since the position can be dynamic, I'm not sure if it should be taken care of here or by what's passed in one level higher by the parent component.
image

bretthayes (changes requested):
This is a great step forward! Nice work Becca! Some suggestions below:

src/components/Tooltip.tsx
st0nebraker: The tooltip to be relative to it's icon though, so it needs to know where the parent is 🤔 I'm not sure how well that can work with Portal. Or am I missing something? Where would you put a relative positioning for this using Portal?

bretthayes: You could render the tooltip popover itself within the portal which would solve the z-index issues, and then use a ref on the icon, so that when you hover, you can listen for a mouse move event and find the pageX and pageY coordinates to position the popover accordingly. There's some clever examples floating around on Stack Overflow for viewport edge detection 🤓 @st0nebraker

PR #5853, opened 2022-10-13
Update git-vs-perforce.md

lcolston08:
Changes per Salesforce's request to: 

With Git, every developer has a complete copy of the codebase locally. This is often an advantage, but not always. For example, it can be inefficient for large changes as every change has to be copied many times to each developer’s machine.

Devs at Salesforce can now directly search for potential issues using Sourcegraph, rather than wasting time going through change lists or using multiple tools to track down those issues.

### Test


abschu (approved):
Looks good!

PR #5863, opened 2022-10-19
Cta changes product pages

elzannewentzel:
Made CTA changes to product pages

### Changelog
 - Removed Hubspot form from product page hero
 - Added two CTAs to the product page hero

### Test

2. Test all CTAs redirects to the correct pages

### DIFF
--- a/src/pages/code-insights.tsx
+++ b/src/pages/code-insights.tsx
@@ -319,33 +319,33 @@ const CodeInsightsPage: FunctionComponent = () => (
                 subtitle="Transform your code into a queryable database to create customizable, visual dashboards in seconds."
                 cta={
                     <div className="tw-text-center tw-flex-col md:tw-flex-row md:tw-flex">
-                    <div className="mb-3 mb-md-0">
-                    <a
-                            href="https://signup.sourcegraph.com"
-                            className="btn btn-primary w-100 max-w-350"
-                            title="Get free trial"
-                            data-button-style={buttonStyle.outline}
-                            data-button-location={buttonLocation.hero}
-                            data-button-type="cta"
-                        >
-                            Get free trial
-                        </a>
-                    </div>
-                    <div className="ml-md-3">
-                    <Link href="/demo" passHref={true}>
-                            {/* eslint-disable-next-line jsx-a11y/anchor-is-valid */}
+                        <div className="mb-3 mb-md-0">
                             <a
-                                className="btn btn-outline-primary w-100 max-w-350"
-                                title="Request a Demo."
-                                data-button-style={buttonStyle.primary}
+                                href="https://signup.sourcegraph.com"
+                                className="btn btn-primary w-100 max-w-350"
+                                title="Get free trial"
+                                data-button-style={buttonStyle.outline}
                                 data-button-location={buttonLocation.hero}
                                 data-button-type="cta"
                             >
-                                Request a demo
+                                Get free trial
                             </a>
-                        </Link>
+                        </div>
+                        <div className="ml-md-3">
+                            <Link href="/demo" passHref={true}>
+                                {/* eslint-disable-next-line jsx-a11y/anchor-is-valid */}
+                                <a
+                                    className="btn btn-outline-primary w-100 max-w-350"
+                                    title="Request a Demo."
+                                    data-button-style={buttonStyle.primary}
+                                    data-button-location={buttonLocation.hero}
+                                    data-button-type="cta"
+                                >
+                                    Request a demo
+                                </a>
+                            </Link>
+                        </div>
                     </div>
-                </div>
                 }
                 displayUnderNav={true}
             />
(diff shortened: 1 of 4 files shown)


src/pages/code-insights.tsx
```diff
@@ -317,7 +317,36 @@ const CodeInsightsPage: FunctionComponent = () => (
                 product="code insights"
                 title={'Track what really matters\nto you and your team.'}
                 subtitle="Transform your code into a queryable database to create customizable, visual dashboards in seconds."
-                cta={<HubSpotForm masterFormName="contactEmail" chiliPiper={true} />}
+                cta={
+                    <div className="tw-text-center tw-flex-col md:tw-flex-row md:tw-flex">
```
bretthayes, line 321: We've got a slightly different issue with this page after changing the alignment to `left`. You'll notice the "Request a demo" button not spanning the full width. This is because the `Hero` component has a `max-w-400` class applied on line `61` in `Hero.tsx`. I think we should change that to `350px` to match with the max width of these buttons and that would fix this issue.

We could also potentially remove `max-w-350` from each button since it would be redundant at that point since each button spans the full width of the parent with `w-100`, but that's not blocking; just a cleanup suggestion.

```suggestion
                    <div className="tw-text-left tw-flex-col md:tw-flex-row md:tw-flex">
```

Screen Shot 2022-10-19 at 12 56 42 PM

src/pages/code-insights.tsx
```diff
@@ -317,7 +317,36 @@ const CodeInsightsPage: FunctionComponent = () => (
...
+                        <div className="mb-3 mb-md-0">
+                            <a
+                                href="https://signup.sourcegraph.com"
+                                className="btn btn-primary w-100 max-w-350"
+                                title="Get free trial"
+                                data-button-style={buttonStyle.outline}
```
bretthayes, line 327:
```suggestion
                                data-button-style={buttonStyle.primary}
```

src/pages/code-insights.tsx
```diff
@@ -317,7 +317,36 @@ const CodeInsightsPage: FunctionComponent = () => (
...
+                                Get free trial
+                            </a>
+                        </div>
+                        <div className="ml-md-3">
+                            <Link href="/demo" passHref={true}>
+                                {/* eslint-disable-next-line jsx-a11y/anchor-is-valid */}
+                                <a
+                                    className="btn btn-outline-primary w-100 max-w-350"
+                                    title="Request a Demo."
+                                    data-button-style={buttonStyle.primary}
```
bretthayes, line 340:
```suggestion
                                    data-button-style={buttonStyle.outline}
```

src/pages/code-search.tsx
```diff
@@ -53,7 +55,36 @@ export const CodeSearchPage: FunctionComponent = () => (
                 product="code search"
                 title={'Search your code.\nAll of it.'}
                 subtitle="Onboard to a new codebase, understand code faster, and identify security risks with universal code search."
-                cta={<HubSpotForm masterFormName="contactEmail" chiliPiper={true} />}
+                cta={
+                    <div className="tw-text-center tw-flex-col md:tw-flex-row md:tw-flex">
```
bretthayes, line 59:
```suggestion
                    <div className="tw-text-left tw-flex-col md:tw-flex-row md:tw-flex">
```

src/pages/code-search.tsx
```diff
@@ -53,7 +55,36 @@ export const CodeSearchPage: FunctionComponent = () => (
...
+                        <div className="mb-3 mb-md-0">
+                            <a
+                                href="https://signup.sourcegraph.com"
+                                className="btn btn-primary w-100 max-w-350"
+                                title="Get free trial"
+                                data-button-style={buttonStyle.outline}
```
bretthayes, line 65:
```suggestion
                                data-button-style={buttonStyle.primary}
```

src/pages/code-search.tsx
```diff
@@ -53,7 +55,36 @@ export const CodeSearchPage: FunctionComponent = () => (
...
+                                Get free trial
+                            </a>
+                        </div>
+                        <div className="ml-md-3">
+                            <Link href="/demo" passHref={true}>
+                                {/* eslint-disable-next-line jsx-a11y/anchor-is-valid */}
+                                <a
+                                    className="btn btn-outline-primary w-100 max-w-350"
+                                    title="Request a Demo."
+                                    data-button-style={buttonStyle.primary}
```
bretthayes, line 78:
```suggestion
                                    data-button-style={buttonStyle.outline}
```

elzannewentzel (commented):
Thank you! Suggestions implemented :)

src/pages/code-insights.tsx
bretthayes: I think if we change the `max-w-400` class on line 61 in `components/Hero.tsx`, then that should fix this issue. Then we should be good to go!

bretthayes (changes requested):
I think we need to change the parent max width on that Hero component div that holds the CTAs. Then we should be good to go!

bretthayes (approved):

PR #5872, opened 2022-10-22
Update multi-version-upgrades.md

tzarsquared:
### Test


content/blogposts/2022/multi-version-upgrades.md
```diff
@@ -18,89 +18,89 @@ The most obnoxious restriction (by a very wide margin) related to administration
...
 
-Because this problem would only get worse with time, we began to squash migration definitions by replacing the set of definitions at the head of the sequence with a single (yet equivalent) migration file. For Sourcegraph v3.9.0, we [squashed definitions](https://github.com/sourcegraph/sourcegraph/pull/5967) that were also defined in Sourcegraph v3.7.0 into a single file. This cut startup time with an empty database from 20s to 5s and also reduced the chance of hitting the startup timeout in our testing environment.
+Because this problem would only get worse with time, we began to squash migration definitions by replacing the set of definitions at the head of the sequence with a single (yet equivalent) migration file. For [Sourcegraph v3.9.0](https://github.com/sourcegraph/sourcegraph/pull/5967), we squashed definitions that were also defined in Sourcegraph v3.7.0 into a single file. This cut startup time with an empty database from 20s to 5s and also reduced the chance of hitting the startup timeout in our testing environment.
```
efritz, line 35: The link here was to a PR that squashed the definitions, so the link should be on the action (the Sourcegraph version isn't relevant).

efritz (approved):

content/blogposts/2022/multi-version-upgrades.md
iskyOS:
```suggestion
Because this problem would only get worse with time, we began to squash migration definitions by replacing the set of definitions at the head of the sequence with a single (yet equivalent) migration file. For Sourcegraph v3.9.0, we [squashed definitions](https://github.com/sourcegraph/sourcegraph/pull/5967) that were also defined in Sourcegraph v3.7.0 into a single file. This cut startup time with an empty database from 20s to 5s and also reduced the chance of hitting the startup timeout in our testing environment.
```

PR #5896, opened 2022-11-02
Final edits

jdorfman:
### Test

### DIFF
--- a/content/blogposts/2022/openssl-307-vulnerabilities.md
+++ b/content/blogposts/2022/openssl-307-vulnerabilities.md
@@ -1,6 +1,6 @@
 ---
 title: 'OpenSSL 3.0.7'
-description: OpenSSL 3.0.7 is available and it addresses two security vulnerabilities. Here's what you need to know.
+description: OpenSSL 3.0.7 is available and addresses two security vulnerabilities. Here's what you need to know.
 authors:
   - name: Adam Harvey 
     url: https://twitter.com/lgnome 
@@ -22,9 +22,9 @@ OpenSSL 3.0.0 to 3.0.6 is affected. No prior OpenSSL version (including OpenSSL 
 
 ### Upgrades
 
-The general advice for OpenSSL 3 users is to upgrade to OpenSSL 3.0.7 as soon as possible.
+The general advice is to upgrade to OpenSSL 3.0.7 as soon as possible.
 
-For users who are only using OpenSSL transitively as a dynamic dependency, this really means rebuilding container images and ensuring bare-metal servers are upgraded to the current version of OpenSSL. 
+If you're only using OpenSSL transitively as a dynamic dependency, this really means rebuilding container images and ensuring bare-metal servers are upgraded to the current version of OpenSSL. 
 
 ### Other mitigations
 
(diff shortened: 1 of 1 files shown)


content/blogposts/2022/openssl-307-vulnerabilities.md
```diff
@@ -11,81 +11,25 @@ heroImage:
...
-_A: Users operating TLS servers may consider disabling TLS client authentication, if it is being used, until fixes are applied._
-
-This is, of course, unhelpful for users who actually need client authentication, but it's possible some users have accidentally enabled client authentication when they don't actually require it. Some initial thoughts on possible code searches are below.
-
-This may also affect users who have code that parses arbitrary server certificates. Detecting that is less a question of code search and is more a question of product knowledge.
-
-
 #### The low level API, and C and C++ users
 
 Client authentication is ultimately enabled by setting [the SSL_VERIFY_PEER option](https://www.openssl.org/docs/manmaster/man3/SSL_CTX_set_verify.html) when issuing or handling a TLS request. The problem is that this option is also used by clients for innocuous purposes, and there are no API differences between server and client mode — or, for that matter, vulnerable and non-vulnerable versions of OpenSSL — to craft a one-size-fits-all code search.
```
amieroth, line 35: @LawnGnome would you reframe this section, or pull it? Mainly “The problem is that this option is also used by clients for innocuous purposes, and there are no API differences between server and client mode” is giving me pause.

LawnGnome: The idea there is basically just messaging that "code search is somewhere between an overly blunt tool and completely useless in this context".

A simpler version of this would be something like:

> TLS client authentication is triggered by setting the `SSL_VERIFY_PEER` flag when handling a TLS request. However, this flag is also used by unrelated client requests, so searching for `SSL_VERIFY_PEER` is unhelpful on its own.

content/blogposts/2022/openssl-307-vulnerabilities.md
```diff
@@ -124,30 +63,26 @@ However, the popular cryptography package does ship static OpenSSL libraries in
...
 ```
 
+#### Go
+
+Go uses its own, completely independent SSL implementation. No action required.
+
 ## OpenSSL Checker
 
-You can also use our bookmarklet-based OpenSSL Checker tool available at [https://sourcegraph-community.github.io/openssl-checker/](https://sourcegraph-community.github.io/openssl-checker/) to check if your own projects are affected.
+You can use our bookmarklet-based OpenSSL Checker tool available at [https://sourcegraph-community.github.io/openssl-checker/](https://sourcegraph-community.github.io/openssl-checker/) to check if your own projects are affected.
```
LawnGnome, line 86: Waters down the takeaway, but is more accurate (_insert Futurama technically correct joke here_):

```suggestion
You can use our bookmarklet-based OpenSSL Checker tool available at [https://sourcegraph-community.github.io/openssl-checker/](https://sourcegraph-community.github.io/openssl-checker/) to check if your Rust and Python projects are affected.
```

LawnGnome (commented):
One general framing question I have: I wrote this initially for an internal, non-user audience, so there's a lot of "users should X" and "users may Y". Should this be written more directly; ie "you should X", "you may Y"?

Otherwise, generally LGTM, but then it would, wouldn't it? 😆

amieroth (approved):

PR #5906, opened 2022-11-09
Update ranking comparison images to be larger

iskyOS:
Updating the PageRank blog to make the comparison image easier to read.

### Test

Blog update; no test.


abschu (commented):
@iskyOS - the text is now pretty far away from the image. What do you think about:

**Without PageRank:** The #1 search result in this image is a match inside a source comment. Although this might be relevant in some contexts, it’s generally not so great to suggest as the top result.

Image

**With PageRank:** With the pageranked results, you only see relevant matches. Over time, this adds up to tremendous time savings. And devs will always gravitate towards things that save them time.

Image

PR #5924, opened 2022-12-02
Adding second Sourcegraph<>Backstage post to the blog 

Joelkw:
This is the second blogpost on how Sourcegraph helps Backstage users set up and maintain their catalog, in collaboration with Taras and Min from Frontside. 

### Test

Reviewing the preview manually.


sqs (approved):
stamp

PR #5927, opened 2022-12-07
Update to event-logger 2.0.3

nathan-downs:
### Test

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1836,21 +1836,11 @@
     eslint-plugin-unicorn "^21.0.0"
     eslint-plugin-unused-imports "^1.1.5"
 
-"@sourcegraph/event-logger@^2.0.2":
-  version "2.0.2"
-  resolved "https://registry.yarnpkg.com/@sourcegraph/event-logger/-/event-logger-2.0.2.tgz#cef23de5afc260f6f00c67db03699de06e6d3451"
-  integrity sha512-Q2VsxsFD+fP+D7xz2dtiRxk+3UMjYDY8wewjwuoM8f1zmWPaCIWKeM8MCTbCCDwLwJXL0vdHqucHNne4lZUtMw==
+"@sourcegraph/event-logger@^2.0.4":
+  version "2.0.4"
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -1845,6 +1845,16 @@
     js-cookie "^3.0.1"
     uuid "^8.3.2"
 
+"@sourcegraph/event-logger@^2.0.3":
+  version "2.0.3"
+  resolved "https://registry.yarnpkg.com/@sourcegraph/event-logger/-/event-logger-2.0.3.tgz#1f9c2c01fe76178b1c2476dd2991ba753f09b1a9"
+  integrity sha512-G7p6rImRE7A/GtEX2MiEqPSGPNYtbhOgPcNbw6pOu8g8onpxQzKiirCwfNtRPLlgoacpC1P3034v2tf8dMrSCA==
+  dependencies:
+    "@sourcegraph/event-logger" "^2.0.2"
```
eseliger, line 1853: huh, why does @sourcegraph/event-logger have a dependency on itself? 🤔

nathan-downs: Should I manually remove that dependency?

nathan-downs: Oh, it's actually pulling that in the package dependency in eventLogger

eseliger: Hmm I don't see 2.0.3 even in that repo 🤔 https://github.com/sourcegraph/eventlogger/blob/main/package.json

nathan-downs: Found the issue, I'm self-referencing the package in the code checked into NPM:
https://www.npmjs.com/package/@sourcegraph/event-logger
Let me get that fixed on NPM and checked into our git repo, THEN I'll update the yarn.lock file here properly.

nathan-downs: @eseliger Fixed this on the [NPM side](https://www.npmjs.com/package/@sourcegraph/event-logger?activeTab=readme) and incremented another patch version (we're at 2.0.4 now). Change should be reflected in https://github.com/sourcegraph/eventlogger/blob/main/package.json now. Re-ran `yarn` to create a new `yarn.lock` file and updated the `package.json` file here to increment the event-logger version.

eseliger (approved):
nice, thanks!

PR #5939, opened 2022-12-22
Cloud landing page

st0nebraker:
### Test


taylorsperry:
Looks good! We still need to sort out which logos we can use for Cloud. Can we remove the logo block for now and add back in with the right copy (depending on the numbers, something like "X+ engineers at more than Y companies use Sourcegraph Cloud to build software you rely on.") @iskyOS did Marketing have any insights on the lists we were cross-referencing? cc @traceyljohnson

marekweb (approved):

PR #5947, opened 2023-01-11
add cody beta page

beyang:
### Test


fabicastp:
Hi @beyang! Here's a Figma mockup by @traceyljohnson https://www.figma.com/file/oQgh5R0iSBG3UII6C8CPJ8/Cody-landing-page?node-id=0%3A1&t=TowxXN4zZ751vDKH-0 - It includes an isometric illustration of the chat, something Tracey has been experimenting with. 

Let us know what you think!

beyang:
@fabicastp I'm hesitant to tie any of the hero graphics to the UI as it exists today, because I expect it will change rapidly.

As far as general style, I think I'd like to feature screenshots or video while keeping the hero pretty simple, as we need to focus on the core product before we invest too much on the brand marketing side of things.

fabicastp:
@beyang Two new iterations are ready for review. V2 (A) has room for two images/videos. V2 (B) for one. 
https://www.figma.com/file/oQgh5R0iSBG3UII6C8CPJ8/Cody-landing-page?node-id=20%3A1357&t=Me5e4sINjidOEL1l-0

beyang:
Let's revisit this after the initial release? V2(A) looks good, but I don't think it is a critical improvement that should block getting the first version of the page out.

fabicastp:
@beyang Agreed! Then Tracey recommends these small improvements that will go a long way in making the existing file look tidier:
- Use the "BETA" badge next to the CODY overline. It should already exist in the components.
- Use a cleaner screenshot that has the drop shadow (.tw-drop-shadow-xl) we typically use across the site.
- Set a max-width of 768px for the paragraph under "Codebase-aware chat"
- Use our current gradient radial-gradient(109.27% 56.82% at 93.02% 27.51%, #6112A3 0.41%, #270741 54.04%, #1E0632 94.78%) for brand consistency/recognition.

rrhyne:
@beyang after looking at the revised design, I think we are missing an easily scannable way for developers to see the power of the solution. The graphics showing questions asked and the Cody's response served that need. 

I think this internal [slack thread](https://sourcegraph.slack.com/archives/C049L7Q3Z0V/p1673541714472559) with the overwhelmingly positive responses based on actual examples proves that out really well. No one needed to watch a video to get stoked, but they'd watch one now! 

Additionally, from experience, I can say that it takes quite a bit for a busy person to commit to playing a video. For instance, less than 5% of visitors to sourcegraph.com watched our video about search.

My suggestion would be to put the question and answer back in with a compelling Q&A you select that is targeted squarely at our audience. I think you'll get higher conversions as a result.

PR #5952, opened 2023-01-15
switch from yarn to pnpm

sqs:
We are now using pnpm in the main repository (https://github.com/sourcegraph/sourcegraph). Let's switch to pnpm here, too, for consistency.

Also adds deps in package.json that were imported from but not mentioned. pnpm (correctly) flagged these.


vovakulikov (dismissed):
Nice, thanks, Sidenote: I think PR title (and therefore squashed commit message) should be "switch from yarn to pnpm"

vovakulikov:
Oh just noticed that E2E testes are failing on CI. I think this might be some sort of problem similar to what we have with esbuild in main repo

sqs:
@vovakulikov Thanks for the catch on the wrong commit message. And the e2e tests are a different issue (node17+ https://github.com/jeffbski/wait-on/issues/109#issue-1071849300). Should be fixed. (Unless you are referring to a different esbuild-related issue.)

PR #5955, opened 2023-01-18
Update youtube video used as About homepage demo

iskyOS:
Update the YouTube video ID to reflect the new demo video.


taylorsperry (approved):
👍

PR #5978, opened 2023-01-29
Fix CTA3 colours issue

### LINKED ISSUE #5977: CTA 3 in dark mode pages isn't visible
image

image

https://about.sourcegraph.com/case-studies/cern-reduces-technical-debt

omar2205:
This fixes #5977
image

adds a check `const isDark = document.querySelector('footer.navbar-dark')`


omar2205:
It's failing because `document is not defined`. :/

omar2205:
This works when navigating but doesn't work if visiting the URL directly. Using a prop would be the better solution, but that would require changing every `<CtaSection />`.

PR #5984, opened 2023-02-02
remove triblio

sqs:
Remove Triblio scripts. Triblio is no longer being used (per [Slack message](https://sourcegraph.slack.com/archives/C03B4EMBXL4/p1672838415721909?thread_ts=1672837088.069059&cid=C03B4EMBXL4)).


sqs:
We are still using Triblio.

PR #5992, opened 2023-02-10
New homepage design ("brand-aid")

sqs:
Implements the new homepage design at https://www.figma.com/file/fudljjrGrUg5cDYTAo4Hra/Homepage?node-id=1691%3A5613&t=OPHDtHk7314EoJsB-0.


sqs:
@traceyljohnson I think I need to get the gradients a bit more precisely correct from Figma, but otherwise it looks great and works on mobile and desktop. I'm going to merge it. (Can always revert/fix if anything is amiss.)

traceyljohnson:
- Rogue strip of purple just below the logo pond. Not sure what that is, but should be removed.
- The “Try Sourcegraph for free” button should not have a stroke. It also appears to be 58px in height vs the 44px height we use for desktop (mobile is 54px).
- I realized I didn’t provide you with a hover state for the new button color. [Can we use this for now?](https://www.figma.com/file/o1QRtdQI0ozKq0n7ATrKlx/Marketing-DLS?node-id=2542%3A21014&t=walG1YqtelQ607V7-1)
- Can you add the 8px corner radius (.rounded-lg) to the video? It just makes it feel a bit more polished.

PR #5996, opened 2023-02-14
blog post: Planes aren't faster cars

sqs:


fabicastp:
@iskyOS assigned to you!

content/blogposts/2023/planes-are-not-faster-cars.md
```diff
@@ -0,0 +1,108 @@
...
+2022-11-09 12:33:07.18994 -0800
+2022-11-21 02:19:47.189964 -0800
+```
+
+
+I thought of a question I could answer with code, wrote the code, got the answer — in just a few _seconds_.
+
+How much time did I save compared to writing this in C? It’s irrelevant: I would never have bothered writing something so trivial in C. The magic of Ruby wasn’t in the time saved. The magic was in the parts of my world that opened up. Suddenly, I could go from question-to-code-to-answer in just a few seconds!
+
+Both for the very large things (travel to distance cities) and for the very small things (simple questions I have where the setup cost just wasn’t going to be worth it), the principle is the same:
```
iskyOS, line 86: Minor edit, I think this is meant to be "distant" rather than "distance":
```suggestion
Both for the very large things (travel to distant cities) and for the very small things (simple questions I have where the setup cost just wasn’t going to be worth it), the principle is the same:
```

iskyOS:
@chrispine, I took another read. I really enjoy this post! I love how much personality comes through in your writing style. 

I found what I think was one minor typo, but I don't have any directional feedback or suggestions for changes. The content is great and the length feels just right. I'm happy to publish this when you're good with it!

chrispine:
Cool, thanks! Yes, I'm happy for you to ship it whenever :-)

iskyOS:
Thanks @chrispine! In that case I am going to go ahead and ship

PR #6060, opened 2023-03-09
no shadow on figures by default

sqs:
Most static `<Figure>`s shouldn't have a drop shadow. Fixes an issue where the shadow was being applied to some figures where it did not look good:

image

For some reason, this is showing up differently in local dev vs. prod.

Pushing to check deploy preview.


rrhyne (approved):
The deploy preview looks good to me.

PR #6061, opened 2023-03-09
Add localhost:3080

taylorsperry:
I'm wondering if it feels like we're missing a final step here, so I've added an indication that the browser will open automatically.


taylorsperry:
Got @iskyOS 's OK on this in Slack, so I'm merging!

PR #6066, opened 2023-03-10
Add eventLogger to track download link clicks and code copy events

marekweb:
Add events (logged via EventLogger) to help track downloads of Sourcegraph App.

**What:**

- Event: `DownloadClick` when a download link is clicked. Used for Sourcegraph App downloads.
  - arguments:
    - `downloadSource`: `about` (to differentiate from `DownloadClick` on Docs site, which has the value `docs`)
    - `downloadName`: example value: `app-download-linux-deb`
    - `downloadLinkUrl`: url of the clicked link
- Event: `CodeSnippetCopied` when a code snippet's copy code button is clicked.
  - arguments: 
    - `codeSnippetName`: example value: `app-brew-install`

**How:**

`<DownloadLink>` is a wrapper around `<Link>` which logs `DownloadClick` and uses the `downloadName` prop as an identifier.

`<CodeSnippet>` now takes an optional `snippetName` prop and logs the copy event if `snippetName` is present.

**Where**:

On the `/app` landing page for Sourcegraph App

**Related**:

The `DownloadClick` event shares the same schema as the download links in the docs:
 
- https://github.com/sourcegraph/sourcegraph/pull/49141


src/hooks/eventLogger.ts
```diff
@@ -9,13 +9,20 @@ interface EventArguments {
...
         eventLogger = new EventLogger('https://sourcegraph.com')
     }
     return eventLogger
 }
 
+export const logCodeSnippetCopied = (snippetName: string = ''): void => {
+    const eventArguments = { snippetName }
+
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
+    getEventLogger()?.log('CodeSnippetCopied', eventArguments, eventArguments)
```
philipp-spiess, line 23:
```suggestion
    getEventLogger().log('CodeSnippetCopied', eventArguments, eventArguments)
```

I don't think this can be null (the signature of `getEventLogger()` is also not nullable)

philipp-spiess (approved):
Noice

src/hooks/eventLogger.ts
marekweb: You're right, I wrote this because the other `getEventLogger` calls used `.?` but it looks like it's unnecessary

PR #6069, opened 2023-03-13
Reduce padding of command box

rafax:
Before: 
Screenshot 2023-03-13 at 11 06 37
After:
Screenshot 2023-03-13 at 11 06 53

### DIFF
--- a/src/pages/app.tsx
+++ b/src/pages/app.tsx
@@ -171,7 +171,7 @@ const PlatformsInstallSnippet: FunctionComponent = () => (
 
                                     <CodeSnippet
                                         code="brew install sourcegraph/sourcegraph-app/sourcegraph"
-                                        className="-ml-2.5"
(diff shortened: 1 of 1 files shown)


src/pages/app.tsx
```diff
@@ -208,7 +208,7 @@ const PlatformsInstallSnippet: FunctionComponent = () => (
                                         and install it:
                                     </div>
 
-                                    <CodeSnippet code="dpkg -i <file.deb>" className="-ml-2.5" />
+                                    <CodeSnippet code="dpkg -i <file.deb>" className="-ml-2.5 pb-3" />
```
danielmarquespt, line 211: From what I can see in the dev tools this is the right way to do it. The code block has a pb-9 which is too much.

danielmarquespt (approved):

unknwon (approved):
Learn a little Tailwind every day 😀

PR #6070, opened 2023-03-13
App: Fix font size of installation instructions

umpox:
## Description

This PR:
1. Updates the font-size of steps 2,3,4 to match step 1 - `text-sm`
2. Updates `CodeSnippet` to remove the `pb-9` that makes the snippet look so big. This was fixed in [this PR](https://github.com/sourcegraph/about/pull/6069) but I noticed that this still doesn't work for me, it seems to take different padding depending on the CSS order. We only use `CodeSnippet` in this one place so we can change this.

Before:

image

After:

image

### DIFF
--- a/src/pages/app.tsx
+++ b/src/pages/app.tsx
@@ -174,7 +174,7 @@ const PlatformsInstallSnippet: FunctionComponent = () => (
                                         className="-ml-2.5"
                                     />
                                 </li>
-                                <li className="ml-2.5 text-sm leading-7">Run `Sourcegraph` in your local directory.</li>
+                                <li className="ml-2.5 text-sm leading-7">Run <code className="bg-gray-200 rounded-md p-1 text-xs text-gray-700">sourcegraph</code> in your local directory.</li>
                                 <li className="ml-2.5 text-sm leading-7">
                                     The app will detect code in the directory you run it in.
                                 </li>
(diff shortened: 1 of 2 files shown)


philipp-spiess (approved):
Nice, ty!

src/pages/app.tsx
```diff
@@ -171,14 +171,14 @@ const PlatformsInstallSnippet: FunctionComponent = () => (
 
                                     <CodeSnippet
                                         code="brew install sourcegraph/sourcegraph-app/sourcegraph"
-                                        className="-ml-2.5 pb-3"
+                                        className="-ml-2.5"
                                     />
                                 </li>
-                                <li className="ml-2.5 text-xs leading-7">Run `Sourcegraph` in your local directory.</li>
-                                <li className="ml-2.5 text-xs leading-7">
+                                <li className="ml-2.5 text-sm leading-7">Run `Sourcegraph` in your local directory.</li>
```
philipp-spiess, line 177: Should this be

```suggestion
                                <li className="ml-2.5 text-sm leading-7">Run <code>sourcegraph</code> in your local directory.</li>
```

(If we have `<code>` styled already, that is).

Should definitely be lowercased, though!

rafax (dismissed):

unknwon (dismissed):
🚀

src/pages/app.tsx
umpox: Not by default, but with a few styles we can get something similar:

`<code className="bg-gray-200 text-xs rounded-md px-2 text-gray-700">sourcegraph</code>`

image

WDYT?

Issue #6204, opened 2023-04-21
Big Code page


fabicastp:
about.sourcegraph.com/big-code/the-rise-of-ai 

[Figma](https://www.figma.com/file/GYzVkbfU4OhYxIAzOyFi9J/Ebook-template?node-id=279-370&t=XKiaq7hlU5SwCqVH-0) --> See webpage page only 

Meta
Copy: Developers are struggling more than ever with issues around tech debt, maintenance, collaboration, and scalability. But with the overnight AI boom, the pain of Big Code is only getting worse. 

Image soon...

fabicastp:
Form Code:
<script charset="utf-8" type="text/javascript" src="//js.hsforms.net/forms/embed/v2.js"></script>
<script>
  hbspt.forms.create({
    region: "na1",
    portalId: "2762526",
    formId: "3612c366-2e75-4a4b-9212-6dcbd6b008fe"
  });
</script>

SF Campaign - [GC - Whitepaper - Big Code in AI Era](https://sourcegraph2020.lightning.force.com/lightning/r/Campaign/7015b000005VLpvAAG/view)

File - https://2762526.fs1.hubspotusercontent-na1.net/hubfs/2762526/Big%20Code%20in%20the%20AI%20era%20report.pdf

PR #6217, opened 2023-04-25
SGM-6204 - Big Code page

gitstart-sourcegraph:
## Description
Design Big Code Page

## Ref
[SG Issue](https://github.com/sourcegraph/about/issues/6204)
[Gitstart ticket](https://clients.gitstart.com/sourcegraph/1/tickets/SGM-6204)

## Demo
https://user-images.githubusercontent.com/51731962/234070334-6af6104a-6b5c-4270-89d9-c4540691756b.mov


traceyljohnson (changes requested):
- Can we keep the hero illustration cropped and pinned to the right at all screen sizes? The right edge doesn't look right on wide screens, so I'd rather keep it cropped and always on the edge. 
- The illustration at the bottom is doing a weird cropping behavior at different screen sizes. Can we make sure it adjusts smoothly?

fabicastp (approved):

PR #6226, opened 2023-05-01
Fix use/challenge/solution/results component

SuperAuguste:
I was showing how awesome Sourcegraph is to some friends in industry when I ran into this:

Screenshot 2023-05-01 at 10 47 23 AM

I decided to fix this accordingly:

Screenshot 2023-05-01 at 10 47 45 AM
Screenshot 2023-05-01 at 10 48 04 AM

I imagine that you'll eventually want to phase this component out completely and write up a nice report for the last three pages that use it, but I think this should be a quick and easy solution for the time being :)


sqs:
Thank you!

PR #6267, opened 2023-05-19
Log video play events

dadlerj:


src/pages/own.tsx
```diff
@@ -107,6 +108,7 @@ const OwnPage: FunctionComponent = () => (
                     playsInline={true}
                     controls={false}
                     data-cookieconsent="ignore"
+                    onPlay={() => getEventLogger().log('StaticVideoPlayed', {title: 'Own landing page video 3'}, {title: 'Own landing page video 3'})}
```
nathan-downs, line 111: Is there any way to extract the `own-landing-page-3` part out rather than having the video number hard coded?

dadlerj: Honestly what I really want is an actual human-understandable title for the video. I'll just go check out the page and see what the video is!

nathan-downs (dismissed):
Looks good to me

src/pages/own.tsx
dadlerj: Updated, thanks

PR #6262, opened 2023-05-18
[SGM-51897]: Improve frictionless sign-up flow by skipping steps

gitstart-sourcegraph:
## Description
Added frictionless sign-up flow

## Ref
[SG Issue](https://github.com/sourcegraph/sourcegraph/issues/51897)
[Gitstart ticket](https://clients.gitstart.com/sourcegraph/1/tickets/SGM-51897)

## Demo

https://github.com/sourcegraph/about/assets/89894075/4806298b-bcfe-4383-8180-264fa65afa32

PR #6272, opened 2023-05-23
Fix inconsistencies in Auth Modal Visibility

gitstart-sourcegraph:
## Description
When we click on the modal button, nav button is being clicked for the header CTA. This PR fixes the button clicked on the modal on small screen for header section.

## Ref
https://github.com/sourcegraph/about/pull/6262

## Demo video

https://github.com/sourcegraph/about/assets/89894075/68bcb69a-07b9-4376-8466-b39e007be3c7

### DIFF
--- a/src/components/Layout/Header/Header.tsx
+++ b/src/components/Layout/Header/Header.tsx
@@ -164,9 +164,6 @@ const HeaderContent: FunctionComponent<Props & { open: boolean; sticky: boolean;
             >
                 Get Cody for free
             </button>
-            <Modal open={isSignUpModalOpen} handleClose={closeModal}>
-                <AuthenticateModalContent source={source} />
-            </Modal>
         </>
     )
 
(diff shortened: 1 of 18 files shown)


src/components/Layout/Header/Header.tsx
```diff
@@ -261,6 +258,9 @@ const HeaderContent: FunctionComponent<Props & { open: boolean; sticky: boolean;
                     <div className="ml-3 !mt-2 flex flex-col items-start space-y-4">{callToAction}</div>
                 </div>
             </Disclosure.Panel>
+            <Modal open={isSignUpModalOpen} handleClose={closeModal}>
```
dadlerj, line 261: To confirm, this is the only meaningful change, right? The rest of this PR is just linting?

dadlerj (approved):

src/components/Layout/Header/Header.tsx
dadlerj: If so, feel free to merge, this makes sense

PR #6290, opened 2023-06-05
Update /app page

taylorsperry:
Though we want to wait until June 28 to do the big marketing push for the Cody app, we're now at a point where the new Tauri version is substantially better than the version we launched at Starship. Anyone who stumbles upon the app on our marketing site should download the new version instead of the old one. 

This is a quick fix to update the download options and is *not* meant to reflect our ideal marketing positioning, which is in flight but shouldn't block this change. I'll follow up with a quick demo video to add to `/app`. 

This PR: 
- Removes the "download the app" CTA from about.sourcegraph.com homepage
- Updates about.sourcegraph.com/app to include the new download link, reorient copy around the Cody app, and remove the marketing material that was oriented around the old Sourcegraph app
- Updates about.sourcegraph.com/get-started to include the new download link and Cody-specific copy

# OLD
Screenshot 2023-06-05 at 12 36 51 PM
Screenshot 2023-06-05 at 12 37 51 PM
Screenshot 2023-06-05 at 12 38 39 PM

# NEW
Screenshot 2023-06-05 at 12 40 45 PM
Screenshot 2023-06-05 at 12 40 59 PM

### DIFF
--- a/src/components/cta/ExternalsAuth.tsx
+++ b/src/components/cta/ExternalsAuth.tsx
@@ -67,7 +67,7 @@ export const ExternalsAuth: React.FunctionComponent<ExternalsAuthProps> = ({
         <Link
             href="https://sourcegraph.com/.auth/github/login?pc=https%3A%2F%2Fgithub.com%2F%3A%3Ae917b2b7fa9040e1edd4"
             className={classNames(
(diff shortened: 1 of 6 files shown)


marekweb (approved):
Looks good!

PR #6370, opened 2023-06-29
Email redirect parameter should be returnTo, not redirect

pjlast:
Closes https://github.com/sourcegraph/sourcegraph/issues/54402

The email signup does not go through the oauth flow, so the redirect parameter should be `returnTo`, not `redirect`


sashaostrikov (approved):

indradhanush (approved):
🚢

rrhyne:
🤦 Thank you!

PR #6371, opened 2023-06-29
feat(cody-survey-toast): set Cody survey state on external auth signup

erzhtor:
Related to https://github.com/sourcegraph/sourcegraph/pull/54422.

When the user signs up via external auth, it skips the dotcom cloud signup page, where flags are set to show Cody survey for new users. This PR adds this flag-setting logic directly to the about site.

## Test plan
- check both PRs: this and https://github.com/sourcegraph/sourcegraph/pull/54422.


vovakulikov (approved):

rrhyne (approved):
LGTM

PR #6384, opened 2023-07-03
blogpost: SCIP-powered Cody

efritz:


content/blogposts/2023/blah.md
```diff
@@ -0,0 +1,138 @@
...
+
+(TODO - introduce the following examples section)
+
+### Risperidone
+
+**As it turns out, the greatest LLM anti-psychotic is giving it enough context in the first place.**
```
mrnugget, line 34: Nitpick at this stage, but I think "enough of _the right context_" is the thing here :)

content/blogposts/2023/blah.md
```diff
@@ -0,0 +1,138 @@
...
+        caption="SCIP-powered Cody giving a precise answer given appropriate context." />
+</a>
+
+Note that we've stopped additional fields from being emitted, but we didn't happen to expand all of the fields that are present. Currently we're only performing single-edge traversals from the source, but plan to extend the search in the future. Including a second-order relationship included in this example may have prompted Cody to expand the fields present in the embedded struct `CompletionRequestParameters`.
+
+### In the land of infinite possibilities...
+
+**Generative AI gonna generate.**
+
+Too many answers are valid when too little context is given to restrict the solution space. This is a special form of garbage-in, garbage-out. In the following example, we ask Cody to describe the output of a particular function when executed. We choose this particular example in which the input question has little to no similarity to the _internal implementation_ of the function to ensure embeddings search alone would not return all of the relevant context.
```
mrnugget, line 72: > Too many answers are valid when too little context is given to restrict the solution space.


Very good!! Maybe you could even add example of daily-life-with-chat-gpt, where folks also have learned to add qualifying statements like "assume I'm a software engineer" or "assume you don't need to explain to me what HTTP is" to their prompts.

mrnugget (commented):
very cool so far!

content/blogposts/2023/blah.md
```diff
@@ -0,0 +1,151 @@
...
+
+In contrast, our question is asking for a _highly precise_ answer related to a specific piece of code. Searching the repository's embeddings for _similar_ code is not the context needed by the LLM to make the correct inferrences. For such implementation-specific questions, we need to be able to explore a different relationship within the code graph.
+
+For the same question, we can generate a very good response from Cody by supplying additional code we find traversing the _definition_ edges of the semantically-indexed portion of the code graph. To start the search, we extract symbol names from the code that is currently in the view of the user, such as the current workspace, the user's text selection, open files, etc. These symbol names are translatted into a location in our SCIP-backed code graph. Then, particular edges are followed out to explore the graph. As we traverse the graph in this manner, we recall code that is semantically related to the code in question.
```
ErikaRS, line 68: "we can generate" reads a bit confusingly to me. Maybe "we can improve the response from Cody..."

content/blogposts/2023/blah.md
```diff
@@ -0,0 +1,151 @@
...
... invocation. From here, the code intelligence context service can perform a cheap syntactic analysis (via TreeSitter) to extract symbol names from the currently visible code. These symbol names can be translated into SCIP names, giving us a set of root symbols to begin a traversal of relationships. The API gathers relevant code metadata, locations, and source text, and returns it to the Cody client for inclusion in the context window. Following different types of relationship edges will have different effects based on the user’s intent. If a user is looking to explain or generate a unit test for an arbitrary piece of code, it’s intuitive that exposing the implementation of invoked functions will be beneficial. If a user is generating a user form that populates a given type, it’s similarly intuitive that exposing the definitions and documentation of each of the type’s properties would prevent a response with hallucinated or mis-typed form fields. Supposing that the Sourcegraph Enterprise has also indexed package hosts, the global SCIP index will also include all relevant downstream dependencies. This allows us to traverse very deeply into the symbol graph to gather relevant context.
```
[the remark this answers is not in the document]
ErikaRS: Actually, you include some of this explanation of how this works below, so perhaps you could strike this paragraph out completely and work bits of it into the rest. Alternately, you could completely remove the "how it works" detail from the examples and keep it in its own section (which I still personally think would go better after the examples). 

But having just one or the other will be good.

content/blogposts/2023/blah.md
```diff
@@ -0,0 +1,151 @@
...
+
+For the same question, we can generate a very good response from Cody by supplying additional code we find traversing the _definition_ edges of the semantically-indexed portion of the code graph. To start the search, we extract symbol names from the code that is currently in the view of the user, such as the current workspace, the user's text selection, open files, etc. These symbol names are translatted into a location in our SCIP-backed code graph. Then, particular edges are followed out to explore the graph. As we traverse the graph in this manner, we recall code that is semantically related to the code in question.
+
+We've already seen Cody infer infer the type of the variable in question. Including the definition (hence the concrete field names) in the LLM context window provides enough information to effectively prevent the spew of probable-but-fallacious field names we got when it had no concrete examples in its working memory.
```
ErikaRS, line 70: I think we can get rid of the first sentence of this paragraph and jump straight to the payload: including the definition makes this better.

content/blogposts/2023/blah.md
```diff
@@ -0,0 +1,151 @@
...
+
+<a href="https://storage.googleapis.com/sourcegraph-assets/blog/scip-powered-cody/req-scip2.png" target="_blank">
+    <Figure
+        src="https://storage.googleapis.com/sourcegraph-assets/blog/scip-powered-cody/req-scip2.png"
+        alt="SCIP-powered Cody giving a precise answer given appropriate context."
+        caption="SCIP-powered Cody giving a precise answer given appropriate context." />
+</a>
+
+Note that we've stopped additional fields from being emitted, but we didn't happen to expand all of the fields that are present. Currently we're only performing single-edge traversals from the source, but plan to extend the search in the future. Including a second-order relationship included in this example may have prompted Cody to expand the fields present in the embedded struct `CompletionRequestParameters`.
```
ErikaRS, line 79: This feels a little distracting coming after the image. Consider moving it before the image. Even better if you can compress it down and make it a parenthetical on the previous paragraph. E.g., "(Cody's response still doesn't expand all of the fields that are present. We plan to add additional relationships that will further improve the response.)"

content/blogposts/2023/blah.md
```diff
@@ -0,0 +1,151 @@
...
+
+### In the land of infinite possibilities...
+
+**Generative AI gonna generate.**
```
ErikaRS, line 83: I like the point this section makes about context narrowing possibilities. Would it be feasible to try to make that point with just one example rather than having two reasonably similar examples? 

A possible flow could be: 

- Briefly make your way through
  - Generative AI gonna generate
  - Context helps narrow the scope
  - Embedding based context is good for high level similarity
- Then go more in depth on
  - SCIP based edges make better as demonstrated through an example
  - Explanation of how it works

There are many other structures you can use, but I think the key takeaway is that you can use less detail on the parts that build up the "reveal" and more of the detail on why what we are doing is helpful.

content/blogposts/2023/blah.md
```diff
@@ -0,0 +1,151 @@
...
+        src="https://storage.googleapis.com/sourcegraph-assets/blog/scip-powered-cody/cursor-scip.png"
+        alt="SCIP-powered Cody inferring the correct struct tags."
+        caption="SCIP-powered Cody inferring the correct struct tags." />
+</a>
+
+The skeptical reader may have noticed a heavy similarity between the first two examples - both questions are basically using an AI layer to emulate a fairly simple jump to definition action. Asking these questions in natural language probably takes more time than hovering over the symbol to peek at its definition. These examples are _highly illustrative_ but also very myopic with respect to what's possible with this technique.
+
+Continue reading to see that precise code intelligence isn't a one-trick-pony source of context.
+
+### Deeplying understanding code
```
ErikaRS, line 113: I really like this example. It is, as you note, simplified. However, it also illustrates the value well.

content/blogposts/2023/blah.md
```diff
@@ -0,0 +1,151 @@
...
+
+(ATTEMPT AT REWRITE:)
+Think of embeddings-powered search as an excellent tool for addressing broad, high-level inquiries. For instance, when you pose a question about authentication, it naturally seeks to incorporate pertinent documents related to this topic. However, when the query shifts to implementation specifics of a certain code, the expansive, overarching 10,000 foot view perspective isn't as effective. That's where our vast repository of compiler-accurate Source Code Indexes (SCIP) comes into play. SCIP allows us to navigate the semantic relationships between different code snippets, including the most relevant ones within the context window. As a result, we gain a detail-oriented source view of the same data, enabling us to provide more granular context with precision.
```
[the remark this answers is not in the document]
efritz: I believe this is could also be ripe for illustration >_>

efritz:
@jdorfman  😠

jdorfman:
@efritz it's delete week! =p

PR #6402, opened 2023-07-11
fix metrics

rrhyne:
A few IDE download metrics were not logged correctly. This change corrects that.


src/pages/cody.tsx
```diff
@@ -122,6 +122,16 @@ const CodyPage: FunctionComponent<CodyProps> = ({ tweets }) => {
         openModal('about-cody-deployment')
     }
 
+    const handleOnClick = (eventName: string, type: string): void => {
+        const eventArguments = {
+            source: 'cody',
+            description: '',
```
dadlerj, line 128: if this is empty, you can just delete this

```suggestion
```

PR #6477, opened 2023-09-06
vscode: update custom command config changes

kalanchan:
Received some feedback that the new custom commands config changes were a bit confusing, these changes to the blog should help address that


toolmantim (approved):
Thanks @kalanchan!

PR #6595, opened 2023-11-27
Blog: add post about SIMD dot product

camdencheek:
---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1205858070858945


BolajiOlajide (dismissed):

jdorfman (dismissed):
All links work, OG image is uploaded. Lets 🚢

content/blogposts/2023/slow-to-simd.md
```diff
@@ -0,0 +1,411 @@
...
+        {name: "DotNaive", value: 941900},
+        {name: "DotUnroll4", value: 1286000},
+    ]}
+/>
+
+Note that we can actually improve this slightly more by twiddling with the number of iterations we unroll. On the benchmark machine, 8 seemed to be optimal, but on my laptop, 4 performs best. However, the improvement is quite platform dependent and fairly minimal, so for the rest of the post, I'll stick with an unroll depth of 4 for readability.
+
+## Bounds-checking elimination
+
+In order to keep out-of-bounds slice accesses from being [a security vulnerability](https://en.wikipedia.org/wiki/Heartbleed), the go compiler inserts checks before each read. You can check it out in the [generated assembly](https://go.godbolt.org/z/qT3M7nPGf) (look for `runtime.panic`).
```
ggilmore, line 79: I don't believe it is correct to assert that out-of-bounds checking is purely a response to Heartbleed. As example, [Pascal featured bounds checking](https://en.wikipedia.org/wiki/Bounds_checking) and the language was developed in 1970 - before the Heartbleed discovery. 

However, if you'd like to you can use Hartbleed as motivation for why bounds checking is important.

content/blogposts/2023/slow-to-simd.md
```diff
@@ -0,0 +1,411 @@
...
+            panic("out of bounds")
+        }
+		s3 := a[i+3] * b[i+3]
+		sum += s0 + s1 + s2 + s3
+	}
+	return sum
+}
+```
+
+In a hot loop like this, even with modern branch prediction, the additional branches per iteration can add up to a pretty significant performance penalty. This is especially true in our case because the inserted jumps limit how much we can take advantage of pipelining.
```
ggilmore, line 109: I think it could be useful with a drawing / data diagram to show _why_  a branch throws a wrench into pipe lining.

content/blogposts/2023/slow-to-simd.md
```diff
@@ -0,0 +1,411 @@
...
+
+If we can convince the compiler that these reads can never be out of bounds, it won't insert these runtime checks. This technique is known as "bounds-checking elimination", and the same pattern can be applied to many different memory-safe compiled languages.
```
ggilmore, line 111: To be pedantic (sorry),  bounds checking also occurs in interpreted languages like Python.

content/blogposts/2023/slow-to-simd.md
```diff
@@ -0,0 +1,411 @@
...
+
+## Quantization
+
+We've improved execution speed of our code pretty dramatically at this point, but there is another dimension of performance that is relevant here: memory usage.
+
+(If you skipped the background section, this part might not make much sense.)
+
+In our situation, we're searching over vectors with 1536 dimensions. With 4-byte elements, this comes out to 6KiB per vector, and we generate roughly a million vectors per GiB of code. That adds up.
+
+When searching the vectors, they need to be held in RAM, which puts some serious memory pressure on our deployments. Moving the vectors out of memory would mean loading them from disk at search time, which is a no-go when performance is so critical.
```
ggilmore, line 161: I think it's worth motivating why disks are so much slower in memory in this case. Maybe talk about https://colin-scott.github.io/personal_website/research/interactive_latency.html (or an example based off of this). 

( Should mention that we're doing a lot of random reads...)

content/blogposts/2023/slow-to-simd.md
```diff
@@ -0,0 +1,411 @@
...
+		s1 := int32(aTmp[1]) * int32(bTmp[1])
+		s2 := int32(aTmp[2]) * int32(bTmp[2])
+		s3 := int32(aTmp[3]) * int32(bTmp[3])
+		sum += s0 + s1 + s2 + s3
+	}
+	return sum
+}
+```
+
+This change yields a 4x reduction in memory usage at the cost of some recall precision (which we carefully measured, but is irrelevant to this blog post).
```
ggilmore, line 187: Did you define what _recall_ means yet?

content/blogposts/2023/slow-to-simd.md
```diff
@@ -0,0 +1,411 @@
...
+
+Re-running the benchmarks shows we suffer a perf hit from this change. Taking a look at the generated assembly (with `go tool compile -S`), there are a bunch of new instructions to convert `int8` to `int32`, so I expect that's the source of the slowdown. We'll make up for that in the next section.
```
ggilmore, line 189: What are these new instructions? How do you know that they're slow? Can you provide an example showing that the conversion is the source of the additoonal cost (compare pure int8 v pure int32 v (int8 -> int32) converted versions of the code

content/blogposts/2023/slow-to-simd.md
```diff
@@ -0,0 +1,411 @@
...
+        {name: "DotInt8BCE", value: 1240000},
+    ]}
+/>
+
+
+## SIMD
+
+I always love an excuse to play with SIMD. And this problem seemed like the perfect nail for that hammer.
+
+For those unfamiliar, SIMD stands for "Single Instruction Multiple Data". Just like it's says, it lets you run an operation over a bunch of pieces of data with a single instruction. This is exactly what we want for calculating a dot product.
```
ggilmore, line 205: Why is the SIMD function faster than the above? Can you motivate this a bit?

See "Web Aside OPT:SIMD" in CS:APP

e.g. If we look at agner fog (https://agner.org/optimize/instruction_tables.pdf) we know SIMD takes XXX cycles, while normal version using XXX cycles + parallelism is greater with SMID, etc.

content/blogposts/2023/slow-to-simd.md
```diff
@@ -0,0 +1,411 @@
...
+
+Now, it wasn't all sunshine and rainbows. Hand-writing assembly in Go is weird. It uses a [custom assembler](https://go.dev/doc/asm), which means that its assembly language looks just-different-enough-to-be-confusing compared to the assembly snippets you usually find online. It has some weird quirks like [changing the order of instruction operands](https://www.quasilyte.dev/blog/post/go-asm-complementary-reference/#operands-order) or [using different names for instructions](https://www.quasilyte.dev/blog/post/go-asm-complementary-reference/#mnemonics). Some instructions don't even _have_ names in the go assembler and can only be used via their [binary encoding](https://go.dev/doc/asm#unsupported_opcodes). Shameless plug: I found sourcegraph.com invaluable for finding examples of Go assembly to draw from.
+
+That said, compared to Cgo, there are some nice benefits. Debugging still works well, the assembly can be stepped through, and registers can be inspected. There are no extra build steps (a C toolchain doesn't need to be set up). It's easy to set up a pure-Go fallback so cross-compilation still works. Common problems are caught by `go vet`.
```
ggilmore, line 298:
```suggestion
That said, compared to Cgo, there are some nice benefits. Debugging still works well, the assembly can be stepped through, and registers can be inspected using `delve`. There are no extra build steps (a C toolchain doesn't need to be set up). It's easy to set up a pure-Go fallback so cross-compilation still works. Common problems are caught by `go vet`.
```

content/blogposts/2023/slow-to-simd.md
```diff
@@ -0,0 +1,411 @@
...
+
+## SIMD...but bigger
+
+Previously, we limited ourselves to AVX2, but what if we _didn't_? The VNNI extension to AVX-512 added the [`VPDPBUSD`](https://www.felixcloutier.com/x86/vpdpbusd) instruction, which computes the dot product on `int8` vectors rather than `int16`s. This means we can process four times as many elements in a single instruction because we don't have to convert to `int16` first and our vector width doubles with AVX-512!
+
+The only problem is that the instruction requires one vector to be signed bytes, and the other to be _unsigned_ bytes. Both of our vectors are signed. We can employ [a trick from Intel's developer guide](https://www.intel.com/content/www/us/en/docs/onednn/developer-guide-reference/2023-0/nuances-of-int8-computations.html#DOXID-DEV-GUIDE-INT8-COMPUTATIONS-1DG-I8-COMP-S12) to help us out. Basically, add 128 to one of our vectors to ensure it's in range of an unsigned integer, then keep track of how much overshoot we need to correct for at the end.
```
ggilmore, line 304: > Basically, add 128 to one of our vectors to ensure it's in range of an unsigned integer, then keep track of how much overshoot we need to correct for at the end

Show an example here

ggilmore (approved):
First pass

content/blogposts/2023/slow-to-simd.md
camdencheek: > use Hartbleed as motivation for why bounds checking is important

Oh, that was the intent. I'll reword that

jtibshirani (commented):
This was a great read! I think it already works well, but sharing some high level comments as "food for thought":
* The progression of loop unrolling -> eliminating bounds checking -> quantization -> SIMD felt a little confusing... first it's incremental additive improvements, then we regress with quantization, then SIMD makes the previous improvements irrelevant. As a tweak, you could present it more in a temporal/ historical way, like ... first you made a round of low-hanging fruit improvements using classic techniques. Then a customer came to us with a massive monorepo and we realized memory was a bottleneck! We needed to implement quantization, but this set us back... so then you brought out the BIG hammer and fearlessly dove into assembly.
* People might be thinking "why did they do all this work instead of just using a vector DB"? I wonder if you could add short a paragraph under "background" of what a vector index is, and how there are trade-offs in using it (much higher index latency, higher memory usage, and it's inherently an approximate algorithm). And that for this reason, we decided to invest a couple weeks in pushing our simple brute force implementation further. And ... that impressively, we achieved a competitive search latency with many vector DBs for the result sets we were targeting (1-10 million vectors).

jdorfman:
@camdencheek lets :shipit: !

content/blogposts/2023/slow-to-simd.md
```diff
@@ -0,0 +1,415 @@
...
+---
+
+So, there's this function. It's called a lot. More importantly, all those calls are on the critical path of a key user interaction.
+Let's talk about making it fast.
+
+Spoiler: it's a dot product.
+
+## Some background (or [skip to the juicy stuff](#the-target))
+
+At Sourcegraph, we're working on a Code AI tool named [Cody](https://sourcegraph.com/cody). In order for Cody to answer questions well, we need to give it (him?) enough [context](https://about.sourcegraph.com/blog/cheating-is-all-you-need) to work with. One of the [ways we do this](https://about.sourcegraph.com/whitepaper/cody-context-architecture.pdf) is by leveraging [embeddings](https://platform.openai.com/docs/guides/embeddings).
```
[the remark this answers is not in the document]
camdencheek: Great suggestion, thanks!

content/blogposts/2023/slow-to-simd.md
```diff
@@ -0,0 +1,415 @@
...
+
+The piece relevant to this blog post is that similarity metric, which is just a function that spits out a number that's higher if vectors are more similar. For similarity search, a common metric is [cosine similarity](https://en.wikipedia.org/wiki/Cosine_similarity). However, for normalized vectors (vectors with unit magnitude), cosine similarity yields an [equivalent ranking](https://milvus.io/docs/v1.1.1/metric.md) to the [dot product](https://en.wikipedia.org/wiki/Dot_product). We do not yet build an index over our embeddings, so to run a search, we need to calculate the dot product for every embedding in our data set and keep the top results. And since we cannot start execution of the LLM until we get the necessary context, optimizing this step is crucial.
```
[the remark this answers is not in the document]
camdencheek: Oh, good catch. I copied in the wrong link from my notes 🤦 I've updated the link to [Google's dev education resources](https://developers.google.com/machine-learning/clustering/similarity/measuring-similarity) which includes a small proof of the equivalence.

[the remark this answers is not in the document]
camdencheek: RE: writing out the math, it does not look like we have math rendering set up for the site. I think that link does a good job of covering it, so I probably won't spend much time trying to get that working

content/blogposts/2023/slow-to-simd.md
```diff
@@ -0,0 +1,415 @@
...
+
+## The target
+
+This is a simple implementation of a function that calculates the dot product of two vectors. My goal is to outline the journey I took to optimize this function, and to share some tools I picked up along the way.
```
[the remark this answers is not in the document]
camdencheek: Good call 👍 The audience is definitely primarily go developers (though some of these techniques apply across languages). I've updated the title to mention go explicitly, and also mentioned go in this paragraph

content/blogposts/2023/slow-to-simd.md
camdencheek: Went ahead and removed the confusing reference to compiled languages, and just rephrased to link to BCE recipes in another language.

content/blogposts/2023/slow-to-simd.md
camdencheek: Oh, that's a great link! I'll incorporate that

content/blogposts/2023/slow-to-simd.md
camdencheek: TBH, I don't know they're to blame, just a hunch. I didn't look too closely at this because I jumped into the SIMD stuff straight from here. I've updated the language here to soften my claim and shuffle people into the next section

camdencheek:
> The progression of loop unrolling -> eliminating bounds checking -> quantization -> SIMD felt a little confusing

Yeah, agreed. I clarified some of the transitions around the quantization step and took your advice on presenting it more from an "our motivation" perspective than a progression perspective. 

>  I wonder if you could add short a paragraph under "background" of what a vector index is

Done!

content/blogposts/2023/slow-to-simd.md
camdencheek: done!

content/blogposts/2023/slow-to-simd.md
camdencheek: I added a little bit of motivation, but I don't think I have the room to get too deep here without burying the lede.

content/blogposts/2023/slow-to-simd.md
camdencheek: I added  alink to [this blog post](https://chadaustin.me/2009/02/latency-vs-throughput/) which I found helpful for going a bit deeper

content/blogposts/2023/slow-to-simd.md
```diff
@@ -0,0 +1,411 @@
...
+}
+```
+
+Unless otherwise stated, all benchmarks are run on an Intel Xeon Platinum 8481C 2.70GHz CPU. This is a `c3-highcpu-44` GCE VM. The code in this blog post can all be found in runnable form [here](https://github.com/camdencheek/simd_blog).
+
+## Loop unrolling
+
+Modern CPUs do this thing called [_instruction pipelining_](https://en.wikipedia.org/wiki/Instruction_pipelining) where it can run multiple instructions simultaneously if it finds no data dependencies between them. A data dependency just means that the input of one instruction depends on the output of another.
+
+In our simple implementation, we have data dependencies between our loop iterations. A couple, in fact. Both `i` and `sum` have a read/write pair each iteration, meaning an iteration cannot start executing until the previous is finished.
```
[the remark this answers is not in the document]
camdencheek: I added a little bit more color, but I don't think I can do a good job explaining this well without basically a whole blog post on its own 😄 If you'd like to take a stab at it, feel free!

jdorfman:
@camdencheek lets update the date and :shipit: =)

content/blogposts/2023/slow-to-simd.md
```diff
@@ -0,0 +1,508 @@
...
+
+This technique translates well to many memory-safe compiled languages like 
+[Rust](https://nnethercote.github.io/perf-book/bounds-checks.html).
+
+Exercise for the reader: why is it significant that we slice like `a[i:i+4:i+4]` rather than just `a[i:i+4]`?
+
+## Quantization
+
+We've improved single-core search throughput by ~50% at this point, but now we've hit a new bottleneck: memory usage.
+Our vectors 1536 dimensions. With 4-byte elements, this comes out to 6KiB per vector, and we generate roughly a million
```
jtibshirani, line 196:
```suggestion
Our vectors are 1536 dimensions. With 4-byte elements, this comes out to 6KiB per vector, and we generate roughly a million
```

content/blogposts/2023/slow-to-simd.md
```diff
@@ -0,0 +1,508 @@
...
+        {name: "DotUnroll4", value: 1286000},
+        {name: "DotBCE", value: 1407000},
+        {name: "DotInt8BCE", value: 1240000},
+    ]}
+/>
+
+
+## SIMD
+
+I always love an excuse to play with SIMD. And this problem seemed like the perfect nail for that hammer.
```
jtibshirani, line 253: I think the transition here could be clearer. Otherwise it's a little confusing how this relates, like is it layering on another improvement, or replacing the previous ones? You mention that the previous improvements become irrelevant, but don't explain why.

Maybe you could present it as: the improvements so far proved too incremental and sensitive to changes in data format. So we looked into a more dramatic optimization.

content/blogposts/2023/slow-to-simd.md
```diff
@@ -0,0 +1,508 @@
...
+
+# What's next?
+
+Well, I'm pretty happy with an 9.3x increase in throughput and a 4x reduction in memory usage, so I'll probably leave it
+here.
+
+The real life answer here is probably "use an index". There is a ton of good work out there focused on making nearest
+neighbor search fast, and there are plenty of batteries-included vector DBs that make it pretty easy to deploy. 
+
+_However_, if you want some fun food for thought, a colleague of built a proof-of-concept [dot product on the
```
jtibshirani, line 494:
```suggestion
_However_, if you want some fun food for thought, a colleague of mine built a proof-of-concept [dot product on the
```

content/blogposts/2023/slow-to-simd.md
```diff
@@ -0,0 +1,508 @@
...
+GPU](https://github.com/sourcegraph/sourcegraph/compare/main...nsc/embeddings-fun#diff-eed4a741ebe632c484dce236a3f4b1eee16e9a2bec6749003b3dbc41449c497c).
```
jtibshirani, line 495: This link feels fragile, maybe he could open it as a draft PR ?

jtibshirani (approved):
Looks great!

content/blogposts/2023/slow-to-simd.md
camdencheek: I've updated the link to use a tag so it should be safe to treat as a permalink

content/blogposts/2023/slow-to-simd.md
camdencheek: Done!

PR #6596, opened 2023-11-28
Phase one

jdorfman:
---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1206004842161819


jdorfman:
@kukicado can we merge this in. @alexAtSourcegraph and I want to start on the others

PR #6612, opened 2023-12-01
Update cloud.md

srmandell:


jdorfman:
@srmandell does this also need merging or is it out of date?

PR #6625, opened 2023-12-08
Design fixes for comparison pages

chris-sev:
To address the following feedback: 

- the gradient under the GitHub logo looks a bit out of place (not sure if it’s just a rendering issue on my browser). Attaching pic. Not sure its needed, anyhow.
- Hero heading should use the global H1 styling. The team is in the process of updating most of our Tailwind typography specs so it would be good to use only the default global styles for now so it will auto-update with the change :slightly_smiling_face:


kukicado:
LGTM!

PR #6687, opened 2024-01-12
fixing typo in subheading of compare page

traceyljohnson:
There is an extra S in the word "assistants" that shouldn't be there. Removed it :)


jdorfman:
Great catch @traceyljohnson !

PR #6714, opened 2024-01-31
Update link to cloud TOS changes

dadlerj:


dadlerj:
This updates the link to https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/sourcegraph/about%24+file:terms/cloud.md+type:diff&patternType=standard&sm=0&groupBy=author, which I believe is what we want here.

PR #6720, opened 2024-02-07
ExternalsAuth: update sign-in URLs to use SAMS

unknwon:
We are going to take down old sign-in URLs in favor of SAMS.

Side note: the Pro plan link for GitLab one never worked... 😂 

We have patched both Cody clients, e.g. https://github.com/sourcegraph/cody/pull/3037


kalanchan (approved):

unknwon:
Merging as-is, happy to address post-merge feedback!

PR #6852, opened 2024-04-08
the-lifecycle-of-a-code-ai-completion: fix multi-line-completions language list

bobheadxi:
per https://sourcegraph.slack.com/archives/C04MSD3DP5L/p1712545179423749 - made a guess at what this might be supposed to point at

Used a permalink - this might become outdated in the future, but at least it won't outright break


bobheadxi:
Merging for now since it's probably a bit better than broken link and the post is already out there

sqs:
Thank you!

PR #6967, opened 2024-05-24
Add v2 telemetry infrastructure to about repo

dadlerj:
Thanks for the guidance on this @bobheadxi at https://sourcegraph.slack.com/archives/C05BGNBEPKL/p1716573233789729

This PR adds a new frontend telemetry service at `src/telemetry/gqlTelemetryExporter.ts` and `src/telemetry/index.ts`. It also duplicates the existing user and session trackers and cookie helpers, adapted to the new service. It also adds a single hook that records telemetry events on every pageview (again, duplicating what the old v1 telemetry was doing). 

Instead of building a pre-rolled `TelemetryExporter` _or_ `TelemetryRecorderProvider`, I decided to just do this all in place to start and to determine, based on the level of overlap I see with other marketing-owned subdomains, whether to eventually pull all of that out into a separate package. I suspect I ultimately will.

I will go through the process of updating all event logging/recording sites to use v2 telemetry once this is merged.

### DIFF
--- a/src/hooks/telemetry.ts
+++ b/src/hooks/telemetry.ts
@@ -1,0 +1,18 @@
+import { useEffect } from 'react';
+
+import { TelemetryRecorder } from '@sourcegraph/telemetry';
+
+export function useRecordPageViews(telemetryRecorder: TelemetryRecorder<'', ''>): void {
+    useEffect(() => {
+        const privateMetadata: { [key: string]: string | null} = {path: location.pathname}
+        const urlSearchParameters = new URLSearchParams(location.search)
+        const utmParameters: string[] = ['utm_source', 'utm_campaign', 'utm_medium', 'utm_term', 'utm_content']
+
+        for (const parameter of utmParameters) {
+            if (urlSearchParameters.has(parameter)) {
+                privateMetadata[parameter] = urlSearchParameters.get(parameter)
+            }
+        }
+        telemetryRecorder.recordEvent('aboutPage', 'view', {privateMetadata})
+    }, [])
+}
--- a/src/pages/_app.tsx
+++ b/src/pages/_app.tsx
@@ -12,6 +12,8 @@ import { PostHogProvider } from 'posthog-js/react'
 import { AuthModalProvider } from '../context/AuthModalContext'
 import { useEventLogger, useLogAllLinkClicks } from '../hooks/eventLogger'
 import { useLandingSource } from '../hooks/landingSource'
+import { useRecordPageViews } from '../hooks/telemetry'
+import { TelemetryRecorderProvider, noOpTelemetryRecorder } from '../telemetry'
 import 'prism-themes/themes/prism-one-light.css'
 
 // Check that PostHog is client-side (used to handle Next.js SSR)
@@ -28,6 +30,13 @@ const App = ({ Component, pageProps }: AppProps): ReactNode => {
     useEventLogger()
     useLandingSource()
     useLogAllLinkClicks()
+
+    let telemetryRecorder = noOpTelemetryRecorder
+    if (typeof window !== 'undefined') {
+      const telemetryRecorderProvider = new TelemetryRecorderProvider()
+      telemetryRecorder = telemetryRecorderProvider.getRecorder()
+    }
+    useRecordPageViews(telemetryRecorder)
 
     const router = useRouter()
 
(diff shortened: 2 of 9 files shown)


src/hooks/telemetry.ts
```diff
@@ -0,0 +1,18 @@
+import { useEffect } from 'react';
+
+import { TelemetryRecorder } from '@sourcegraph/telemetry';
+
+export function useRecordPageViews(telemetryRecorder: TelemetryRecorder<'', ''>): void {
```
dadlerj, line 5: This file duplicates this logic from the hooks/eventLogger.ts file in this repo

src/pages/_app.tsx
```diff
@@ -29,6 +31,13 @@ const App = ({ Component, pageProps }: AppProps): ReactNode => {
     useLandingSource()
     useLogAllLinkClicks()
 
+    let telemetryRecorder = noOpTelemetryRecorder
+    if (typeof window !== 'undefined') {
```
dadlerj, line 35: This is seemingly required to prevent this code from running on the server, a well known workaround for next js: https://stackoverflow.com/questions/55151041/window-is-not-defined-in-next-js-react-app

src/telemetry/cookies.ts
```diff
@@ -0,0 +1,73 @@
+/**
```
dadlerj, line 1: This file duplicates https://sourcegraph.com/github.com/sourcegraph/sourcegraph@90826c6d7370511469e528258ee6ee446f784870/-/blob/client/shared/src/telemetry/web/cookies.ts from the sourcegraph/sourcegraph repo

src/telemetry/sessionTracker.ts
```diff
@@ -0,0 +1,81 @@
+import { type Cookies, defaultCookies } from './cookies'
```
dadlerj, line 1: This file duplicates logic from the sourcegraph/sourcegraph repo: https://sourcegraph.com/github.com/sourcegraph/sourcegraph@90826c6d7370511469e528258ee6ee446f784870/-/blob/client/shared/src/telemetry/web/sessionTracker.ts

src/telemetry/userTracker.ts
```diff
@@ -0,0 +1,109 @@
+import { formatISO, startOfWeek } from 'date-fns'
```
dadlerj, line 1: This file duplicates logic from the sourcegraph/sourcegraph repo: https://sourcegraph.com/github.com/sourcegraph/sourcegraph@90826c6d7370511469e528258ee6ee446f784870/-/blob/client/shared/src/telemetry/web/userTracker.ts

src/telemetry/index.ts
```diff
@@ -0,0 +1,69 @@
...
+            },
+            new GraphQLTelemetryExporter('https://sourcegraph.com'),
+            [new MarketingTrackingTelemetryProcessor(new TrackingMetadataProvider())],
+            {
+                /**
+                 * Use buffer time of 100ms - some existing buffering uses
+                 * 1000ms, but we use a more conservative value.
+                 */
+                bufferTimeMs: 100,
+                bufferMaxSize: 10,
```
bobheadxi, line 35: I think we should disable buffering for the about site, I'm not sure the mechanism is very well-proven yet (IIRC someone from one of the client teams had taking a closer look at buffering on their backlog)

bobheadxi:
```suggestion
                /**
                 * Disable buffering for now
                 */
                bufferTimeMs: 0,
```

bobheadxi (dismissed):
Nice, overall LGTM! Places where you noted `This file duplicates logic from the sourcegraph/sourcegraph repo` could be good candidates for placing in a shared repo 😁 As usages grow, I think you can add some of this as subpackages of https://github.com/sourcegraph/telemetry

bobheadxi:
That said, am I the only person who can review this? Do we have other owners of the about site we can pull in?

dadlerj:
Awesome, thanks @bobheadxi ! Yes, agreed on moving some of this to sourcegraph/telemetry at some point. Especially as we add more sites. I suspect the code for s0, for example, will be 99% identical...

And I've pinged #discuss-marketing for another reviewer

gitstart-sourcegraph (approved):
Thanks @dadlerj, this LGTM!.

PR #6971, opened 2024-05-29
Remove v1 telemetry and fully replace with v2 telemetry

dadlerj:
Follow up to https://github.com/sourcegraph/about/pull/6967

This PR:
* Add v2 telemetry to replace all v1 telemetry
* Fully remove v1 telemetry
* Delete a few unused/outdated files

Tested locally and using CI


dadlerj:
Vova, feel free to tell me if you're not a good reviewer for this one!

src/components/Resources/useFilters.ts
```diff
@@ -14,18 +14,18 @@ interface UseFilters {
...
  * This hook is used for getting, setting, and resetting filters
  * based on our resource items data.
  */
-export const useFilters = (): UseFilters => {
-    const logResourceClickEvent = (groupTitle: string, type: string): void => {
-        const eventName =
-            groupTitle === 'Content Type' ? EventName.RESOURCE_CONTENT_TYPE_FILTER : EventName.RESOURCE_SUBJECT_FILTER
+export const useFilters = ({ telemetryRecorder }: UseFilterProps): UseFilters => {
+    const logResourceClickEvent = (groupTitle: string, type: string): void =>
+        telemetryRecorder.recordEvent(`resources.filter.${groupTitle === 'Content Type' ? 'contentType' : 'subject'}`, 'toggle', { privateMetadata: { type }})
```
vovakulikov, line 27: I see that previously, we used different event names, and that's why we do it in the new telemetry, but wouldn't it be better if we used one event name but put this content type property in the event metadata?

vovakulikov (approved):
LGTM, one comment that I really liked that previously we used enums for 
event names, this allowed us to easily find event name usage. I would stick for it
with new event names as well.

dadlerj:
@vovakulikov using enums for event names gets a bit less clean, as we now split event "names" into features and actions, rather than a single string.

In my opinion, the _more_ important thing is to avoid using variables as event names, which we (to some degree), get from the `KnownString` type in the telemetry package. You can see I still occasionally use string interpolation for sub-features, but I'd love to avoid that (and I'll take your recommendation above!).

As long as the string itself is searchable, I don't particularly care if it's in an enum or just a hardcoded string. What's your reaction to that?

gitstart-sourcegraph (commented):
Thanks @dadlerj 
Changes LGTM. Can we fix the failing `Ci/CD`?

dadlerj:
@gitstart-sourcegraph that would be great, can you take a stab at it? Let me know if any significant changes are required and I can dig in.

src/components/Video.tsx
```diff
@@ -39,6 +39,7 @@ export const Video: FunctionComponent<Video> = ({
     autoPlay = true,
     className,
     thumbnail,
+    telemetryRecorder
```
gitstart-sourcegraph, line 42: Component Video.tsx now has a required prop "telemetryRecorder" however it is not being passed in all places where the component is being used such as in `batch-changes.tsx`, please add the required telemetryRecorder prop to the Video component

src/components/cta/ExternalProvider.tsx
```diff
@@ -1,16 +1,26 @@
-import { FunctionComponent, PropsWithChildren, ReactNode, useEffect } from 'react'
+import { FunctionComponent, PropsWithChildren, Provider, ReactNode, useEffect } from 'react'
```
gitstart-sourcegraph, line 1: Also lets fix this error that is causing the yarn lint fail, "'Provider' is defined but never used"

src/pages/demo/cody.tsx
```diff
@@ -15,7 +15,6 @@ import {
 } from '../../components'
 import { useAuthModal } from '../../context/AuthModalContext'
 import { breakpoints } from '../../data/breakpoints'
-import { EventName, getEventLogger } from '../../hooks/eventLogger'
 import { useWindowWidth } from '../../hooks/windowWidth'
```
gitstart-sourcegraph, line 18: Also lets fix the error on `line 1` that is causing the yarn lint fail:  "'useEffect' is defined but never used"

src/pages/cody.tsx
```diff
@@ -22,7 +22,6 @@ import { HowCodyWorks } from '../components/cody/HowCodyWorks'
 import { ChooseYourLlmSection } from '../components/Enterprise/ChooseYourLlmSection'
 import { useAuthModal } from '../context/AuthModalContext'
 import { breakpoints } from '../data/breakpoints'
```
gitstart-sourcegraph, line 24: Also lets fix the error on `line 1` that is causing the yarn lint fail:  "'useEffect' is defined but never used"

PR #7001, opened 2024-06-12
Stage VSC 1.22 blog and model updates for pricing

iskyOS:
- Adds blog file for the VS Code release post
- Updates pricing page for Cody Pro to include Gemini 1.5 Flash and Pro


kalanchan (approved):
excellent content as always!

PR #7049, opened 2024-06-26
Add documentation on turning banner on and off

jdorfman:


jdorfman:
@iskyOS check the diff and merge so you and other know how to set the state of the banner

iskyOS (approved):
nice!

PR #7059, opened 2024-07-03
Update link to previous version of self-hosted terms

dadlerj:


varungandhi-src:
@dadlerj if you rebase on top of `main`, that should fix the issue with pnpm/action-setup

PR #7080, opened 2024-07-12
fixing papercuts on homepage and menu

traceyljohnson:
- Adding a little more padding above the H1
- Adjust size of icons so they're less chongo 
- Fixed wonky width on bentos in Code Search section
- Fixing flickering of the navmenu on scroll


iskyOS:
Added a commit to fix the navbar transition

PR #7217, opened 2024-09-06
Updating Inter font family to latest version

traceyljohnson:
- Instead of using Google font version of Inter, I've swapped it for the one directly from https://rsms.me/inter/
- Added html to head
- Added root css to global.css
- Added a fancy gradient to emphasized homepage hero text 
- Cleaned up that styling while I was at it


traceyljohnson:
@iskyOS can you verify that I haven't messed anything up? I've swapped our Inter font family for the latest version and made some adjustments to the homepage hero text treatments. I know you're running AB tests on it, so LMK if I need to wait till those are done before merging this.

iskyOS (approved):
Looks great and doesn't break the A/B/C/D test!

PR #7293, opened 2024-10-11
Lint bad SVGs

toolmantim:
Adds a linting action and package script (`pnpm run lint:images`) to prevent bad SVGs from making their way into the repo


toolmantim:
Ah, a bug on Linux. Will fix…

chris-sev (approved):

PR #7439, opened 2024-11-19
feat: add event end date and format date range

jdorfman:
- Add `endDate` field to `Event` interface
- Implement `formatDateRange` function to handle single-day and multi-day events
- Update event date display to use `formatDateRange`

### DIFF
--- a/src/components/EventsList.tsx
+++ b/src/components/EventsList.tsx
@@ -268,10 +268,10 @@ const formatDateRange = (startDate: Date, endDate: Date): string => {
     const year = format(startDate, 'yyyy')
 
     if (startMonth === endMonth) {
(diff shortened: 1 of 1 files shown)


jdorfman:
@chris-sev, this has been on my to-do list for a while. Most events are multi-day, and sometimes, they start in one month and go on to the next. 

Zight 2024-11-19 at 12 57 40 PM png

chris-sev (changes requested):
looking good! my only nit is could we add some spacing before and after the dash?

CleanShot 2024-11-19 at 14 52 12@2x

jdorfman:
Good call, looks way better with the space

Zight 2024-11-19 at 1 54 54 PM png

chris-sev (approved):

